PR #13, opened 2019-06-21
Update chapter10.md

rabiabdel:
Adding proposed outline for Chapter 10.

### DIFF
--- a/doc/ref_model/chapters/chapter10.md
+++ b/doc/ref_model/chapters/chapter10.md
@@ -5,10 +5,8 @@
 ## Table of Contents
 * [10.1	Introduction.](#10.1)
 * [10.2	Compliance, Verification, and Certification process and life cycle.](#10.2)
-  * [10.2.1	NFVI reference architectures.](#10.2.1)
-  * [10.2.2 NFVI reference implementations.](#10.2.2)
-  * [10.2.3 VNFs reference implementations.](#10.2.3)
-  * [10.2.4 Test suites.](#10.2.4)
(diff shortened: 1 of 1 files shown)


cntt-n:
Add more entries to outline

rabiabdel:
Ok Change made

rabiabdel:
@ Pankaj, please add your comment to this Commit in here.

pgoyal01:
Unable to edit the "pull".
I think the Chapter is about Compliance, Verification and Certification (CVC) and thus needs to focus on the CVC Lifecycle including processes and tools – model level. The CVC services shall be performed on CNTT specified Reference Implementations – but these reference implementations are not yet specified.  It appears to me that the best plan would be to modify the OPNFV CVC processes (diagrams) to refer to CNTT NFVi Reference Implementation wherever they have NFVi etc. Do you agree? If yes, then the chapter content could be as follows:
 
#  Compliance, Verification and Certification (CVC)
[10.1  Introduction]
[10.2  Compliance process adapted from OPNFV (specifies the process with changes)]
[10.4  Verification (rest same as above)]
[10.4  Certification (rest same as .1)]
[10.5  Comparison chart of CVC and recommendation for NFVi and C/VNF’s]

cntt-n:
Hi @pgoyal01 i have made changes to reflect your feedback, is the change aligned with your expectation?

rabiabdel:
Hi @pgoyal01 i have made changes to reflect your feedback, is the change aligned with your expectation?

PR #16, opened 2019-06-24
Update chapter04 from origional content.md

rabiabdel:


rabiabdel:
@pgoyal01: i have created your Pull Request on top of origonal contnet, please double check and let me know if you are ok with it.

doc/ref_model/chapters/chapter04.md
```diff
@@ -21,25 +26,139 @@ The idea of the infrastructure profiles catalogue is to have a predefined set of
...
+
+1. **CPU Oversubscription Ratio**: is based on the number of threads available. For example, on a 24-core host with HT, there are 48 vCPUs with 1:1 CPU Ratio and 96 vCPUs with 2:1 CPU Ratio.
+
+2. **HT (Hyper Threading support)**: Enabled on all servers. Gets 2 hyper threads per physical CPU.  Always ON. 
+3. **CPU Pinning**: vCPU is pinned to a physical core and dedicated to the requesting VM. 
+4. **NUMA (Non-Uniform Memory Access) Alignment**: Indicates that vCPU will be on a Socket that is aligned with the associated NIC card and memory.  Important for performance optimized VNFs. 
+5. **Huge Pages**: By default, CPU allocate RAM in 4K chunks. Huge Pages enable to allocate in larger Chunks (such as 2MB, 1GB).  This helps improve performance in some cases. 
+6. **SR-IOV (Single-Root Input/Output Virtualization)**: Allows SR-IOV ports to be used in VM. 
+7. **DPDK vRouter**: Virtual Router integrated with the Intel DPDK (Data Plane Development Kit) libraries. 
+8. **Kernel vRouter**: Virtual Router that runs in the Kernel space
```
burakkazan: What is the correct terminology here ? virtual router or virtual switch ? I think we need to replace router term with switch

doc/ref_model/chapters/chapter04.md
```diff
@@ -21,25 +26,139 @@ The idea of the infrastructure profiles catalogue is to have a predefined set of
...
+
+Flavour Capabilities:
+
+1. **CPU Oversubscription Ratio**: is based on the number of threads available. For example, on a 24-core host with HT, there are 48 vCPUs with 1:1 CPU Ratio and 96 vCPUs with 2:1 CPU Ratio.
+
+2. **HT (Hyper Threading support)**: Enabled on all servers. Gets 2 hyper threads per physical CPU.  Always ON. 
+3. **CPU Pinning**: vCPU is pinned to a physical core and dedicated to the requesting VM. 
+4. **NUMA (Non-Uniform Memory Access) Alignment**: Indicates that vCPU will be on a Socket that is aligned with the associated NIC card and memory.  Important for performance optimized VNFs. 
+5. **Huge Pages**: By default, CPU allocate RAM in 4K chunks. Huge Pages enable to allocate in larger Chunks (such as 2MB, 1GB).  This helps improve performance in some cases. 
+6. **SR-IOV (Single-Root Input/Output Virtualization)**: Allows SR-IOV ports to be used in VM.
```
burakkazan: Maybe it is better to explain as Direct Transfer which is kernel bypass technologies , SR-IOV and PCI Pass. 

**Direct Transfer:** This is the fastest option for handling packet in the user data plane. There are two options available which are SR-IOV and PCI pass-through, SR-IOV can virtualize the NIC to several VM to use but PCI pass-through will cause NIC to be available only for one VM.

doc/ref_model/chapters/chapter04.md
ASawwaf: The industry name is vSwitch , vRouter is vendor naming  (Juniper , Nuage and E//) which vRouter is an enhancement  vSwitch + DPDK

doc/ref_model/chapters/chapter04.md
ASawwaf: Great , but we initially  agreed that SRIOV and PCI-PT not our focus in this document , 
also small comment on the definition "SR-IOV can virtualize the NIC to several VM to use but PCI pass-through will cause NIC to be available only for one VM."

we can define it SR-IOV can virtualize the NIC to several VF not VM

rabiabdel: Agree with Ahmed. This PR has been brought directly from the original CR. We need to make sure it is aligned with the principles we defined in chapter 1. I will take a stab at editing it in a way that I believe will align with our principles and bring it for review. Is that OK Burak?

pgoyal01:
Rabi,

That would be great. Just a comment on Ahmed’s comment about SR-IOV not being a focus of the initial scope – does the catalogue not have to support the hardware profiles and software flavours based on SR-IOV?

Best regards,
Pankaj
--------
pg683k@att.com | w  720.475.4530  |  m 720.327.0104

cntt-n:
Hi @pgoyal01 will it be ok to close your other Pull Request for Chapter 4? it was not created against the original content and this one is intended to do that. if you are OK, i will close the other one and we iterate on this one?

pgoyal01:
Yes – it is OK

Best regards,
Pankaj
--------
pg683k@att.com | w  720.475.4530  |  m 720.327.0104

doc/ref_model/chapters/chapter04.md
burakkazan: it’s ok @rabiabdel

pgoyal01:
Yes it is Ok to close

Best regards,
Pankaj
--------
pg683k@att.com | w  720.475.4530  |  m 720.327.0104

cntt-n:
Hi @pgoyal01 : as discussed during the meeting, we are not excluding parameterised options but i see no value of specifying naming convention or encouraging it. The way i see it, we recommend T-shirt Sizes and we made it clear we allow custom ones, but i have some concerns promoting it in the document. we can just say there are our preferred ones, but operators might allow other sizes? doesn't this suffice?

rabiabdel:
Hi @pgoyal01 : as discussed during the principles meeting, we are not excluding parameterised options but i see no value of specifying how this should be done in details as this will touch on orchestration and VNFD which is not the focus here. The way i see it, we recommend T-shirt Sizes and we made it clear we allow custom ones, but i am not sure we need to be promoting it in the document. we can just say these are our preferred ones, but operators might allow other sizes? doesn't this suffice? let me know.

pgoyal01:
If we allow custom sizes we should specify how they can be defined -- only in abstract terms. The deleted parameterised flavor section has one abstract method of specifying the sizes.  It is independent of orchestration (just as the named flavours in this chapter are). There are multiple advantages of parameterised flavours.  The best practice for VNFC is to be reusable and granular (single capability) and to keep them as lightweight as possible (while providing single capability). Named flavours usually force these VNFC's to be deployed using a larger sized flavour then necessary; the VNFC being too large for the next small version. 
On another note, cloud service providers (CSP) have their own named flavour series.  Instead of depending upon names would it not be preferable that VNF vendors provide their specifications as a  parameterised list that is CSP agnostic?

rabiabdel:
Hi Pankaj,
I understand the benefits you highlighted, my point in regards to T-shirt sizes is the strategy, commercial and business benefits we gain, as well as long term gain moving to cloud native. We discussed those in Barcelona in length and during many meetings, hence We decided not disallowing the customised one. 

In case of custom flavours, I agree too with the benefits of specefyin how to parameterise them in an abstracted way to be cloud platform agnostics, but my concern is that this is what TOSCA/HEAT are there for and I really think it is out of the scope of this work, we don't want to go to that path, it is like boiling the ocean in my views. 

Similarly for prefereed T-shirt sizes, we don't need to dectate how they are created, we just need to clarify the sizes, types and leave it to TOSCA or Heat do it. If there are anything for pre-dedined flavours that is trying to dictate it, we should remove it (probably we should remove the naming convention at the end). What do you think?

pgoyal01:
So can we include a section on Parameterised Flavours with the statement that these may be created using TOSCA, HEAT or VIM APIs.

rabiabdel:
OK @pankaj I will add a paragraph to clarify it and put it for your review.

rabiabdel:
Hi @pgoyal01 based on the discussion we had by email, here is the sentence added for Customised Flavours: 

>_*Note:* Customised (Parameterized) flavours can be used in concession by operators and , if needed, are  created using TOSCA, HEAT templates and/or VIM APIs.

doc/ref_model/chapters/chapter04.md
```diff
@@ -4,13 +4,16 @@
...
+* [4.3 Storage Extensions.](#4.3)
+  * [4.3.1 Available storage extensions.](#4.3.1)
+* [4.4 Instance types.](#4.4)
+  * [4.4.1 B Instances (Basic).](#4.4.1)
+  * [4.4.2 N Instances (Network Intensive).](#4.4.2)
+  * [4.4.3 C Instances (Compute Intensive).](#4.4.3)
+* [4.5 One stop shop.](#4.5)
+  * [4.5.1 Naming convention.](#4.5.1)
 
 Infrastructure profiles are collection of capabilities, metrics, compute flavours, interface options, storage extensions, and acceleration capabilities that are offered by the infrastructure to VNFs. Infrastructure profiles are offered to VNFs in form of instance types with their corresponding options and extensions.
```
hdamker: "Infrastructure profile" is a) not defined in terminology and can easily be confused with NFVI Software Profile and NFVI Hardware Profile. Both is not intended here, I suppose. @rabiabdel please compare with your comment https://github.com/cntt-n/CNTT/pull/28#issuecomment-507197821

doc/ref_model/chapters/chapter04.md
```diff
@@ -21,25 +24,73 @@ The idea of the infrastructure profiles catalogue is to have a predefined set of
...
+n10, n10D, n10T, n10Q, n10P, n10H	|1x 10, 2x 10, 3x 10, 4x 10, 5x 10, 6x 10 Gbps
+n25, n25D, n25T, n25Q, n25P, n25H	|1x 25, 2x 25, 3x 25, 4x 25, 5x 25, 6x 25 Gbps
+n50, n50D, n50T, n50Q, n50P, n50H	|1x 50, 2x 50, 36x 50, 4x 50, 5x 50, 6x 50 Gbps
+
+<p align="center"><b>Table 4-3:</b> Virtual Network Interface Specification Examples.</p>
+
+<a name="4.3"></a>
+##  4.3 Storage Extensions
+Multiplee non-ephemeral storage volumes can be attached to virtual computes (*replace with decided term*) for persistent data storage. Each of those volumes can be configured with the required performance catagories (*storage performance prfiles*).
+***Add comment about CEPH distributed storage. (Potentially create new profile for it).***
```
hdamker: Why is CEPH distribute storage different? It can be the block storage backend behind e.g. Cinder in OpenStack. 

I assume that Object Storage (with Swift or S3 API) is/will be described somewhere else in the document, as it would be an infrastructure service not bound to a specific compute instance.

ASawwaf: @hdamker, I recommend not to mention any specific backend storage ( CEPH), Just block, file, object storage ( as the type of storage that required from multiple VNF vendor ) and most seen storage is block

hdamker: Yes, I agree - that's how my comment can be resolved: Delete "***Add comment about CEPH distributed storage. (Potentially create new profile for it).***" in 4.3.

ASawwaf: Great , Thanks

ASawwaf:
@rabiabdel , Just a question, where is Flavour Capabilities, should be defined? , as I believe it is related to VNF requirements, as I will have some small comments on the definition ( i will send it shortly )

doc/ref_model/chapters/chapter04.md
TommyKTheDJ: Agree Herbert - I took file/object out of that section in a much earlier version for that reason 👍

markshostak:
Approved contingent upon addressing assorted comments listed above. Issue to be created to track.

rabiabdel:
Thanks everyone, i will merge this one then and create another one to address comments.

rabiabdel:
> @rabiabdel , Just a question, where is Flavour Capabilities, should be defined? , as I believe it is related to VNF requirements, as I will have some small comments on the definition ( i will send it shortly )

Hi Ahmed, it depends on what you mean by flavour here. capabilities are represented by the profiles and they are defined in Chapter 3. Hope this help.

PR #17, opened 2019-06-24
Section 1.6 - ETSI NFV architecture mapping

xavier-grall:


BernardTsai-DT:
Would it not make sense to include the VIM into the NFVI as well since we are also intending to expose the APIs to the workloads.

PR #20, opened 2019-06-25
Updated Chap 3

markshostak:
Added 3.2, including comments from Bernard and Pankaj.


BernardTsai-DT:
I had the impression that the virtual resources would belong to the external entities as well.

PR #30, opened 2019-07-01
Update chapter02.md

ASawwaf:
TEST


markshostak:
Per mtg 2/7/19 ASawwaf  to update or withdraw.

PR #37, opened 2019-07-02
Adding image to chap05

karinesevilla:


karinesevilla:
To be inserted in chap5 , figure 5-1.
The intention is to make the link with the chap3 section3.1:
"NFVI software profile: This layer consists of both the host Operating System (OS) responsible for managing the hardware resources of the layer below as well as the virtualization/containerization technology which turns hardware components into a pool of logical resources and dynamically allocates them to the layer above."

pgoyal01:
Bertrand still to correct Chapter 3 model -- all changes being made now referring to the current model will have to be revisited. Technically, profile is just information.

Issue #32, opened 2019-07-02
Rabi Chap 4 PR


markshostak:
Various comments recorded against PR need to be resolved.

rabiabdel:
new PR was created to reflect this issue: https://github.com/cntt-n/CNTT/pull/38

rabiabdel:
Resolved, New PR created for this: https://github.com/cntt-n/CNTT/pull/38

rabiabdel:
Close Now

PR #38, opened 2019-07-02
Chapter 04 (based on comments (Re Issue# 32)) 

rabiabdel:
Addressing Comments form @hdamker, @ASawwaf, @TommyKTheDJ

Issue: https://github.com/cntt-n/CNTT/issues/32

### DIFF
--- a/doc/ref_model/chapters/chapter04.md
+++ b/doc/ref_model/chapters/chapter04.md
@@ -15,7 +15,7 @@
 * [4.5 One stop shop.](#4.5)
   * [4.5.1 Naming convention.](#4.5.1)
 
-Infrastructure exposes set of capabilities, metrics, compute flavours, interface options, storage extensions, and acceleration capabilities to VNFs. Those are offered to VNFs in form of instance types with their corresponding options and extensions.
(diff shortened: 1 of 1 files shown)


rabiabdel:
@hdamker: 
Text Changed to:
Infrastructure exposes set of capabilities, metrics, compute flavours, interface options, storage extensions, and acceleration capabilities to VNFs. Those are offered to VNFs in form of instance types with their corresponding options and extensions.

@hdamker, @ASawwaf, @TommyKTheDJ: CEPH distributed sotrage removed.

hdamker:
I'm supporting Rabi's version, would just slightly change it to make the reference clear:

"Infrastructure exposes set**s** of capabilities, metrics, compute flavours, interface options, storage extensions, and acceleration capabilities to VNFs. Those **sets** are offered to VNFs in form of instance types with their corresponding options and extensions."

Shortly checked: "instance types" is mainly used by Amazon AWS, but as it is also a generic term, that's fine for me.

pgoyal01:
Herbert,
That is fine. We need a definition of instance types -- it is not a term that is currently being used in the document.  Could we then replace "flavours" by "Instance Types" -- much simpler then using NFVI Software Profile or Configuration?

rabiabdel:
Hi @hdamker : I have updated the text to reflect your suggestion above.
@pgoyal01 : Agree, i like your suggestion. let me think about it and propose something.

PR #67, opened 2019-07-09
Chap-1.6 Updated NFV mapping scope image

markshostak:
Chap-1.6 Updated NFV mapping scope image


rabiabdel:
That looks very good. @karinesevilla, @xavier-grall are you guys aligned with this update?

pgoyal01:
Does the Reference Model include the VIM as is implied by the diagram?

karinesevilla:
@rabiabdel, @pgoyal01
Yes agree with the picture, I think it would have been difficult to exclude the VIM of our scope

pgoyal01:
Do we need a chapter on VIM -- essential capabilities?

rabiabdel:
@pgoyal01 this is a great point, i think this can be part of Chapter 3. i don't think we need a whole chapter for it. if you look at some of the tables in Chapter 3, they are VIM metrics. we might just need to clarify that bu adding a third category (exposed, internal, and VIM).

rabiabdel:
48 hours passed, i will go and approve this.

PR #68, opened 2019-07-09
Chapter 3( cleanup)

pgoyal01:
Updated teh figures to align with the modified model text added on Monday July 8, 2019.


rabiabdel:
Hi @pgoyal01 some of the terms introduced here are not aligned with the terminology approved. for example: we agreed on NFVI HW profile whereas this proposed changing it to Physical resources later.

can you update please?

BernardTsai-DT:
Hi @rabiabdel , @pgoyal01 - I would recommend to keep section 3.1. as simple as possible since it should serve as an overview and introduction to the following chapters. There name NFVI HW Profile was regarded to be a bit to specific in the sense that it relates to the information of how to structure the physical resources. Therefore the proposal was to simplify the layer description in such a way that it could relate to both the actual resources and the configuration aspects of these resources.

rabiabdel:
> Hi @rabiabdel , @pgoyal01 - I would recommend to keep section 3.1. as simple as possible since it should serve as an overview and introduction to the following chapters. There name NFVI HW Profile was regarded to be a bit to specific in the sense that it relates to the information of how to structure the physical resources. Therefore the proposal was to simplify the layer description in such a way that it could relate to both the actual resources and the configuration aspects of these resources.
> 
> The layers are all part of the Reference Model (RM). I have compared section 1.5 and 3.1. and the chapters headers:
> 
> **We have in 1.5:**
> 
> * Infrastructure Abstraction & Catalogue
> * NFVI SW
> * NFVI HW
> 
> **In 3.1:**
> 
> * Virtual Infrastructure Resources
> * NFVI SW
> * Physical Infrastructure Resources
> 
> **Chapters:**
> 4. Catalogue
> 5. Reference NFVI SW profiles and configurations
> 6. Reference NFVI HW profiles and configurations
> 
> I would recommend to streamline and simplify the wording as much as possibe, e.g.:
> 
> * Virtual Infrastructure Resources
> * NFVI SW
> * NFVI HW
>   and then explain that the catalogues and profiles relate to the configurations of the layers.

I agree @BernardTsai-DT. @pgoyal01 do you agree? Can you  reflect those changes or would you like me to do that on your behalf?

PR #69, opened 2019-07-09
Update Chapter04.md

pgoyal01:
Consolidated Network Interface Options so as to not repeat them for the different types (B, N and C); adjusted Table and Sections numbers.


rabiabdel:
Passed 48 hours, approved.

PR #72, opened 2019-07-10
Ch01 (1.5 Scope)

rabiabdel:
Place holder to clarify the relation between RM, RA, and RI. it is also to update the scope figure.


BernardTsai-DT:
I still believe that the VIM is to be regarded part of the NFVI SW layer and therefore should be included into scope of the Reference Model. And when we will be talking about Reference Architectures I would imagine that the typical APIs e.g. OpenStack, OpenContrail, etc. would be the actual APIs which the NFVI would expose from a management perspective.

rabiabdel:
> I still believe that the VIM is to be regarded part of the NFVI SW layer and therefore should be included into scope of the Reference Model. And when we will be talking about Reference Architectures I would imagine that the typical APIs e.g. OpenStack, OpenContrail, etc. would be the actual APIs which the NFVI would expose from a management perspective.

agreed, will reflect it in the text/diagrams

rabiabdel:
@BernardTsai-DT Changes made to diagram as discussed!

rabiabdel:
replace figure to say VIM and approved afterwords

PR #114, opened 2019-07-16
CH 6 HW/SW Config Diagram chnage

pgoyal01:
Modified CH 6 HW/SW Config Diagram -- NUMA added to both Network and Compute profiles


rabiabdel:
Hi @pgoyal01 This Pull Request updates the powerpoint file (which is important, Thanks for it) but it doesn't upload an update figure (under the figures directory).

PR #116, opened 2019-07-17
update to latest Rls 1.x master

markshostak:


markshostak:
This commit is erroneous. It will be deleted.

PR #121, opened 2019-07-24
Fixing numbering in chapter 3

jproque-dell:
changing 3.5 to 3.3


jproque-dell:
My first PR! proud to be part of the community, will do my best to participate.

PR #123, opened 2019-07-24
Disambiguate VNF vs CNF terminology

fzdarsky:
This patch disambiguates Chapter 1 terminology definitions of VNFs and CNFs.
Current definition mixes (at least) two different aspects: delivery as virtual appliance vs as container and whether or not the implementation adheres to cloud-native principles.
Resolving this in terms of VMs vs container aspect, a) for consistency with NF definition above and b) because this aspect actually impacts the following sections.

### DIFF
--- a/doc/ref_model/chapters/chapter01.md
+++ b/doc/ref_model/chapters/chapter01.md
@@ -57,10 +57,10 @@ This section defines the main terms used in this document; these deinitions are 
 - **Network Function (NF)**:  functional block or application within a network infrastructure that has well-defined external interfaces and well-defined functional behaviour.
   - Within **NFV**, A **Network Function** is implemented in a form of **Virtualised NF** or a **Containerised NF**.
 - **Network Service (NS)**: composition of **Network Function**(s) and/or **Network Service**(s), defined by its functional and behavioural specification, including the service lifecycle.
-- **Virtual Network Function (VNF)**: a software implementation of a **Network Function**, capable of running on the **NFVi**.
-  - **VNF**s are built from one or more VNF Components (**VNFC**) and, in most cases,  the VNFC is hosted on a single VM or Container.
-- **Cloud-native (containerised) Network Function (CNF)**: **VNF** with a full adherence to cloud native principles, or a **VNF** that is transitioning to cloud native. 
-  >_*Note:*_ It is a containerised **VNF** that is microservices-oriented, to increase agility and maintainability, and that can be dynamically orchestrated and managed to optimize resource utilization; the containers can be Linux, Docker or other similar container technology.
+- **Virtualised Network Function (VNF)**: a software implementation of a **Network Function** delivered as set of VMs capable of running on a **NFVI**.
+  - **VNF**s are built from one or more VNF Components (**VNFC**) and, in most cases, a single VNFC instance is hosted on a single VM.
+- **Cloud-native (containerised) Network Function (CNF)**: a software implementation of a **Network Function**, delivered as a set of container images capable of running on a container orchestration platform. 
+  >_*Note:*_ It is a containerised **NF** that is microservices-oriented, to increase agility and maintainability, and that can be dynamically orchestrated and managed to optimize resource utilization; the containers can be Linux (OCI-compliant) or other operating system container technology.
 - **Virtual Application (VA)**: is more of a general term for software which can be loaded into a Virtual Machine. 
   >_*Note:*_ a **VNF** is one type of VA.
 - **Workload**: Workload refers to software running on top of compute resources such as **VMs** or **Container**s. Most relevant workload categories in context of NFVI are:


doc/ref_model/chapters/chapter01.md
```diff
@@ -57,10 +57,10 @@ This section defines the main terms used in this document; these deinitions are
...
   - Within **NFV**, A **Network Function** is implemented in a form of **Virtualised NF** or a **Containerised NF**.
 - **Network Service (NS)**: composition of **Network Function**(s) and/or **Network Service**(s), defined by its functional and behavioural specification, including the service lifecycle.
-- **Virtual Network Function (VNF)**: a software implementation of a **Network Function**, capable of running on the **NFVi**.
-  - **VNF**s are built from one or more VNF Components (**VNFC**) and, in most cases,  the VNFC is hosted on a single VM or Container.
-- **Cloud-native (containerised) Network Function (CNF)**: **VNF** with a full adherence to cloud native principles, or a **VNF** that is transitioning to cloud native. 
-  >_*Note:*_ It is a containerised **VNF** that is microservices-oriented, to increase agility and maintainability, and that can be dynamically orchestrated and managed to optimize resource utilization; the containers can be Linux, Docker or other similar container technology.
+- **Virtualised Network Function (VNF)**: a software implementation of a **Network Function** delivered as set of VMs capable of running on a **NFVI**.
```
ulikleber: We should not create a definition that is different from the ETSI ISG NFV definition of a VNF, which is available in ETSI GS NFV 003. Not only is the ETSI definition not restricted to VM based VNFs, the VNF is not delivered as a set of VMs, but as a VNF package, containing e.g. a VNFD and images.

doc/ref_model/chapters/chapter01.md
```diff
@@ -57,10 +57,10 @@ This section defines the main terms used in this document; these deinitions are
...
+  - **VNF**s are built from one or more VNF Components (**VNFC**) and, in most cases, a single VNFC instance is hosted on a single VM.
+- **Cloud-native (containerised) Network Function (CNF)**: a software implementation of a **Network Function**, delivered as a set of container images capable of running on a container orchestration platform.
```
ulikleber: An implementation of a NF cannot be delivered just as a set of container images. There needs to be other information combined with that.

doc/ref_model/chapters/chapter01.md
fzdarsky: The ETSI NFV definition of VNF ("Virtualised Network Function (VNF): implementation of an NF that can be deployed on a Network Function Virtualisation Infrastructure (NFVI)") *is* implicitly restricted to VM-based NFs, because one cannot run OS containers on a machine (compute resource) without the OS, a container runtime and (arguably) container orchestration system on top of the compute resource provided by the NFVI.

It seems the intention of the original author(s) was to differentiate between NFs deployed in VMs and those deployed in containers. Do you have another proposal how to extend ETSI NFV definition in a way that addresses this in a more compatible manner?

ulikleber: The definition of the VNF is not focussing on the type of execution environment. It is should be seen from a MANO point of view. A VNF is explicitly meant to be agnostic of the used virtualization technology. E.g. it does not assume that there is no operating system below. Current ETSI NFV container study IFA029 which is currently in final review also does not change the definition of the VNF, but studies how to allow containerized workloads managed as VNFs.
If we need a term for a NF implemented in VMs and not in containers, we should use a new acronym, e.g. VMNF or V-VNF.

For the definition of CNF I recommend, that also MANO aspects should be considered, and allow orchestration of containerized and VM-based VNFs in a similar way. Nobody wants to create a new MANO stack for CNFs.

rabiabdel:
Issue: https://github.com/cntt-n/CNTT/issues/139

doc/ref_model/chapters/chapter01.md
xavier-grall: I fully agree that we should remain aligned with ETSI definitions/specifications, so CNF being only a possible implementation of ETSI-defined VNF, the current VNF/CNF-related definitions should be preserved.

fzdarsky: I'd suggest "remain aligned with ETSI defs/specs where it makes sense", i.e. take a non-dogmatic approach here.

For CNTT to have impact, it needs to be a) as concrete as possible, b) aligned with the reality where the industry has been/is moving, and c) timely. Being dogmatic on following ETSI NFV would be counter-productive to this. For sure we don't want to repeat another IFA005/006 odyssey.

I'm fully aware of ETSI NFV's _intention_ when defining VNFs/VNFCs as agnostic towards VM vs container deployment. After all, I was the one leading the definition back then.
The reality is, though, that many of the following ETSI NFV specs were written with "VNF = VM-deployed" in mind. Do the experiment and mentally substitute VNFs --> containerized app and you'll easily find technical flaws all over the specs.

More importantly, though:
Suggesting containerized NFs should be managed and orchestrated the same as VM-based NFs completely misses the point of Kubernetes cloud-native operations!
Note, for example, that Kubernetes *is* a VNFM (cf. Operator pattern) and actually ticks all the boxes that after 6 years we as an industry failed to deliver on the VM side of the aisle. Let's not lose that!

So I fully agree with the sentiments that "MANO aspects should be considered" and "nobody wants to create a new MANO stack for CNFs." This is exactly why we need to remain aligned with / address potential gaps in the CNCF's solutions.

rabiabdel:
This will be closed for now and issues will be discussed. @bfcohen to assign a person to create a PR.

PR #133, opened 2019-07-30
Post Paris Cleanup

rabi-abdel:

PR #138, opened 2019-08-01
Empty chapter after merge with chapter 5

samuelhellec:


rabiabdel:
Already covered in https://github.com/cntt-n/CNTT/pull/133

PR #148, opened 2019-08-02
Ref Arch ToC proposal and cleanup

rabi-abdel:


rabi-abdel:
@pgoyal01it is already the case

Issue #201, opened 2019-08-12
[Chapter 3] Clarify the intention of the design principle: Declarative


BernardTsai-DT:
Clarify the intention of the design principle:

**Declarative**: the model should allow for a declarative description of the required NFVI capabilities for on-boarding and maintaining workloads

proposal:

_**Specifc:**_ the model should allow for a **_machine-readable_** description of the required NFVI capabilities for on-boarding and maintaining workloads

PR #202, opened 2019-08-12
[Reference Model] Chapter 3 - Clarify design principle: Declarative

BernardTsai-DT:
#201 Clarify the intention of the design principle: Declarative

### DIFF
--- a/doc/ref_model/chapters/chapter03.md
+++ b/doc/ref_model/chapters/chapter03.md
@@ -23,7 +23,7 @@ Although a couple of explicit and implicit abstraction models (e.g. in the conte
 -	**Scope:** the model should describe the most relevant virtualised infrastructure resources (incl. acceleration technologies) an NFVI needs to provide for hosting Telco workloads
 -	**Separation of Concern:** the model should support a clear distinction between the responsibilities related to maintaining the network function virtualisation infrastructure and the responsibilities related to managing the various VNF workloads
 -	**Simplicity:** the amount of different types of resources (including their attributes and relationships amongst one another) should be kept to a minimum to reduce the configuration spectrum which needs to be considered
--	**Declarative:** the model should allow for a declarative description of the required NFVI capabilities for on-boarding and maintaining workloads
+-	**Specific:** the model should allow for a machine-readable description of the required NFVI capabilities for on-boarding and maintaining workloads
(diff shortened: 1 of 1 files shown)


pgoyal01:
@BernardTsai-DT The intent, in my opinion, is the same as "Infrastructure as Code" (IAC).  As you know, the IAC concept allow infrastructure to be treated the same way as application programs -- human readable code specification.  The IAC code allows us to declaratively script the final state of the infrastructure environment; the execution of the IAC code automatically provisions the target environment. The software development process including change management are applicable and hence can adapt the CI/CD framework for infrastructure build, deployment, provisioning, updates, etc.

BernardTsai-DT:
@pgoyal01 I totally agree - would you like to modify the statement or does it already capture the view you have expressed in your comment?

pgoyal01:
@BernardTsai-DT Your definition restricts it to (1) machine readable and (2) on-boarding and maintaining workloads. We can have human readable but yet not be declarative (what but not the how).  Does the following work: "The model should allow for the description of the final state of the NFVI resources for automated life cycle management."

BernardTsai-DT:
I very much like the proposal and have integrated the proposal into chapter 3 design principles.

doc/ref_model/chapters/chapter03.md
```diff
@@ -23,7 +23,7 @@ Although a couple of explicit and implicit abstraction models (e.g. in the conte
 -	**Scope:** the model should describe the most relevant virtualised infrastructure resources (incl. acceleration technologies) an NFVI needs to provide for hosting Telco workloads
 -	**Separation of Concern:** the model should support a clear distinction between the responsibilities related to maintaining the network function virtualisation infrastructure and the responsibilities related to managing the various VNF workloads
 -	**Simplicity:** the amount of different types of resources (including their attributes and relationships amongst one another) should be kept to a minimum to reduce the configuration spectrum which needs to be considered
--	**Declarative:** the model should allow for a declarative description of the required NFVI capabilities for on-boarding and maintaining workloads
+- **Declarative**: the model should allow for the description of the final state of the NFVI resources for automated life cycle management
```
CsatariGergely: It is not only the NFVI resources, but also the NFVI configuration what we would need to describe in a declarative way.

BernardTsai-DT: That is correct - I have included the configuration aspect into this principle

PR #208, opened 2019-08-19
Chapter 1 -  tenant definition

BernardTsai-DT:
Tenant definition focuses on the pooling of resources provided by the NFVI and not on any organisational structures and roles.


rabi-abdel:
to remove duplicates, can we not remove one fo those definitions? ideally, we keep the one in the Chapter 01 and remove the one from Chapter 03 (unless it adds something incremental to the definition in Chapter 1), what do you think @BernardTsai-DT ?

BernardTsai-DT:
@rabi-abdel we are detailing the term tenant in more detail in Chapter 3 - so I would prefer to keep the definitions in both chapters

Issue #203, opened 2019-08-12
VNF/CNF in Chapter 1 refer to #198


bfcohen:
Release 1 MVP is only Infra and VNFs; CNF-s will be partially supported in Release 2
JB - clarify position and look at best way to support/define CNF articulated. (also impacts Ch. 3)+1 (high level only in CP1)
Can this be clearly stated in the document and can we remove the refernces ot CNF-s from the Release 1 branch to avoid confusion? Overall CP1 is missing the use cases, once these are added, much of the confusion should be resolved.

bfcohen:
Refer to #198 for the related issue.

pgoyal01:
Also related to #139

BernardTsai-DT:
Added following text to the Roadmap section of Chapter 1:
(in the branch: VNF/CNF-in-Chapter-1)

Release 1 of this document will focus on the network function virtualisation infrastructure and virtualised network functions which are based on virtual machines. In Release 2 there will be partial support for cloud native functions which make use of container technology.

@bfcohen, @rabi-abdel: Could you please check if the statement is correct?

pgoyal01:
Removing Chapter03 Label for this issue as it pertains only to text in RM Chapter 01.

bfcohen:
Added a roadmap item in Chapter 01 that explicitly demonstrates that Containerization will be started at least in Release 2.

Issue #198, opened 2019-08-07
[Chapter 1] VNF/CNF


BernardTsai-DT:
**Comment from Paris Meeting: Chapter 03 Infrastructure Abstraction - GAPS (Day 1):** 

**VNF/CNF**

VNFs/CNFs are not managing their own resources via API-s.
VNFs and CNFs are using different abstractions.

VNFs are VMs working on an infrastructure (compute, NW, storage) abstraction, CNFs are processes working on an operating system abstraction. We must not pretend VNFs and CNFs were simply different ways of packaging software (CNFs need OS + container runtime + container orchestration on top of NFVI) --> have to keep these layers separate to ensure CNFs are abstracted from infra mgmt. and run on different NFVIs (OpenStack, VMw, AWS, ...).

Release 1 MVP is only Infra and VNFs; CNF-s will be fully supported in Release 2

    Can this be clearly stated in the document and can we remove the refernces ot CNF-s from the Release 1 branch to avoid confusion?

pgoyal01:
@BernardTsai-DT: aren't VMs also processes running in the Hypervisor space on top of an OS (Host OS). Agree that VMs and Containers are "virtualized," organized and orchestrated differently but both VMs and Containers consume NFVI resources and services.

xavier-grall:
NFVI hardware & software can be fully abstracted from VM perspective (unless they directly use some hardware-specific features, eg with pcipassthrough).
Container must share the NFVI nodes kernel, unless they run themselves on a VM or a dedicated Baremetal (eg, in BMaaS mode) but they must still share the kernel with other containers on the same VM/BM.
So the abstraction between VM and containers is well different, so they require different control/management functions at the NFVI level (hypervisor vs container runtime, eg kvm vs docker) and MANO level (VIM vs CISM as per ETSI NFV IFA029, eg openstack vs kubernetes).
Moreover, it should be noticed that kubernetes could also be a VNFM.
Discussions about that in 1 issue #122 and 2 pending PR #123, #193

ulikleber:
Just to clarify a bit the terminology. 
CNF is a type of VNF. VNFs can use VMs or containers. Also VNFs usually are composed of multiple components. See definition in chapter 1.

PR #209, opened 2019-08-19
Position CNFs as Release 2 roadmap item

BernardTsai-DT:
Referring to #203, #198

Describes in section 1.9 how Releases 1 and 2 of the Reference Model will address VNFs and CNFs.


rabi-abdel:
We need a roadmap diagram to compliment this sentence.

karinesevilla:
I thought that the reference model will stay agnostic to VM and containers

BernardTsai-DT:
The reference model should be agnostic but possibly certain aspects of the model might have to be readjusted due to additional aspects which may become of relevance when analysing CNF workloads.

rabiabdel:
Defer to future if needed.

Issue #145, opened 2019-08-01
[Chapter 5] What does exactly support of flavours mean? Can we add a bit of clarification here?


samuelhellec:
good question 

maybe we have to removed that row from the table 5.1?

pgoyal01:
Flavours are artifacts used by consumers of NFVI  resources -- used to create an execution environment instance. But for the execution environment instance to be created actual physical resources with certain capabilities need to exist or another way of saying "support for flavours." I think the intent of the ask here is to elaborate on what "support" means.

samuelhellec:
we agree with your interpretation. So what could "support" mean ?

karinesevilla:
This intent was to ensure the support of the flavours defined in Chapter 4.
It can be removed.

pgoyal01:
Per @karinesevilla close the issue.

Issue #140, opened 2019-08-01
[Chapter 5] How the NFVI software profile is a set of virtual resources 


samuelhellec:
In 5.1
"A NFVI SW profile is a set of virtual resources with specific behaviour, capabilities and metrics."

Needs rewording

karinesevilla:
The sentence is wrong. It can be replaced by 
"A NFVI SW profile is a set of features, capabilities and metrics offered by an NFVI SW layer"

PR #210, opened 2019-08-19
Chapter 5 - Sections 5.1 and 5.2

karinesevilla:
5.1 Introduction, modification referring to issue #140 
5.1.1 and 5.2.1, "support of flavours" removed referring to issue #145 
5.1.3 and 5.2.3, "Horizontal scaling" replaced by " NFVI scaling"

### DIFF
--- a/doc/ref_model/chapters/chapter05.md
+++ b/doc/ref_model/chapters/chapter05.md
@@ -34,7 +34,7 @@ For a host (compute node or physical server), the virtualization layer is an abs
 <p align="center"><img src="../figures/ch05_b_ref_profile.PNG" alt="b_ref_profile" title="Reference Profile" width="70%"/></p>
 <p align="center"><b>Figure 5-2:</b> NFVI- Virtual resources.</p>
 
-Depending on the requirements of VNFs, a VNFC will be deployed with a NFVI instance type and an appropriate compute flavour. A NFVI instance type is defined by a NFVI SW profile and a NFVI HW profile. A NFVI SW profile is a set of virtual resources with specific behaviour, capabilities and metrics. **Figure 5-3** depicts a high level view of software profiles for Basic, Network Intensive and Compute intensive instances types.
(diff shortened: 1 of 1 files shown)


pgoyal01:
@karinesevilla Couple of questions:

Section 5.1: "Depending on the requirements of VNFs, a VNFC will be deployed with a NFVI instance type and an appropriate compute flavour." What is the difference between instance type and compute flavour? 	

As you know, instance type as used by AWS and other public cloud providers differs significantly from the definition here, "A NFVI instance type is defined by a NFVI SW profile and a NFVI HW profile. " Would this cause confusion?

Would a compute flavour be used when creating an instance (VM)?

Figure 5.3: what does "HA" represent? 

Nitpick: CPU ALlocation Ratio is defined as "Number of virtual cores per physical core" -- a CPY Allocation Ratio of 1:4 would then imply 1 vCPU on 4 physical CPUs.

karinesevilla:
@pgoyal01 
1-What is the difference between instance type and compute flavour?
The instance type (basic, network intensive, compute intensive) caracterizes a set of capabilities for a compute environment on top of which applications will run.
The compute flavour will be choosen for a specific VNF-C to fit the application requirement in terms of cpu, memory and  storage ("size").
2- I think the notion of type versus size is also present in AWS offer.
3-Would a compute flavour be used when creating an instance (VM)?
Yes 
4- For me, HA means High Availability in Figure 5.3, but I wasn't the author of the initial figure. By the way, I think that we need to translate it into features in SW profile or to remove it of the figure if we don't find features relevant features on HA.
5- CPU ALlocation Ratio 
 Yes totally agree, It must be changed to 4:1  to fit the definition

pgoyal01:
@karinesevilla Thanks. I think the document needs to make clear the differences between instance type and flavours as this is the first they may encounter both terms.

karinesevilla:
@pgoyal01 
Instance type and compute flavour are defined in chapter 1. If it isn't clear enough, we must modify them. I'll write an issue for chapter 1 to propose modifications.

PR #225, opened 2019-08-25
Update Chapter 03.md

pgoyal01:
Added preliminary content -- still WIP


pgoyal01:
@walterkozlowski Agree about Ceilometer and Prometheus -- would like us to decide on a direction; in Paris Ceilometer was presented as part of the Reference Architecture.

pgoyal01:
Should we specify a base Host OS or Kernel base version?

ASawwaf:
> @walterkozlowski Agree about Ceilometer and Prometheus -- would like us to decide on a direction; in Paris Ceilometer was presented as part of the Reference Architecture.

my recommendation to go with Prometheus ( or similar approaches Sensu , Collectd , telegraf )  as in real deployment ceilometer cannot scale at all

ASawwaf:
> Should we specify a base Host OS or Kernel base version?

i dont think so

rabi-abdel:
@pgoyal01 great work. the one thing I would say this requires a lot of stripping down as the content shown here is more component level than High Level. The focus of this chapter needs to be on the interfaces northbound (towards VNFs) and VIM. high level Networking Fabric.

rabi-abdel:
@pgoyal01 please update prior meeting to only include outline proposal.

rabi-abdel:
DEFERED

Issue #243, opened 2019-09-03
[Chapter 1] Correct Language Describing Figure 1-7 


markshostak:
Description currently reads such that VIM is _not_ included in CNTT's "NFVI" scope. It should read that VIM _is_ included.

bfcohen:
Fixed in new pull request.

PR #253, opened 2019-09-04
[RM Chapter 1] Fix for Issue #243 (Correct Language Describing Figure…

markshostak:
… 1-7)

Revises description which currently reads such that VIM is not included in CNTT's "NFVI" scope. It should read that VIM is included.


karinesevilla:
I agree

Issue #206, opened 2019-08-12
[Chapter 6] APIs and Interfaces


pgoyal01:
Cleanup Chapter 6 content and align with RL-0 and RL-1 models. Any APIs beyond those  are out-of-scope; possible belong to the Reference Architecture External Interfaces Chapter.

rabi-abdel:
PR created: https://github.com/cntt-n/CNTT/pull/257

PR #257, opened 2019-09-04
Update Chapter06.md

pgoyal01:
Cleanup compleetd.

### DIFF
--- a/doc/ref_model/chapters/chapter06.md
+++ b/doc/ref_model/chapters/chapter06.md
@@ -3,44 +3,18 @@
 <p align="right"><img src="../figures/bogo_sdc.png" alt="scope" title="Scope" width="35%"/></p>
 
 ## Table of Contents
-* [6.1 Infra related APIs.](#6.1)
+* [6.1 Introduction.](#6.1)
 * [6.2 NFVI APIs.](#6.2)
   * [6.2.1 Tenant Level APIs.](#6.2.1)
-* [6.3 Supporting Enabler Service APIs (not-MVP).](#6.3)
-  * [6.3.1 NTP, DNS, etc.](#6.3.1)
-  * [6.3.2 Licensing and imaging connectivity.](#6.3.2)
-* [6.4 Hardware Acceleration Interfaces and APIs.](#6.4)
-* [6.5 Tool functionalities needed (not-MVP).](#6.5)
-  * [6.5.1 Categorized (not specifically named).](#6.5.1)
-  * [6.5.2 Policies and Security related primarily.](#6.5.2)
-  * [6.5.3 If embedded in VM.](#6.5.3)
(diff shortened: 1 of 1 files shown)


rabi-abdel:
fixing issue: https://github.com/cntt-n/CNTT/issues/206

rabiabdel:
will be kept open for few more days

markshostak:
Per review w/ Tom and Mark, Enabler Services requirements need to remain in the RM, as they are common to all RAs. Also, strongly recommend retaining chapter introduction. Thanks, Mark & Tom

rabi-abdel:
@pgoyal01 would it be possible please to keep the original text in the introduction and the structure as it was and focus this Pull request on the cleanup of the rest of the document?

tomkivlin:
How about this as a more concise introduction?

> The purpose of this chapter is to define and catalogue a common set of open (not proprietary) APIs, of the following types:
> 
> - **Intra-NFVI APIs**: These APIs are provided and consumed directly by the infrastructure.  These APIs are purely internal to the NFVI and are not exposed to the workloads.
> - **NFVI APIs**: These APIs are provided to the VNF workloads (i.e. exposed), by the infrastructure in order for VNF workloads to access (i.e. request, consume, control, etc.) NFVI resources.
> - **Enabler Services APIs**: These APIs are provided by non-NFVI services and provide capabilities that are required for a majority of VNF workloads, e.g. DHCP, DNS, NTP, DBaaS, etc.

pgoyal01:
@rabi-abdel, @tomkivlin please review updates.

markshostak:
Hi Pankaj, What is the motivation to remove the already agreed upon introduction? Clarifying the language describing the categories of APIs makes sense, but removing previously agreed content such as the intended application of the chapter, its relationship to the RA, reiteration of utilizing open interfaces, etc. introduces ambiguity with no obvious benefit. **Suggest you constrain edits to Tom's clarification of API categories, and retain the rest of the intro.** Thanks, -Mark

pgoyal01:
@markshostak Basically to make the introduction concise and to the point. The new summary was provided by @tomkivlin.

tomkivlin:
I'm happy with the intro to stay as-was until after Antwerp, given we haven't really got time to debate it.  My "minimum" suggestion so to speak would be the intro I proposed but I'm happy to have a longer intro than that.

pgoyal01:
@tomkivlin I think the new Introduction (Thanks to you) succinctly captures the content of the Chapter.  We need some others to review and comment.

rabiabdel:
@pgoyal01 it seems to me the overall conclusion to keep the intro as it was since both @tomkivlin and @markshostak are ok with it. seems to me the best way forward (for now).

would it possible put the original text back so we can approve it and merge it tomorrow.

Issue #264, opened 2019-09-07
[RM Chapter 4] Clarify Network Interface Specifications (Table 4-15)


markshostak:
Paris followup: Tracking PR to resolve note JB-7, 9 and 16 from Ch3 epad.

markshostak:
Committed PR #265 to resolve. Pending review.

markshostak:
Update approved and merged. Closing.
Note: More refinements will be done to the networking virtual interface specification, post-Antwerp.

PR #265, opened 2019-09-07
[RM Chapter 4] Clarify Network Interface Specifications (Table 4-15)

markshostak:
Tracking Issue #264
Edits to resolve notes JB-7, 9 and 16 from Paris Ch3 epad.
This is a first past at reworking Table 4-15, and doing so before Antwerp. The table has been simplified, but currently retains functional compatibility with the original table, as well as its original style. Post-Antwerp, we need to clarify the application of "n", and possibly consider a fully parametric approach.
Current updates agreed to per 4/9/2019 Chapter 2/4/5 review mtg.


markshostak:
Approved per Chapter review meeting.

PR #245, opened 2019-09-03
[Reference Model] Chapter 7 - 7.10 addition of VIM

karinesevilla:
Modification of the title "NFVI and VIM Vendors responsibility"
Addition of §7.10.4

PR #272, opened 2019-09-10
[RM Chapter 7] Sect 7.10.5 Relocation of VIM Sec'y Tables

markshostak:
This commit is a continuation of Issue #245, to relocate Tables 4-9 and 4-14 from Chapter 4 to Chapter 7.


markshostak:
> I think that the content of table 4-9 should be part of section 7.6 Workload security.
Hi Karine, I agree, however it's up to them to figure out how to organize it. Our main objective here, is to move the material. They can integrate it more tightly in their next iteration.

PR #307, opened 2019-09-16
Replace a few American spelling words

collivier:
It replaces flavor by flavour and behavior by behaviour.

It's worth mentioning that they are cases where flavor makes sens even
if en_GB is selected (e.g. the OpenStack resources [1])

[1] https://docs.openstack.org/nova/latest/user/flavors.html


markshostak:
I'm good with the sedits, with the exception of the modifications to the "CODE_OF_CONDUCT". While I don't see any material changes, when it comes to legal documents, I like to get buy-in from the owner. Let's let Jonathan render an opinion.
Thanks, 
-Mark

rabiabdel:
exlude CODE_OF_CONDUCT and will be approved afterwordrs.

Issue #296, opened 2019-09-12
[RM Chapter 5] NUMA description, tab 5-13


karinesevilla:
As agreed during Chapter 5 meeting (09/11), remove reference to BIOS on table 5-13 related to NUMA description
New proposal: "NUMA supported and enabled"

rabiabdel:
PR was created: https://github.com/cntt-n/CNTT/pull/310

karinesevilla:
Approved, modifications merged 09/17

PR #310, opened 2019-09-16
[RM Chapter 5] HW profile, BIOS reference removed from NUMA description

karinesevilla:
Table 5.13: BIOS removed from NUMA description -> #issue296


rabiabdel:
issie: https://github.com/cntt-n/CNTT/issues/296

PR #320, opened 2019-09-17
[RA Ch05] Update Chpater05.md

pgoyal01:
Added content to Chapter 05

### DIFF
--- a/doc/ref_arch/openstack/chapters/chapter05.md
+++ b/doc/ref_arch/openstack/chapters/chapter05.md
@@ -72,6 +72,7 @@ This section illustrates some of the Interfaces provided by OpenStack; the exhau
 
 **APIs**
 | OpenStack Service | Link for API list | API Baseline Version |
+|------------------|--------------------|------------------------|
 | Identity: Keystone | https://docs.openstack.org/api-ref/identity/v3/index.html?expanded=#identity-api-operations  | Version 3.10 |
 | Compute: Nova | https://docs.openstack.org/api-ref/compute/  | Version 2.1 |
 | Networking: Neutron | https://docs.openstack.org/api-ref/network/  | Version 2.0 |
@@ -84,6 +85,7 @@ This section illustrates some of the Interfaces provided by OpenStack; the exhau
 
 **CLIs**
 | OpenStack Service | Link to Client | Client Baseline Version |
+|------------------|--------------------|------------------------|
 | Identity: Keystone | https://docs.openstack.org/python-keystoneclient/latest/using-api-v3.html | Version 3.10 |
 | Compute: Nova | https://docs.openstack.org/python-novaclient/latest/cli/index.html | Version 2.1 |
 | Networking: Neutron | https://docs.openstack.org/python-neutronclient/latest/cli/index.html | Version 2.0 |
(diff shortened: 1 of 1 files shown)


doc/ref_arch/openstack/chapters/chapter05.md
```diff
@@ -5,17 +5,106 @@
...
+The Kubernetes APIs are available at https://kubernetes.io/docs/concepts/overview/kubernetes-api/. 
+
+### 5.3.3. KVM Interfaces
+The KVM APIs are documented in Section 4 of the document https://www.kernel.org/doc/Documentation/virtual/kvm/api.txt. 
+
+#### 5.3.3.1. Libvirt Interfaces
+The Libvirt APIs are documented in https://libvirt.org/html/index.html.
+
+### 5.3.4. vSphere/ESXi APIs
+The ESXI PAIs are documented together with vCenter and available at https://code.vmware.com/apis/62/vcenter-management
```
markshostak: Hi Pankaj,
- Typos: "i" is lower case (ESXi) and PAIs should be APIs
- WRT the link, those APIs are for vCenter/VCSA, _not_ ESXi. Sect 5.3.4 is introducing a new set of APIs that RA#1 does not have the components to support, b/c ESXi is not currently a component that is compatible with / can be interchanged with KVM or the other Linux kernel-based hypervisors. It is not architecturally compatible with the existing RA #1 design, and will require special accommodations/modifications (of RA#1) to deploy. I think this topic needs further discussion of the implications before inclusion in RA#1. Thanks, -Mark

rabiabdel: @markshostak agree. Let us approve and merge for now and i will provide an overall cleanup for it to be discussed Friday prior the release. OK?

pgoyal01: @markshostak @rabiabdel Will correct typos. Using ESXi or vSphere (both natively supported in nova-compute) is a matter of configuring nova-conf and nova networking. The necessary drivers already exist. Please remember that 1. Hyper-V is also supported natively by nova-compute; and 2. the VIM to Hypervisor interface is internal.

markshostak: > @markshostak agree. Let us approve and merge for now and i will provide an overall cleanup for it to be discussed Friday prior the release. OK?

Sounds good, Rabi. Just let me know when. Thx, -Mark

rabi-abdel: It will be provided for this Friday final review prior to the release.

iangardner22: Hi @markshostak @rabi-abdel - since VMWare are part of the CNTT community I guess we should acknowledge that they have a good and very pervasive hypervisor that we can't really ignore...  As @pgoyal01 mentioned it hooks in seamlessly with nova and while, at least in VIO, they have a single large nova host per ESXi cluster it appears to respect scheduler hints, anti-affinity, affinity etc.   If we include VIO we know some features like Ironic won't be available today and also VIO mandates the use of NSX.   I would suggest we go with it and seek feedback but we should ask VMWare to be more active in the RA going forward so we can understand if we have larger gaps.   Also worth noting that from a certification point of view the VNF vendors need to ensure they can work against a slightly different stack but with thorough test automation this shouldn't be a major issue

iangardner22: They key is it needs to slot in as a building block without any special handling / treatment.   Ultimately the operators will decide if they need the additional bells and whistles VMWare brings and the TCO.

Issue #319, opened 2019-09-17
[RA1 Ch03] Move APIs to Ch5 and add reference (link) in ch3.


rabiabdel:

pgoyal01:
@rabi-abdel The APIs had been consolidated and moved to RA-1 CH 05 prior to Antwerp and there is a reference to Chapter 5 in Chapter 4 ("The APIs for these OpenStack services are listed in Chapter 5: Inetrafces and APIs."). 

Can this issue be closed?

pgoyal01:
PR merged.

PR #325, opened 2019-09-17
Update Chapter03.md

pgoyal01:
Issue #319 "Move APIs to Ch5 and add reference (link) in ch3" -- Done.


rabi-abdel:
No need to discuss in a new meeting. this is based on what we agreed during Tuesday 17th Meeting. will approved and merged.

Issue #288, opened 2019-09-11
[RM Chapter 8 Annex] describe and select test frameworks prior to the review process


collivier:
CVC and Dovetail are currently well described compared to the OPNFV test frameworks which implement or aggregate the tests. We could add sections about the OpenStack gates, the main upstream frameworks (Rally, Tempest, etc.) and OPNFV Functest for instance.

I also consider that it should be written somewhere that our reference implementation should meet the upstream verification criteria (see OpenStack Gates). Then we should select the related testcases here from the test frameworks from OPNFV (Functest, YardStick, etc.).
We shouldn’t forget that the current CVC test list selection is very low compared to the OpenStack gates (or Functest smoke and benchmarking).

The link to Dovetail seems unclear as it’s designed to allow external commercial product verification. Why couldn’t the referent implementation verification be done via Jenkins or any other scheduler (see OPNFV Installer daily model)?
I don’t see the point to forward all logs from Functest or YardStick to a reviewer committee when verifying the reference implementation.

mf4716:
@collivier - agree with the issue.  Please generate a PR, and make recommended changes.  We'll review / discuss at the next chapter meeting.

rabi-abdel:
seems resolved since a PR was pushed for it.

PR #335, opened 2019-09-18
[Chapter 8] Add Functest description

collivier:
It partially implements #288


collivier:
WIP: few recommendations has to be added

PR #321, opened 2019-09-17
[RM Ch10] Update chapter10.md

rabi-abdel:
Moving @jbeltranatl Pull Request to use the right file.
Replacement for https://github.com/cntt-n/CNTT/pull/292

PR #340, opened 2019-09-18
Chapter 10 - Initial Verbiage

jbeltranatl:
In addition to content feedback. Needs to be cleaned up with proper formatting.


jbeltranatl:
This is replace by #321

PR #345, opened 2019-09-19
Botrange cleanup

rabi-abdel:


rabiabdel:
@pgoyal01 @iangardner22 @karinesevilla @samuelhellec @xavier-grall @markshostak 
The Majority of change corresponds to the OpenStack Reference Architecture as to try to stick to the L0-L5 concept agreed on in Paris and used across different chapters,  slides, materials, etc. I am worried if we don't stick to it, this will make our deliverables inconsistent. if we all are happy with this PR, we will have a quick meeting tomorrow :)

PR #349, opened 2019-09-20
Update Kubernetes chapter01.md

tomkivlin:
Added some thoughts around Kubernetes in this domain.


rabiabdel:
initial content and thoughts for Antwerp.

doc/ref_arch/kubernetes/chapters/chapter01.md
```diff
@@ -15,6 +15,57 @@
...
+
+There are a number of different approaches that could be adopted:
+1. Software vendor brings Kubernetes
+    - This option places the operator purely as the infrastructure provider, and the software vendor uses the VIM, or IaaS APIs, to manage the lifecycle of the infrastructure required to create, update and delete Kubernetes clusters ***and*** the software being deployed into Kubernetes
+2. Operator brings Kubernetes
+    - This option places the operator as not only the infrastructure provider, but also a "platform-as-a-service" provider (for Kubernetes Platform). The operator may already be providing platform-as-a-service offerings such as database-as-a-service, or other underlying "enabler services" for the software such as DNS, NTP, mail relay, etc.
+    - It is not necessarily the case that the vendor that the operator chooses for Kubernetes is the same vendor that provides the NFVI and/or VIM for that operator. That is a procurement decision for each operator.
+
```
ASawwaf: 3rd option Telcom vendor brings K8s , right now all Telco vendor like Huawei , E/// , Cisco has their own Container platform whick based on K8s ( Huawei has Fusion stage ) , E// ( Cloud Container Distribution (CCD) ) and Cisco has Cisco Container Platform 
Which lead us to the same dilemma that shall we ( Operators ) go with Sw vendor such as RH OCP , Pivtol , .. or  go with Telco vendor solution !!!

doc/ref_arch/kubernetes/chapters/chapter01.md
```diff
@@ -15,6 +15,57 @@
...
+- Should each operator decide how Kubernetes is shared (if at all)?  Should we/someone suggest some best practice (such as, link Kubernetes cluster lifecycle to application lifecycle - so VNFv1.0 is in one cluster, VNFv2.0 is in another cluster. Are clusters shared between VNF vendors, or VNF types?). 
+
+A note on virtualised and containerised workloads:
+It is highly likely that a single VNFM will be managing applications (or components) that are virtualised (i.e. running in VMs) and applications (or components) that are containerised (i.e. running in containers), at the same time.  There are different approaches to achieving this:
+1. Application manager (EMS, VNFM) uses both IaaS (VIM) API and Kubernetes API - the former for VM based workloads, the latter for containerised workloads.
+2. Application manager (EMS, VNFM) uses just the Kubernetes API for both VM-based and container-based workloads, with the Kubernetes PaaS managing the lifecycle of VMs using on of the following methods:
+    a) Kubernetes interacts with IaaS/VIM API (Cluster API model)
+    b) Kubernetes is the IaaS/VIM and interacts direct with the hypervisor (Kubevirt model)
+
```
ASawwaf: we need to add the deployment mode,  K8S over IaaS ( Openstack ) or K8s over BareMetal

doc/ref_arch/kubernetes/chapters/chapter01.md
tomkivlin: This is the 1st option, the software vendor is the telecom vendor.

doc/ref_arch/kubernetes/chapters/chapter01.md
tomkivlin: See the later section about bare metal considerations. The general model is the same, even with bare metal we need an IaaS API it's just for a Physical Infra Manager instead of VIM.

doc/ref_arch/kubernetes/chapters/chapter01.md
ASawwaf: SW vendors are different than the Telco vendor , we as opertaors need to avoid vendor lock as we did in NFV ( VIM ) that we dont like to have the VIM from the same vendor of VNF

So the SW is CNF/VNF agnostic such as RH , Vmware , ... but the Telco vendor that there own CNF

doc/ref_arch/kubernetes/chapters/chapter01.md
ASawwaf: Good  , but when we mention IaaS doesn't mean BareMetal , so I preferred to mention is a BareMetal that exposed open APIs 
what did u think @tomkivlin

tomkivlin: More than happy to call it out, will do so when at my desk next. Not sure I agree that IaaS != BareMetal though, agree we can make this clearer here though.

pgoyal01:
Comments on 1.2 Terminology:
Container Image: Please add "Note: that the package still requires a container runtime and operating system."

Container: an infrastructure-agnostic (physical or virtual) machine for the delivery of software (applications).  One or more (closely related) applications and storage volumes can be deployed in a container

Pod: a logical deployment unit (abstract resource) for applications. A pod is the smallest unit that can be created and managed by kubernetes. A pod represents one or more containers and shared resources. The containers and shared resources within a pod share the pod IP Address and its port space.

Cluster: an abstraction that represents a set of nodes (worker machines) and a master.  

Master: is responsible for managing the cluster and coordinates all activities in a cluster.

Node: a kubernetes worker machine (physical or virtual) that can host one or more pods (viz., running application. 

Kubernetes Service: is an abstraction to expose applications running on a set of pods. Service is a unit of consumption as it allows applications to be accessible from the outside.

Service mesh: a method for reliable microservice-to-microservice communications where the control governing the communications is built into the microservice.

pgoyal01:
Comments on Terminology 1.2:
we should generalize for example, not reference vendors such as "Docker"

Issue #355, opened 2019-10-01
[RA2 Ch01] Initial Content for K8s Based Reference Architecture 


rabi-abdel:
Based on discussion on Antwerp, the decision was to start a PR to initiate the content for Chapter 1 to talk about the introduction and to start thinking about what is needed from platform prospectives to deliver on the cloud native definition.

rabi-abdel:
As a greed on Antwerp, I will go a head and create the initial PR for the discussion on Tuesday next week.

tomkivlin:
Can this be closed now @rabi-abdel ?

rabiabdel:
Yes

PR #383, opened 2019-10-04
Initial Content for K8s RA

rabi-abdel:
Resolving Issue: https://github.com/cntt-n/CNTT/issues/355

As agreed in Antwerp, we will start by looking at the definition of Could native and what it means to Platform. it might worth spending sometime defining terminology to remove confusion. I expect this PR to stay open for a while.


rabi-abdel:
Let us start with some terminology work. @tomkivlin @TamasZsiros @CsatariGergely can you help providing some terminology definitions here? also please tag others you know who also can help with this PR.

tomkivlin:
Suggest we start from NIST and add anything as required: https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-145.pdf

I also suggest we try not to get involved in (re-)defining cloud native software terminology, but rather focus on the cloud platform / infrastructure (hence starting with NIST).

TamasZsiros:
Sounds like a sound proposal, probably we'll have to add "CaaS" though.
Also we might want to include the CNF definition from CNCF TUG work:
"A Cloud Native Virtual Network Function - or CNF for short - is an implementation of a Virtual Network Function (as defined by ETSI GS NFV 003 V1.4.1 (2018-08) that adheres to the CNCF Cloud Native Definition. "

pgoyal01:
@tomkivlin @TamasZsiros The NIST document definitions the generic cloud technology terms and are not targeted to cloud native; belong to RM Chapter 1. May I suggest that for RA-2 Chapter 1, we may want to define the objectives, terms that are specific to cloud native design principles and certain cloud native technologies (kubernetes, containers, pods, etc.). 

The CNF definition in RM Chapter 1 should be aligned with the definition by @TamasZsiros.

pgoyal01:
@rabi-abdel May I suggest a slight reorganization of the Chapter -- more in alignment with RM and Ra-1 Chapter 1's? 
1.0 Introduction (What, Why and Context)
1.1 Cloud Native Principles
1.2 Scope of RA-2
2.0 Terms (specific to RA-2 not previously introduced in RM Ch 1)
2.1
.....

rabi-abdel:
> @rabi-abdel May I suggest a slight reorganization of the Chapter -- more in alignment with RM and Ra-1 Chapter 1's?
> 1.0 Introduction (What, Why and Context)
> 1.1 Cloud Native Principles
> 1.2 Scope of RA-2
> 2.0 Terms (specific to RA-2 not previously introduced in RM Ch 1)
> 2.1
> .....

Yes I am OK with that. let me try to re-arrange.

rabi-abdel:
@tomkivlin are you suggesting to copy those terminolgies from the NIST into this document? and introduce any news ones we need to include?

tomkivlin:
@rabi-abdel I was, but if they are already used in RM chapter 1 then we should point to that, as per @pgoyal01 suggestion.  

I disagree that the NIST definition isn't targeted to cloud native but I do agree we need to call out the capabilities we think a cloud platform need in order to support cloud native software in addition to what NIST specifies. i.e. what is required from the cloud itself to enable immutable infrastructure, such as closed loop orchestration of infrastructure resources, etc.

pgoyal01:
@tomkivlin Should these topics be part of "Chapter 2 Requirements for Cloud Native Cloud Platform" to cover @tomkivlin "we need to call out the capabilities we think a cloud platform need in order to support cloud native software" including "what is required from the cloud itself to enable immutable infrastructure, such as closed loop orchestration of infrastructure resources, etc."

ASawwaf:
Just for brainstorming, I put down some keyword that should be covered in RA of K8s in terms of chapter topics, Architecture, Terminologies 


o	Cloud Native definition 
o	Micro severs definition
o	Service Mesh definition
o	CaaS and PaaS definition
o	Namespace , Cgroups definition and usage 
o	Helm Charts , Kubevirt definition and usage 
o	Container infrastructure Manager ( CIM )
o	DevOps and DevOps Toolchain
o	What are key factors that CSP looking for in DevOps
o	Service based Architecture 
o	Data Store 
o	Container Networking 
	Networking requirements 
	SDN Plugin 
	CNI 
•	Calico 
•	Flannel 
•	Multus 
•	Contiv
	Challenges in Container networking 
•	Multiple IPs 
•	Isolation
o	Deployment model CaaS / IaaS , CaaS/BM 
o	Acceleration in Container infrastructure
	OVS
	Fd.io 
o	Challenges in Container infrastructure readiness ( what shall Container infrastructure should be ?)
o	Challenges in CNF readiness ( what shall CNF should be ?)
o	Add-On ( beyond infrastructure ) 
	Orchestration of CNF ( ETSI MANO part )
	 API exposed to NB systems

TamasZsiros:
> 
> 
> Just for brainstorming, I put down some keyword that should be covered in RA of K8s in terms of chapter topics, Architecture, Terminologies
> 
> o Cloud Native definition
> o Micro severs definition
> o Service Mesh definition
> o CaaS and PaaS definition
> o Namespace , Cgroups definition and usage
> o Helm Charts , Kubevirt definition and usage
> o Container infrastructure Manager ( CIM )
> o DevOps and DevOps Toolchain
> o What are key factors that CSP looking for in DevOps
> o Service based Architecture
> o Data Store
> o Container Networking
>  Networking requirements
>  SDN Plugin
>  CNI
> • Calico
> • Flannel
> • Multus
> • Contiv
>  Challenges in Container networking
> • Multiple IPs
> • Isolation
> o Deployment model CaaS / IaaS , CaaS/BM
> o Acceleration in Container infrastructure
>  OVS
>  Fd.io
> o Challenges in Container infrastructure readiness ( what shall Container infrastructure should be ?)
> o Challenges in CNF readiness ( what shall CNF should be ?)
> o Add-On ( beyond infrastructure )
>  Orchestration of CNF ( ETSI MANO part )
>  API exposed to NB systems

While I certainly agree that all of these are interesting topics, I think there is value in simplification and doing an RA from  top-down perspective. So I would suggest starting with defining the CaaS layer, with its integration points on the north-bound (management APIs, e.g. K8s & Helm), extention points towards the bottom of the stack (storage and networking plugins), and a high level summary of the mandatory and optional things inside (Host OS, container orchestration and LCM, assurance, logging, etc.) + some surroundings (like infrastructure LCM).

doc/ref_arch/kubernetes/chapters/chapter01.md
```diff
@@ -54,7 +120,7 @@ Figure 1-1 below shows this at a high level.
 <p align="center"><img src="../figures/ch01_k8s_arch.png" alt="Kubernetes Architecture for NFV" title="Kubernetes Architecture for NFV" width="100%"/></p>
 <p align="center"><b>Figure 1-1:</b> Kubernetes Architecture for NFV</p>
 
-### Other thoughts
+### More thoughts
 - Kubernetes / CaaS should be recognised in ETSI NFV / MANO
```
Levovar: IMO it should be made clear from the beginning that CaaS = Kubernetes + lots of backends behind the generic K8s APIs
the two are not synonyms, K8s in itself is not a CaaS IMO

doc/ref_arch/kubernetes/chapters/chapter01.md
```diff
@@ -54,7 +120,7 @@ Figure 1-1 below shows this at a high level.
...
 - It is feasible that Kubernetes, using operators and custom resources, could become a generic VNFM (i.e. the Kubernetes parts are the underlying engine, with the operators and custom resources being the equivalent of the specific VNFM).
```
Levovar: you mean VIM? :)

doc/ref_arch/kubernetes/chapters/chapter01.md
Levovar: and CaaS is the thing which should be recognized as a VIM (or NFVI?) by ETSI

doc/ref_arch/kubernetes/chapters/chapter01.md
```diff
@@ -5,16 +5,82 @@
...
+
+
+<a name="1.7"></a>
+## 1.7 Roadmap
+
+
+## General thoughts
+
 This document (at this stage, not a RA, we're just using the template as a placeholder), discusses the role of Kubernetes in managing NFV software components (e.g. VNFCs). Application containers such as those managed by Kubernetes can be considered a Foundational PaaS capability<sup>1</sup> and so the following terms can all be thought of as being the topic of this document:
 - Kubernetes PaaS (Platform-as-a-Service)
```
Levovar: refer to next comment: definition of CaaS and PaaS is not consistent throughout the document
I propose starting the terminology chapter with these definition
Kubernetes: a tool to orchestrate containers
CaaS: Containers-as-a-Service layer providing container orchestration, and must have services to be able to manage CNFs at scale
practically it is K8s + minimal set of backends serving K8s abstract APIs
PaaS:  additional optional services on top of CaaS, providing value added centralized, and re-usable services for micro-service based CNFs requiring them
e.g. service mesh etc.

doc/ref_arch/kubernetes/chapters/chapter01.md
```diff
@@ -5,16 +5,82 @@
...
+>CNCF Cloud Native Definition v1.0
+>Approved by TOC: 2018-06-11
+
+>“Cloud native technologies empower organizations to build and run **scalable** applications in modern, **dynamic environments** such as public, private, and hybrid clouds. Containers, **service meshes**, **microservices**, **immutable infrastructure**, and **declarative APIs** exemplify this approach.
+
+>These techniques enable **loosely coupled** systems that are **resilient**, **manageable**, and **observable**. Combined with **robust automation**, they allow engineers to make high-impact changes frequently and predictably with minimal toil.
+
+>The Cloud Native Computing Foundation seeks to drive adoption of this paradigm by fostering and sustaining an ecosystem of open source, vendor-neutral projects. We democratize state-of-the-art patterns to make these innovations accessible for everyone”
+
+The definition above is very applicatin centric and look at cloud native from the application point of view, in here, we will try and relate those definitions to the platforms of which cloud native applicatiosn run on top of.
```
CsatariGergely: application

doc/ref_arch/kubernetes/chapters/chapter01.md
CsatariGergely: ETSI NFV MANO recognise Kubernetes CaaS:

- VNFC definition contains containers from the beginning
- Support for pods in Vnfd is progressing in NFVIFA(19)000499 (r6 at the moment, but I'm pretty sure it will increase)
- Several NWI-s were accepted in the topic after IFA029

Also we should follow the same approach with standardization bodies as with open source projects. If we think that something is missing we should go there and fix the issue. I suggest to add a list of missing items to the Gaps chapter with their assigned resources to fix the issue. If we end up with a too long list we might consider to open a liaison connection towards the ISG.

doc/ref_arch/kubernetes/chapters/chapter01.md
CsatariGergely: A generic VNFM which is specific to the VIM is not really generic...

doc/ref_arch/kubernetes/chapters/chapter01.md
tomkivlin: Agree that CaaS = Kubernetes + other things.  I think we also need to define a **CaaS Manager** which is a management plane function that manages the lifecycle (instantiation, scaling, healing, etc.) of Kubernetes clusters, including communication with VIM for master/worker node lifecycle management.

doc/ref_arch/kubernetes/chapters/chapter01.md
tomkivlin: How would it be specific to the VIM?  Not sure I understand?

tomkivlin: No - I mean VNFM :)

doc/ref_arch/kubernetes/chapters/chapter01.md
Levovar: that would be the NFVO, right?
not necessarily specific for CaaS, IMO we need such orchestration/federation/POD LCM capabilities for all clouds, regardless type
I would say we simply need a CaaS-aware orchestrator

doc/ref_arch/kubernetes/chapters/chapter01.md
CsatariGergely: Kubernetes operators will work only on Kubernetes, while in ideal case a VNFM supports OpenStack, Kubernetes and maybe even vCloud...

tomkivlin: Ah I see - my thoughts were that those VNFM operators could make use of the Cluster API providers to talk to multiple VIMs.  I don't see Kubernetes as a VIM.

CsatariGergely: In my head it is more a VIM/NFVI than any other thing in the ETSI NFV MANO stack, however I agree, that it implements some VNFM features and lacks some VIM/NFVI features. At the end Kubernetes is the component which orchestrates the containers, which are the pieces of infrastructure resources the applications are requesting.

fzdarsky:
Agree with @TamasZsiros, if we want the K8s RA specification to be a) robust and still apply as technology evolves and b) stay aligned with the wider ecosystem than just telco, we need to work top-down or rather "start minimal, add only if absolutely needed."

For example, start with a "plain" k8s (say, v1.16) feature set for a fully cloud-native CNF, provide clear evidence that a functional requirement cannot be met by that system (say, multi-NIC support) and only then add the least invasive, k8s-community aligned extension (say, multus) to fill this gap. Then iterate.

As general best-practice, I hope we agree to focus on specifying the WHAT (interfaces and functionality of the "black box") not the HOW (internal implementation of the "black box"). Obviously, we'll want to give guidance on K8s deployment profiles, but this must not impact the "interface specification" or we're doing it wrong.

doc/ref_arch/kubernetes/chapters/chapter01.md
tomkivlin: Interesting, I would have put the container as part of the application, rather than an infrastructure resource, but I see where you're coming from.

deebhatia: When we say generic VNFM, does it mean something like [this](https://openbaton.github.io/documentation/vnfm-generic/) running over kubernetes?

doc/ref_arch/kubernetes/chapters/chapter01.md
TamasZsiros: It is important to differentiate between LCM for workloads executed on top of CaaS instance(s) (CNFs) - which should be the NFVO and the LCM for the CaaS instances themselves, which should be separated from the NFVI in the fuctional architecture (similarly to how it is described in the RM).

So NFVO manages the workloads, CaaS manager manages the CaaS (and/or?) IaaS instances.

fzdarsky: Yes, a CaaS is more than K8s. If you make the analogy to Linux, K8s is the kernel, but you need way more components in a Linux distribution.
We need to be more specific on what a "CaaS" is, though, because it's often misinterpreted as "platform to run containers" when in fact it also handles the full application lifecycle management --> architecturally, it sits _above_ the NFVI/VIM layer and goes up to the NFVO layer (it performs VNFM functionality).

doc/ref_arch/kubernetes/chapters/chapter01.md
TamasZsiros: I'm wondering, whether structure wise it would make sense to handle the RA for the PaaS separately compared to the CaaS. This would limit complexity in the work, and the PaaS would have to be run on top of a CaaS anyway (at least according to the definition above, i.e. when PaaS does not CONTAIN the CaaS.

doc/ref_arch/kubernetes/chapters/chapter01.md
TamasZsiros: I don't see K8s becoming the new VNFM, TBH it rather reduces the need to have VNFM in general. The key is to make it mandatory to rely on K8s for the LCM of the CNF in terms of upgrade, healing, etc. Then operators kind-of become a neccesity for stateful parts (databases). But the point with K8s is that the VNFM is not needed, or even if it's needed, a much thinner VNFM is necessary.

doc/ref_arch/kubernetes/chapters/chapter01.md
Levovar: I'm fully onboard with that.
PaaS isn't necessarily part of the cloud infrastructure itself: it can be brought by an application, an independent vendor, multiple can co-exist on top of the same CaaS etc.

doc/ref_arch/kubernetes/chapters/chapter01.md
```diff
@@ -54,7 +120,7 @@ Figure 1-1 below shows this at a high level.
 <p align="center"><img src="../figures/ch01_k8s_arch.png" alt="Kubernetes Architecture for NFV" title="Kubernetes Architecture for NFV" width="100%"/></p>
 <p align="center"><b>Figure 1-1:</b> Kubernetes Architecture for NFV</p>
 
-### Other thoughts
+### More thoughts
 - Kubernetes / CaaS should be recognised in ETSI NFV / MANO
 - It is feasible that Kubernetes, using operators and custom resources, could become a generic VNFM (i.e. the Kubernetes parts are the underlying engine, with the operators and custom resources being the equivalent of the specific VNFM).
 - Should each operator decide how Kubernetes is shared (if at all)?  Should we/someone suggest some best practice (such as, link Kubernetes cluster lifecycle to application lifecycle - so VNFv1.0 is in one cluster, VNFv2.0 is in another cluster. Are clusters shared between VNF vendors, or VNF types?).
```
TamasZsiros: Having a K8s instalce per VNF is a bit of an overkill from overhead perspective. (Perhaps a necessary overkill for some, but still). I believe a reasonable recommendation could be a K8s cluster per vendor, or perhaps one per security domain. 

K8s domains could be shared between VNF vendors, but I feel that is out of the comfort zone of many.

doc/ref_arch/kubernetes/chapters/chapter01.md
Levovar: > So NFVO manages the workloads, CaaS manager manages the CaaS (and/or?) IaaS instances.

makes sense! only point I would add is that the two could be possibly collocated on the same physical entity in real life, even if we define them separately (seen such approach in open source, hence the comment).
and then maybe we should just call it "cloud manager"
and maybe it should manage multiple clouds not just one. especially because a big use-case of CaaS IMO will be the edge, and we shouldn't ask operators to manage 1000s of sites one-by-one (regardless if it is automated, or manual management)

doc/ref_arch/kubernetes/chapters/chapter01.md
fzdarsky: K8s sits _on top of_ the NFVI/VIM, because it consumes infrastructure (machines, networks, disks).

It _does_ provide VNFM functionality. And, yes, k8s operators are k8s-specific. At the end of the day, NS-level service+resource orchestrators will have to deal with multiple deployment targets, whether application level (k8s, swarm, ...) or infra level (OpenStack, GCE, ...), via adapters. A common abstraction layer across all these won't fly, just like it didn't above the NFVI/VIM layer.

doc/ref_arch/kubernetes/chapters/chapter01.md
TamasZsiros: Agreed. There are pros and cons to both approaches. I would like to differentiate between "Infrastructure LCM" and "Workload LCM", because these functions exercise different APIs in different timeframes. Then, whether a given management system (implementation) deploys both together or separately should be a strictly implementation specific question.

doc/ref_arch/kubernetes/chapters/chapter01.md
fzdarsky: In general, there will be multiple k8s clusters for operational reasons, e.g. limiting the failure domain or providing (hard) multi-tenancy. Whether a SP runs different vendors' VNFs in the same or segregated clusters is a matter of policy. I don't think it's necessary for us to define that.

tomkivlin: > In general, there will be multiple k8s clusters for operational reasons, e.g. limiting the failure domain or providing (hard) multi-tenancy. Whether a SP runs different vendors' VNFs in the same or segregated clusters is a matter of policy. I don't think it's necessary for us to define that.

Totally agree.  We need to address the need for at least one Kubernetes clusters managed by a **CaaS Manager** (multi-cluster management plane). I reckon that has similar requirements on virtualised resource management as is already defined for VNFM in ETSI, and then similar requirements on Kubernetes cluster LCM as VNFM has for VNF LCM in ETSI.

ASawwaf:
Gents, it is very interesting discussion, but we need to organize it ( as I lost :) ) 
so my recommendation, to start and agree with all terminology to be on same page 
2nd , we discussed the difference between CaaS manager and VNF manager  and LCM for both that should be covered 
3rd : the limitation  that native k8s has 
4th :  CNF deployment on CaaS ( Single . Multiple pod )

tomkivlin:
I am going to try and put all comments into the doc and we can tweak from  there.

tomkivlin:
All - I've added most if not all the definitions discussed above.  The main things that need agreement are:
1. Is CaaS Manager part of NFVO (my view is no, it must be separate - e.g. OpenShift, PKS, Rancher, Loodse, etc. are all CaaS Managers that have their own API to manage the lifecycle of CaaS platforms.  A CaaS platform being a k8s cluster plus all the add-ons.
2. Does k8s sit in VIM, VNFM or somewhere in between - my view is somewhere in between.  It does workload LCM, on instruction from the VNFM. The CaaS manager does infra LCM on instruction from the NFVO.

Levovar:
> 1. Is CaaS Manager part of NFVO (my view is no, it must be separate - e.g. OpenShift, PKS, Rancher, Loodse, etc. are all CaaS Managers that have their own API to manage the lifecycle of CaaS platforms.  A CaaS platform being a k8s cluster plus all the add-ons.

I would argue the Cloud Manager would be something which manages the life-cycle of multiple e.g. Openshift clusters :)
IMO managing the life-cycle -upgrade- of a local cloud stack is kind of a basic NFVI requirement/feature. Sure, the NFVI can and prob should provide LCM related APIs for a centralized manager entity, but at least in my terminology -I might be alone with this ofc :) - the "Cloud Manager" is the one invoking

tomkivlin:
Yes sorry I mean the CaaS Manager invokes the VIM to manage the LCM of infra.  

My point is that each CaaS platform (k8s cluster + add-ons) should be able to be defined in a "descriptor" and that should include things like:
- Location, size, capabilities of the infra (so CaaS Manager can instruct VIM)
- Version of k8s
- Included add-ons and versions of them
- etc.

Maybe we call it a CaaS Descriptor or something in line with other ETSI NFV terms?

doc/ref_arch/kubernetes/chapters/chapter01.md
```diff
@@ -5,21 +5,91 @@
...
+
+
+<a name="1.2"></a>
+### 1.2 Terminology
+
+|Term|Definition|Reference|
+|---|---|---|
+|Container Image|A container image is a lightweight, standalone, executable package of software that includes everything needed to run an application: code, runtime, system tools, system libraries and settings.|[Link](https://www.docker.com/resources/what-container)|
+|Container|A running instantiation of a container image.|-|
+|Pod|A Pod (as in a pod of whales or pea pod) is a group of one or more containers (such as Docker containers), with shared storage/network, and a specification for how to run the containers. A Pod’s contents are always co-located and co-scheduled, and run in a shared context.|[Link](https://kubernetes.io/docs/concepts/workloads/pods/pod/)|
```
CsatariGergely: I think we should add that pod is the atomic unit of LCM in Kubernetes and co-located in the same host.

doc/ref_arch/kubernetes/chapters/chapter01.md
```diff
@@ -5,21 +5,91 @@
...
+|Cluster|A cluster consists of at least one cluster master and multiple worker machines called nodes. These master and node machines run the Kubernetes cluster orchestration system.|[Link](https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-architecture)|
```
CsatariGergely: We should clarify that this is a Kubernetes cluster.

CsatariGergely: We should use the definition form the Kubernetes documentation. @TamasZsiros please post the link.

bfcohen:
General comment about the definitions section.   Once the definitions are finalized, they need to be moved to a glossary, except for a few that are essential to the understanding of the rest of the document.  Per previous agreement, the definitions should be as much as possible, based on existing definitions.  Need to reconcile K8S, docker and CNCF definitions.  If there is a difference, K8S should be considered the master.

doc/ref_arch/kubernetes/chapters/chapter01.md
```diff
@@ -5,21 +5,91 @@
...
+|Master|The cluster master runs the Kubernetes control plane processes, including the Kubernetes API server, scheduler, and core resource controllers.|[Link](https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-architecture)|
+|Node|Nodes are the worker machines (physical or virtual) that run your containerised applications and other workloads. Each node is managed from the master, which receives updates on each node's self-reported status. A node runs the services necessary to support the Docker containers that make up your cluster's workloads. These include the Docker runtime and the Kubernetes node agent (kubelet) which communicates with the master and is responsible for starting and running Docker containers scheduled on that node.|[Link](https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-architecture)|
```
CsatariGergely: - Lets use "container runtime" instead of Docker. 
- Should be formulated in passive mode. 
- kubelet is responsible for the whole LCM of the container runtimes.

doc/ref_arch/kubernetes/chapters/chapter01.md
```diff
@@ -5,21 +5,91 @@
...
+|PaaS|The capability provided to the consumer is to deploy onto the cloud infrastructure consumer-created or acquired applications created using programming languages, libraries, services, and tools supported by the provider. The consumer does not manage or control the underlying cloud infrastructure including network, servers, operating systems, or storage, but has control over the deployed applications and possibly configuration settings for the application-hosting environment.|[NIST SP 800-145, The NIST Definition of Cloud Computing](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-145.pdf)|
+|CaaS|Container-as-a-Service, a foundational application PaaS capability, in this case delivered as a Kubernetes cluster and other technologies such as CNI, Ingress, Service Mesh, LMA, etc. CaaS sits above VIM/NFVI and performs some application LCM based on instructions received from VNFM.|[Foundational PaaS Definition](http://www.theenterprisearchitect.eu/blog/categorize-compare-cloud-vendors/#row5column3)|
```
CsatariGergely: - CaaS is not part of PaaS 
- CaaS can run on bare metal, not only on top of a VIM/NFVI
- From where are the list of these technologies are comming from?
@tomkivlin can you please clarify these?

pgoyal01:
@tomkivlin Can you please add a NOte to "Container Image", "Note: that the package still requires a container runtime and operating system."

pgoyal01:
@tomkivlin Can you please make changes to the definitions including some of the content as in below:

Container: an infrastructure-agnostic (physical or virtual) machine for the delivery of software (applications). One or more (closely related) applications and storage volumes can be deployed in a container

Pod: a logical deployment unit (abstract resource) for applications. A pod is the smallest unit that can be created and managed by kubernetes. A pod represents one or more containers and shared resources. The containers and shared resources within a pod share the pod IP Address and its port space.

Kubernetes Service: is an abstraction to expose applications running on a set of pods. Service is a unit of consumption as it allows applications to be accessible from the outside.

Service mesh: a method for reliable microservice-to-microservice communications where the control governing the communications is built into the microservice.

pgoyal01:
@tomkivlin we should be vendor agnostic in our definitions. For example, instead of "Docker containers" just "containers" and instead of "Docker runtime" use "container runtime."

doc/ref_arch/kubernetes/chapters/chapter01.md
CsatariGergely: CaaS functionality of Kubernetes which is absolutely necessarry to LCM containers, while PaaS functionality of Kubernetes is everything on top of it.

bfcohen:
Make sure that the introduction notes that the requirements for this reference architecture is that we are only defining the MINIMUM amount of infrastructure that is needed to support containers.

doc/ref_arch/kubernetes/chapters/chapter01.md
```diff
@@ -5,21 +5,91 @@
...
+|CaaS Manager|A management plane function that manages the lifecycle (instantiation, scaling, healing, etc.) of one or more Kubernetes clusters, including communication with VIM for master/node lifecycle management.|-|
```
CsatariGergely: [link](https://docs.google.com/document/d/1-zqxz5bdCLTuOEvi2ybADR3PcmzbBhNt6YkNnvx-KoA/edit#heading=h.5x0d5h95i329)

doc/ref_arch/kubernetes/chapters/chapter01.md
```diff
@@ -5,21 +5,91 @@
...
+|CNF|A Cloud Native Virtual Network Function - or CNF for short - is an implementation of a Virtual Network Function (as defined by ETSI GS NFV 003 V1.4.1 (2018-08) that adheres to the CNCF Cloud Native Definition.|CNCF TUG (link tbc)|
```
CsatariGergely: [link](https://docs.google.com/document/d/1-zqxz5bdCLTuOEvi2ybADR3PcmzbBhNt6YkNnvx-KoA/edit#heading=h.5x0d5h95i329)

ASawwaf:
> Yes sorry I mean the CaaS Manager invokes the VIM to manage the LCM of infra.
> 
> My point is that each CaaS platform (k8s cluster + add-ons) should be able to be defined in a "descriptor" and that should include things like:
> 
> * Location, size, capabilities of the infra (so CaaS Manager can instruct VIM)
> * Version of k8s
> * Included add-ons and versions of them
> * etc.
> 
> Maybe we call it a CaaS Descriptor or something in line with other ETSI NFV terms?

@

> All - I've added most if not all the definitions discussed above. The main things that need agreement are:
> 
> 1. Is CaaS Manager part of NFVO (my view is no, it must be separate - e.g. OpenShift, PKS, Rancher, Loodse, etc. are all CaaS Managers that have their own API to manage the lifecycle of CaaS platforms.  A CaaS platform being a k8s cluster plus all the add-ons.
> 2. Does k8s sit in VIM, VNFM or somewhere in between - my view is somewhere in between.  It does workload LCM, on instruction from the VNFM. The CaaS manager does infra LCM on instruction from the NFVO.

@tomkivlin , valid question, from my POV , CaaS manager not be part of NFVO ( but if we need to be ETSI aligned we can put it part of MANO ) 
For 2nd point: I don't like to mention k8s as standalone, i like to have CaaS which is K8s ++ ,, in that case, VNF-M ( CNF-M) do the LCM ++ ( ++ means day 0 and maybe day 1 configuration ) for workload while the CaaS manager does infra LCM on

doc/ref_arch/kubernetes/chapters/chapter01.md
pgoyal01: To resolve this and also discussion that follows later on CaaS, Paas, Iaas, etc. would it be worth creating a couple of high level diagrams for discussion purposes?

doc/ref_arch/kubernetes/chapters/chapter01.md
kedmison: One of the benefits of K8S and Helm is increased granularity over upgrades along with canary testing, etc, so there would absolutely be a drive to have VNF 1.0 reside in one cluster and have it gradually transformed, via granular K8S operations, into VNF 2.0.

kedmison: > In general, there will be multiple k8s clusters for operational reasons, e.g. limiting the failure domain or providing (hard) multi-tenancy. Whether a SP runs different vendors' VNFs in the same or segregated clusters is a matter of policy. I don't think it's necessary for us to define that.

I think we should highlight some of these reasons for multiple K8S clusters, as the technical issues behind these reasons will be addressed over time. One example is the security partitioning: while we may choose to use NFVi VMs as a resource boundary, CRIs are moving to incorporate significantly stronger security partitioning that will remove a major reason for multiple K8S clusters.

doc/ref_arch/kubernetes/chapters/chapter01.md
kedmison: Where would you position Helm in this environment?  Mandated? Optional?  Would you consider it taking on the role of a lighter-weight VNFM?

doc/ref_arch/kubernetes/chapters/chapter01.md
walterkozlowski: PaaS vs CaaS was discussed at length at the meeting on Tuesday.  I would like to point people to the following article:
https://containerjournal.com/features/paas-vs-caas-wrong-question-ask/
I would agree with the author of this article (except his references to Docker, which we want to avoid).  In essence the article definitions are close to what was proposed by Gergely.  The article provides short definitions:
 1. PaaS – supports development & deployment but it happens on predefined platform.
 2. CaaS – supports packing & deployment but it allows for personalized platform.
We could work out our definitions based on these ideas, I suppose. If we agree to such ideas, I would suggest us limiting our K8s RA to CaaS, as we were inclined in the meeting; this is because I do not think that we want to start with any development frameworks (PaaS).  Any thoughts?

doc/ref_arch/kubernetes/chapters/chapter01.md
ASawwaf: > Where would you position Helm in this environment? Mandated? Optional? Would you consider it taking on the role of a lighter-weight VNFM?

@kedmison, I will go with it should be part of the CNF ( VNF) package onboarding task, so either from VNF-M or NFVO

doc/ref_arch/kubernetes/chapters/chapter01.md
```diff
@@ -5,21 +5,91 @@
...
-This document (at this stage, not a RA, we're just using the template as a placeholder), discusses the role of Kubernetes in managing NFV software components (e.g. VNFCs). Application containers such as those managed by Kubernetes can be considered a Foundational PaaS capability<sup>1</sup> and so the following terms can all be thought of as being the topic of this document:
-- Kubernetes PaaS (Platform-as-a-Service)
-- CaaS (Containers-as-a-Service)
-- Kubernetes-based CaaS (Containers-as-a-Service)
-- Kubernetes-as-a-Service
+> to be written
+
+
+<a name="1.2"></a>
+### 1.2 Terminology
```
fzdarsky: Ensuring consistency between terms is hard. IMO, it helps taking them from a single source rather than mixing them from several sources. The k8s glossary contains many of the terms (container, cluster, pod, ...) we need:
https://kubernetes.io/ja/docs/reference/glossary/?fundamental=true

doc/ref_arch/kubernetes/chapters/chapter01.md
fzdarsky: To address his, we could add the half sentence following the linked definition, too:
"A Pod models an application-specific “logical host” - it contains one or more application containers which are relatively tightly coupled."
An alternative would be to quote the definition in the glossary:
https://kubernetes.io/ja/docs/reference/glossary/?fundamental=true#term-pod

doc/ref_arch/kubernetes/chapters/chapter01.md
fzdarsky: https://kubernetes.io/ja/docs/reference/glossary/?fundamental=true#term-cluster

doc/ref_arch/kubernetes/chapters/chapter01.md
fzdarsky: https://kubernetes.io/ja/docs/reference/glossary/?fundamental=true#term-node

I agree we should keep this technology neutral ("container runtime" instead of "Docker runtime" and "OS containers" instead of "Docker containers"), as it applies to any CRI/OCI compliant implementation. In this context, it'd be more correct to talk about pods instead of containers anyway.

doc/ref_arch/kubernetes/chapters/chapter01.md
CsatariGergely: @ASawwaf in my view it is more the capability of the VIM to manage complex applications and not part of the VNFM or the NFVO.

doc/ref_arch/kubernetes/chapters/chapter01.md
fzdarsky: I'd avoid the term PaaS as it's overloaded by now, because "platform" is too unspecific (platform for what?). Originally, it denoted app development platforms like Heroku and Google App Engine with their specific development tools, APIs, etc.

We should only focus on CaaS, but clarify CaaSes don't just run+LCM containers, but complex application stacks, too.

doc/ref_arch/kubernetes/chapters/chapter01.md
```diff
@@ -5,21 +5,91 @@
...
+|CNF|A Cloud Native Virtual Network Function - or CNF for short - is an implementation of a Virtual Network Function (as defined by ETSI GS NFV 003 V1.4.1 (2018-08) that adheres to the CNCF Cloud Native Definition.|CNCF TUG (link tbc)|
+|Service|A Service in Kubernetes is an abstraction which defines a logical set of Pods and a policy by which to access them. Services enable a loose coupling between dependent Pods.|[Link](https://kubernetes.io/docs/tutorials/kubernetes-basics/expose/expose-intro/)|
+|Service Mesh|A service mesh is a dedicated infrastructure layer for handling service-to-service communication. It’s responsible for the reliable delivery of requests through the complex topology of services that comprise a modern, cloud native application. In practice, the service mesh is typically implemented as an array of lightweight network proxies that are deployed alongside application code, without the application needing to be aware.|[Link](https://buoyant.io/2017/04/25/whats-a-service-mesh-and-why-do-i-need-one/)|
```
fzdarsky: Let's keep terminology sparse, adding terms only when needed. Service Mesh (esp. in the k8s sense) is not an obvious one.

doc/ref_arch/kubernetes/chapters/chapter01.md
```diff
@@ -30,42 +100,18 @@ Kubernetes itself is a “system for automating deployment, scaling, and managem
...
 <p align="center"><img src="../figures/ch01_k8s_arch.png" alt="Kubernetes Architecture for NFV" title="Kubernetes Architecture for NFV" width="100%"/></p>
 <p align="center"><b>Figure 1-1:</b> Kubernetes Architecture for NFV</p>
 
-### Other thoughts
-- Kubernetes / CaaS should be recognised in ETSI NFV / MANO
-- It is feasible that Kubernetes, using operators and custom resources, could become a generic VNFM (i.e. the Kubernetes parts are the underlying engine, with the operators and custom resources being the equivalent of the specific VNFM).
-- Should each operator decide how Kubernetes is shared (if at all)?  Should we/someone suggest some best practice (such as, link Kubernetes cluster lifecycle to application lifecycle - so VNFv1.0 is in one cluster, VNFv2.0 is in another cluster. Are clusters shared between VNF vendors, or VNF types?).
-
 ### A note on virtualised and containerised workloads
 It is highly likely that a single VNFM will be managing applications (or components) that are virtualised (i.e. running in VMs) and applications (or components) that are containerised (i.e. running in containers), at the same time.  There are different approaches to achieving this:
```
fzdarsky: I'm wondering what let's us conclude that "it is highly likely that a single VNFM" will be managing both? I don't see a necessity / advantage to this nor does this seem to be matching the direction in the industry currently?

doc/ref_arch/kubernetes/chapters/chapter01.md
```diff
@@ -30,42 +100,18 @@ Kubernetes itself is a “system for automating deployment, scaling, and managem
...
 1. Application manager (EMS, VNFM) uses both IaaS (VIM) API and Kubernetes API - the former for VM based workloads, the latter for containerised workloads.
 2. Application manager (EMS, VNFM) uses just the Kubernetes API for both VM-based and container-based workloads, with the Kubernetes PaaS managing the lifecycle of VMs using on of the following methods:
     a) Kubernetes interacts with IaaS/VIM API (Cluster API model)
     b) Kubernetes is the IaaS/VIM and interacts direct with the hypervisor (Kubevirt model)
-    
+
 ### A note on bare metal containerisation
 As described above, Kubernetes is an application manager and therefore the lifecycle of Kubernetes clusters should closely match the lifecyle of the application or applications being deployed into Kubernetes. It therefore goes that the engineering teams defining the lifecycle of their applications will also define the lifecycle of the clusters they use for those applications. In an operator this might be a network engineering team, for example.
```
fzdarsky: Wouldn't the "engineering teams defining the lifecycle of their application" be the application vendor's team and the team "defining the lifecycle of the clusters" be the service provider's ops team?

doc/ref_arch/kubernetes/chapters/chapter01.md
TamasZsiros: > 
> 
> > Where would you position Helm in this environment? Mandated? Optional? Would you consider it taking on the role of a lighter-weight VNFM?
> 
> @kedmison, I will go with it should be part of the CNF ( VNF) package onboarding task, so either from VNF-M or NFVO

Part of the VIM (CISM is the proper term I believe). Also part of the CaaS (there is a need for a package manager to allow an abstraction that corresponds to "an application").

doc/ref_arch/kubernetes/chapters/chapter01.md
```diff
@@ -78,33 +124,3 @@ This has the following considerations when it comes to bare metal:
 Regarding the size of the nodes, this is about trying not to reduce the benefits we gained with NFV in the first place. For example, if three hypervisor hosts have 60CPUs and 1.5TB RAM between them, sharing those resources between a large number of applications (lets say 50) is made efficient by the use of virtualisation. For bare metal containerisation, that same number of applications would need a much larger number of bare metal hosts, with each host being a much smaller unit.
```
fzdarsky: Both containers in VMs and on bare metal have their use cases. That the number of application would generally need to be higher for bare metal hosts is not true; it depends on the resource consumption of the workloads, obviously. Virtualization has benefits for running multiple clusters on the same infra, but drawbacks (e.g. resource overhead and fragmentation) when running a single cluster. --> it all depends on the use case.

doc/ref_arch/kubernetes/chapters/chapter01.md
fzdarsky: > Where would you position Helm in this environment? Mandated? Optional? Would you consider it taking on the role of a lighter-weight VNFM?

Arguably, Helm shouldn't be positioned at all here. How to do package management is an implementation detail. It's as (non-)important as discussing "should we agree on apt + deb-packages or yum + rpm-packages".

It will also evolve over time. Helm was last year's darling, this year it's kustomizations and operators, next year something else.

doc/ref_arch/kubernetes/chapters/chapter01.md
TamasZsiros: I tend to agree with @fzdarsky. Yes, K8s is more than just a CaaS, however for simplicity's sake define CaaS as a container orchestration engine + bare minimums that is necessary to deploy a telco application. So in practice CaaS is:
- a K8s distro
- a Host OS
- a package manager (Helm)
- minimum set of standard networking and storage plugins

doc/ref_arch/kubernetes/chapters/chapter01.md
```diff
@@ -30,42 +100,18 @@ Kubernetes itself is a “system for automating deployment, scaling, and managem
...
-
-There are a number of different approaches that could be adopted:
-1. Software vendor brings Kubernetes
-    - This option places the operator purely as the infrastructure provider, and the software vendor uses the VIM, or IaaS APIs, to manage the lifecycle of the infrastructure required to create, update and delete Kubernetes clusters ***and*** the software being deployed into Kubernetes
-2. Operator brings Kubernetes
-    - This option places the operator as not only the infrastructure provider, but also a "platform-as-a-service" provider (for Kubernetes Platform). The operator may already be providing platform-as-a-service offerings such as database-as-a-service, or other underlying "enabler services" for the software such as DNS, NTP, mail relay, etc.
```
TamasZsiros: Maybe replace PaaS provider with CaaS provider?

doc/ref_arch/kubernetes/chapters/chapter01.md
Levovar: isn't CNTT ultimately for nailing down implementation details?
BTW can you show me a CaaS infra which does not currently include Helm?

ok, let's not say Helm explicitly, but I would argue a "package manager" is a must have CaaS component

doc/ref_arch/kubernetes/chapters/chapter01.md
Levovar: I would propose adding some FCAPS to scope as well, at least logging and performance data management (still bare minimum)

Reasoning for logging: one could argue that the "cloud-native" app design for logging is just pushing things to standard out and error, and let the runtime handle it.
Logging related configuration resides in the container runtime (driver, rotation etc.), application has no control over it. So from this perspective at least gathering logs of the standard channels, and forwarding them somewhere for further processing or storing I would argue is very much within CaaS responsibility. Note: I'm not saying central collection, and analysis is, only gathering and forwarding

PM: basic core metric collection and store is actually done automatically by Kubelet, so it would be hard to pretend CaaS does not do PM :)
moreover K8s provides native API for scaling based on metrics. putting PM responsibility into PaaS layer would make "native" K8s horizontal scaling feature dependent on an optional PaaS layer

doc/ref_arch/kubernetes/chapters/chapter01.md
Levovar: +1
I would even argue the number of hosts required to run the same number of applications is usually smaller if you do it on bare metal, compared to running them on top of a CaaS in VMs
IMO the RA can define a CaaS with all the necessary features, and in a separate chapter / paper we could mention the many (at least two, but let's keep in mind we might need to have a "public cloud" discussion at some point :) ) ways of how one can deploy this reference CaaS

doc/ref_arch/kubernetes/chapters/chapter01.md
kedmison: I agree re: nailing down implementation details.  Sometimes that is at RA level, and sometimes at RI level.  I think we have to specify the package manager as part of the RA; CNF developers need to know how to package their software in order to successfully deliver to Implementations that meet this Architecture.

fzdarsky: I strongly disagree. A golden rule of standardization (and this is a standardization exercise at the end of the day!) is to specify external behavior and explicitly _not_ prescribe implementations (and to know where _not_ to standardize). The reason is that you typically don't gain anything with respect to interoperability (because you're addressing a symptom, not a cause) but you risk stifling innovation and competition.

Levovar: do you disagree with explicitly naming a package manager implementation in the RA, or with the statement that package management is required in CaaS?

tomkivlin:
I am ***loving*** the engagement on this. I'm not in the office today or tomorrow morning so will aim to respond and update tomorrow afternoon.

doc/ref_arch/kubernetes/chapters/chapter01.md
fzdarsky: > BTW can you show me a CaaS infra which does not currently include Helm?

That's a trick question :)
All k8s distros _support_ Helm, because it runs on any k8s.
Few _include_ it (e.g. OpenShift, GKE, AKS, Weaveworks, etc. don't), because it's trivial to install on top if someone wants it.
None of them would make itself _dependent_ on it.

The more important question is which tool has the most momentum with the app developer and operations communities. And this is where Helm lost momentum versus other tools in my observation. For example:
* "kustomize" seems popular with those working with a devops (esp. gitops) methodology, are deploying mainly stateless, componentized "web service" kind of workloads, and don't want/need a package manager in the loop. That Google integrated it into kubectl will likely make it the de facto standard in that space.
* "operators" seem popular with those working with complex, often stateful workloads that are very common in the enterprise/telco space. These workloads are notoriously hard to 
operate and what you want is to bottle/"package" this operational knowledge somehow. This is where Helm's package management approach has little to offer, forcing Chart developers to use awkward, hard to maintain workarounds.

fzdarsky: > do you disagree with explicitly naming a package manager implementation in the RA, or with the statement that package management is required in CaaS?

I wouldn't say a package manager is _required_ and in fact the kustomize model (result of the deep analysis of app config management in [1]) shows how one can do pretty well without.
But my disagreement was with picking a single PM tool and declaring it "the standard" by adding it to the RA, especially when done without analysis of alternatives and when it might lead to a disconnect between telco and the direction of the wider community.

[1] https://docs.google.com/document/d/1cLPGweVEYrVqQvBLJg6sxV-TrE5Rm2MNOBA_cxZP2WU/edit

kedmison: That's a great document, and I'll have to go through it a few times to really internalize all the content.  My high-level reactions to it are:
1) we need to define the package management pattern for CNFs to use, whether it be this model, Helm, or something else, if only so that CNF vendors know what to build, test, and expect from K8S clusters, and so that we can build automation pipelines, from CNF vendor through to production in operator networks, that are well tested and work consistently and reliably.
2) Our selection process should be weighted both on what is popular in the industry (i.e. has a well-supported ecosystem around it) as well as its intrinsic technical merits.

doc/ref_arch/kubernetes/chapters/chapter01.md
CsatariGergely: > 
> 
> To address his, we could add the half sentence following the linked definition, too:
> "A Pod models an application-specific “logical host” - it contains one or more application containers which are relatively tightly coupled."
> An alternative would be to quote the definition in the glossary:
> https://kubernetes.io/ja/docs/reference/glossary/?fundamental=true#term-pod

I agree, but I would rather link the [English documentation page](https://kubernetes.io/docs/reference/glossary/?fundamental=true#term-pod) than the Japanese...

doc/ref_arch/kubernetes/chapters/chapter01.md
```diff
@@ -30,42 +100,18 @@ Kubernetes itself is a “system for automating deployment, scaling, and managem
...
-- Option 2, operator brings Kubernetes to meet model for other "as-a-Service" components that are required such as DNS, NTP, etc.
-
-Figure 1-1 below shows this at a high level.
+Figure 1-1 below shows how Kubernetes, CaaS, and the CaaS Manager might fit within the current ETSI architecture.  In terms of this Reference Architecture, we should focus on the requirements of the new interfaces and the capabilities within the CaaS Manager and the Kubernetes clusters.
 
 <p align="center"><img src="../figures/ch01_k8s_arch.png" alt="Kubernetes Architecture for NFV" title="Kubernetes Architecture for NFV" width="100%"/></p>
 <p align="center"><b>Figure 1-1:</b> Kubernetes Architecture for NFV</p>
 
-### Other thoughts
-- Kubernetes / CaaS should be recognised in ETSI NFV / MANO
```
TamasZsiros: It is recognized. IFA029 says:
Container Infrastructure Service: Service that provides runtime environment for one or more container virtualization technologies.
NOTE 4:              Container Infrastructure Service can run on top of a bare metal or hypervisor-based virtualization

Container Infrastructure Service Management: a function that manages one or more Container Infrastructure Services.
NOTE 5:              It provides mechanisms for lifecycle management of the containers, which are hosting application components as services or functions.

Container Infrastructure Service Instance: an instance providing runtime execution environment for container.
EXAMPLE 2:     In Kubernetes® this is a node, which consists of three components: container runtime, kubelet and kube-proxy.

doc/ref_arch/kubernetes/chapters/chapter01.md
pgoyal01: we need to add, "A pod is the smallest unit that can be created and managed by kubernetes." We may want to clearly state that a pod is a logical deployment unit.

doc/ref_arch/kubernetes/chapters/chapter01.md
pgoyal01: May I suggest that we also tie the "node" to "pods" as in "A node hosts one or more pods.

doc/ref_arch/kubernetes/chapters/chapter01.md
pgoyal01: May I suggest that we add the following note to "Service" -- "Service is a unit of consumption as it makes applications externally accessible."

tomkivlin: > May I suggest that we add the following note to "Service" -- "Service is a unit of consumption as it makes applications externally accessible."

That's not always true though (if you mean external to k8s) - e.g. ClusterIP is a type of Service that only makes applications accessible within a cluster.

Levovar: also headless Services don't even have ClusterIPs :)
in which case a Service is for service discovery, but not for traffic routing / loadbalancing

doc/ref_arch/kubernetes/chapters/chapter01.md
tomkivlin: I have updated the terminology for CaaS, removed PaaS, but haven't added package management as a constituent part of the CaaS.

doc/ref_arch/kubernetes/chapters/chapter01.md
tomkivlin: I've removed PaaS from the terminology.

doc/ref_arch/kubernetes/chapters/chapter01.md
tomkivlin: Updated.

doc/ref_arch/kubernetes/chapters/chapter01.md
tomkivlin: Thanks - I've removed this part of the document, will add references to ETSI docs in due course.

doc/ref_arch/kubernetes/chapters/chapter01.md
tomkivlin: Given this passage in the test has been removed, I'll resolve this conversation.

doc/ref_arch/kubernetes/chapters/chapter01.md
tomkivlin: I've used the Kubernetes glossary.  Will resolve this conversation.

doc/ref_arch/kubernetes/chapters/chapter01.md
tomkivlin: I've used the Kubernetes glossary. Will resolve this conversation.

doc/ref_arch/kubernetes/chapters/chapter01.md
tomkivlin: I've removed PaaS and hopefully clarified what constitutes and CaaS and the CaaS Manager.  Will resolve this conversation - can raise new conversations based on latest commits.

doc/ref_arch/kubernetes/chapters/chapter01.md
tomkivlin: Yes I was a little OTT there.  I have changed "highly likely" to "possible".  In any case I think this section will probably get removed, but have left it for now until I get time to add a good deal more content.

doc/ref_arch/kubernetes/chapters/chapter01.md
tomkivlin: I have removed PaaS and simplified the CaaS and CaaS Manager definitions.  Resolving this - any new comments can be on the latest commit.

doc/ref_arch/kubernetes/chapters/chapter01.md
tomkivlin: Thank you - will update to reflect this.

doc/ref_arch/kubernetes/chapters/chapter01.md
tomkivlin: Sure - I will add this into the document.

doc/ref_arch/kubernetes/chapters/chapter01.md
pgoyal01: Minor correction: ETSI GR NFV-IFA 029 V0.14.0 (2019-03) no longer numbers the notes or the example.

CsatariGergely:
We have really good discussions here, but we should 1) start merging content 2) identify issues to work on 3) assign persons and 4) work. 

Would it be okay for everyone, especially to @tomkivlin , if we identify issues what we will not resolve in this pr and we create issues to resolve them later?

As far as I see these more difficult issues are: 

1. Scope of one Kubernetes cluster. One VNF, group of VNF from the same vendor, group of VNF-s. The management of Kubernetes clusters and if this is part of the NFVO
2. Is Helm in or out.
3. Place of Kubernetes in the ETSI NFV MANO stack

Any comments?

TamasZsiros:
> We have really good discussions here, but we should 1) start merging content 2) identify issues to work on 3) assign persons and 4) work.
> 
> Would it be okay for everyone, especially to @tomkivlin , if we identify issues what we will not resolve in this pr and we create issues to resolve them later?
> 
> As far as I see these more difficult issues are:
> 
>     1. Scope of one Kubernetes cluster. One VNF, group of VNF from the same vendor, group of VNF-s. The management of Kubernetes clusters and if this is part of the NFVO
> 
>     2. Is Helm in or out.
> 
>     3. Place of Kubernetes in the ETSI NFV MANO stack
> 
> 
> Any comments?

I really support the idea of merging and lifting important gaps.

For 2 and 3 we need separate issues. For 1 it’s a question of “is this recommendation really necessary” and more importantly: can we give a blanket statement understanding that the answer might depend a lot on operator policies (security) and operational model?

pgoyal01:
@TamasZsiros Agree with your "is this recommendation really necessary" for #1." Not only will the answer "depend upon operator policies" but would also depend upon the nature of the VNFs/CNFs and whether some VNFs/CNFs need to be chained together to create an offered network service.

rabi-abdel:
@CsatariGergely: @tomkivlin is on holiday, would you mind please consolidate feedback above into the PR while @tomkivlin is on holiday? to make sure we don't loose momentum here. let me know if this is OK.

tomkivlin:
@rabi-abdel I think all the feedback that has consensus has been applied and the PR is ready to merge as-is and then we can create issues to deal with the gaps that need filling. Do you agree @csatarigergely?

pgoyal01:
@tomkivlin "CaaS: Container-as-a-Service. A complete set of technologies to enable the management of containerised software, including a Kubernetes cluster, container networking, storage, routing, service mesh, etc." Does CaaS include the IaaS (e.g., NFVI + software (VIM))?

tomkivlin:
@pgoyal01 no, they are separate capabilities

pgoyal01:
@tomkivlin I am confused as "CaaS Manager: A management plane function that manages the lifecycle (instantiation, scaling, healing, etc.) of one or more CaaS instances, including communication with VIM for master/node lifecycle management. **CaaS Manager** could be considered the "**undercloud**" where the **CaaS** is the "**overcloud**"."

ASawwaf:
@tomkivlin @pgoyal01  CaaS can be on IaaS ( one option)

rabi-abdel:
OK it makes sense to merge it then but that requires 4 approvals: may I ask @CsatariGergely @TamasZsiros @pgoyal01 and @tomkivlin as WSL to approve it?

once this is done, It will be automatically merged.

pgoyal01:
@rabi-abdel OK -- we can open issues against the merged version

CsatariGergely:
> For 2 and 3 we need separate issues. For 1 it’s a question of “is this recommendation really necessary” and more importantly: can we give a blanket statement understanding that the answer might depend a lot on operator policies (security) and operational model? I don't think so, and perhaps we should just say that deploying a separate cluster per X (X being vendor, CNF whatnot) is the highest form of separation with the capabilities of K8s as of today, and it will be an operator decision based on policy and op. model.

I'm not sure is a recommendation is needed for it, I'm tending towads the "not needed" side, but from this thread it looks like, that we do not have an consensus on a "no". Let me create an issue about it what we can always close with a nop.

CsatariGergely:
>  we can create issues to deal with the gaps that need filling. Do you agree @CsatariGergely?

I've created the issues about the open issues: 

- #449 
- #450 
- #451 

Please indicate if something is missing.

fzdarsky:
Could we please also add an issue to fix the section "On bare metal containerization"?
It is written with a very obvious bias towards promoting a containers-on-virtualization model and stigmatizing the containers-on-baremetal model as foresaking "the benefits we gained with NFV in the first place" with technically incorrect or one-sided arguments. We agreed both models are equally important.

CsatariGergely:
> 
> 
> Could we please also add an issue to fix the section "On bare metal containerization"?
> It is written with a very obvious bias towards promoting a containers-on-virtualization model and stigmatizing the containers-on-baremetal model as foresaking "the benefits we gained with NFV in the first place" with technically incorrect or one-sided arguments. We agreed both models are equally important.

#453 added (I edit this to the comment about the list of open issues)

PR #434, opened 2019-10-18
Test setup clarification

### LINKED ISSUE #428: [RM Ch8]: Test setup clarification is needed
e.nfvi.per.met.001, 002 and 003: TST009 defines a set of test setups on which the different benchmarks are measured. It should be stated in the definition if these are for p2p, pvp or pvvp. (And in case of containers pcp, pvcvp, pvvcp or pvcvcp).

CsatariGergely:
This change adds a new attribute which enables the definition of the test
setup used to measure e.nfvi.per.met.001, e.nfvi.per.met.002 and e.nfvi.per.met.003.

Fixes #428 .


rabiabdel:
@mf4716 @kagreenwell please approve so it can be merged.

rabiabdel:
@Gaoweitao please approve

markshostak:
@rabi-abdel Al Morton isn't showing up as a potential reviewer. Can you send him an invite (or whatever is needed to prime that pump). If I need to ask him to take an action other than just confirming the email link, pls let me know. Thx, -M

xavier-grall:
This information about the path across the NFVI that the network perf tests have to consider is quite relevant, but it should not be in that table which is (and should stay) only a metrics/measurements list.
Regarding the paths themselves:
- Phy2Phy should not be considered since it is not always applicable (eg for sriov) and does not really reflect a NFVI capacity but only a vswitch one (ie without the vPORT/vNIC).
- Acronyms with C (as pvcvp) should not be added since PVP and PVVP are ETSI definitions which apply to VM as well as Container (V is for VNF rather than VM, as already discussed).
- Maybe we should also consider V2V path despite it is not described in ETSI NFV-TST009 since it would allow to measure NFVI capacity to forward East/West traffic in a better way than only PVVP.

Consequently, I would suggest to "only" add the note text after the table for providing precisions about the different paths to consider for network perf tests, removing Phy2Phy and xCx, and possibly adding V2V (?).
Note that we may need to also add precisions about other perf network test configuration, such as, for instance the traffic processing functions to consider: vlan, vxlan, anti-spoofing, filtering, nat...  , filtering, nat...

xavier-grall:
PS: I am also currently preparing a change for that section

rabiabdel:
> @rabi-abdel Al Morton isn't showing up as a potential reviewer. Can you send him an invite (or whatever is needed to prime that pump). If I need to ask him to take an action other than just confirming the email link, pls let me know. Thx, -M

Do you know his Github Id?

mf4716:
I've reached out and asked Al.  Will revert back when I confirm Al's ID.

rabi-abdel:
@mf4716 I found it from the WS list. invitation on it's way to him.

mf4716:
Thanks - btw, I just heard back from Al, and he confirmed the same, @acmacm

rabi-abdel:
Due to what agreed on last technical steering meeting, only 2 reviewers are needed now (instead of 4). I am OK to wait for more approvals if you would like to get explicit approval from AI. otherwise, I can merge now. let me know

xavier-grall:
Some updates (including a prepared commit) for that section are proposed in issue #462 as I was writing above (https://github.com/cntt-n/CNTT/pull/434#issuecomment-545417885)

markshostak:
> 
> 
> > @rabi-abdel Al Morton isn't showing up as a potential reviewer. Can you send him an invite (or whatever is needed to prime that pump). If I need to ask him to take an action other than just confirming the email link, pls let me know. Thx, -M
> 
> Do you know his Github Id?

@rabi-abdel , I've asked him what it is. I'll let you know. 
I'm ok w/ this update if Xavier is ok w/ it...
Thx, -M

rabiabdel:
@xavier-grall are you OK with Approving and merging this PR?

xavier-grall:
@rabiabdel @CsatariGergely 
Imo, it cannot be merged as i was commenting above: https://github.com/cntt-n/CNTT/pull/434#issuecomment-545417495
The opinion of Gergely regarding that comment would be interesting.
Note also that I have a commit prepared for that section which notably consider that PR proposal (test path topology) : https://github.com/xavier-grall/CNTT/commit/f30e9de72115e83c644842fe494c1b2f8a22d7ae

CsatariGergely:
> 
> 
> This information about the path across the NFVI that the network perf tests have to consider is quite relevant, but it should not be in that table which is (and should stay) only a metrics/measurements list.
> Regarding the paths themselves:
> 
>     * Phy2Phy should not be considered since it is not always applicable (eg for sriov) and does not really reflect a NFVI capacity but only a vswitch one (ie without the vPORT/vNIC).
> 
>     * Acronyms with C (as pvcvp) should not be added since PVP and PVVP are ETSI definitions which apply to VM as well as Container (V is for VNF rather than VM, as already discussed).
> 
>     * Maybe we should also consider V2V path despite it is not described in ETSI NFV-TST009 since it would allow to measure NFVI capacity to forward East/West traffic in a better way than only PVVP.
> 
> 
> Consequently, I would suggest to "only" add the note text after the table for providing precisions about the different paths to consider for network perf tests, removing Phy2Phy and xCx, and possibly adding V2V (?).
> Note that we may need to also add precisions about other perf network test configuration, such as, for instance the traffic processing functions to consider: vlan, vxlan, anti-spoofing, filtering, nat... , filtering, nat...

I think the definition of what we are measuring should go together with the metrics list and should be in this same table. Please propose alternatives what we can discuss, if you see a better one. 
It should be part of the model definition which path setups are relevant a which are not. 

- A agree, that Phy2Phy is not really relevant as it is not used in case of SR-IOV or in case of bare metal containers. Should I remove it from the note?
- Ah, okay, I was not aware of this conclusion. I can remove the c-s.
- V2V assumes that the two VNF-s are colocated on the same host which is rarely the case. 

I agree, that this solution is not perfect, but I believe that the description of the setup should be mapped to the measurements in a way. If we keep only the note than we are not able to make this mapping.

CsatariGergely:
@xavier-grall Okay, now I see that the pr you are [working on](https://github.com/xavier-grall/CNTT/commit/f30e9de72115e83c644842fe494c1b2f8a22d7ae) addresses this issue in a better way than this pr.

What do you think about proposing it as a pr and continue to work on that instead of my vesion here?

acmacm:
@CsatariGergely   I support your proposal to look at @xavier-grall [change](https://github.com/xavier-grall/CNTT/commit/f30e9de72115e83c644842fe494c1b2f8a22d7ae) instead. I made some comments there.

CsatariGergely:
As it was discussed we will go on with @xavier-grall -s approach in a different pr and close this one (and I mistakenly rebased a lots of stuff to here anyways (what I do not know how to clean up)).

Issue #457, opened 2019-10-25
[RA2 Core]: Chapter structure for RA2 core


CsatariGergely:
Propose a chapter structure for RA2.

tomkivlin:
PR458 merged.

Issue #445, opened 2019-10-22
[RA2 - Ch2] Requirements


pgoyal01:
Modify req.gen.ost.01 and req.gen.ost.02 to be req.gen.k8st.01 and req.gen.k8s.02 respectively. In same PR add an initial set of other requirements.

tomkivlin:
From 31.10 meeting: Contributors to the PR (#452) feedback to contact Pankaj with availability to have a meeting to discuss each in line.

pgoyal01:
@tomkivlin Can we close as the issues were taken care of.

tomkivlin:
> @tomkivlin Can we close as the issues were taken care of.

Once #452 is merged, yes - will be automatic if you used "Fixes #445" in the PR comments.

PR #458, opened 2019-10-25
[RA2 Core]: ToC proposal for Ch3 and Ch4

CsatariGergely:
This change contains an open ended proposal for the structure of
Ch3 and Ch4.

### DIFF
--- a/doc/ref_arch/kubernetes/chapters/chapter03.md
+++ b/doc/ref_arch/kubernetes/chapters/chapter03.md
@@ -9,7 +9,8 @@
 * [3.2 Infrastructure Services](#3.2)
     * [3.2.1 Container Runtime Services](#3.2.1)
     * [3.2.2 Container Networking Services](#3.2.2)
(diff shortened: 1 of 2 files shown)


CsatariGergely:
Fixes #457

pgoyal01:
@CsatariGergely Where would you discuss the  physical resources and k8s cluster considerations?

CsatariGergely:
> 
> 
> @CsatariGergely Where would you discuss the physical resources and k8s cluster considerations?

In Chapter 2 we should sort out the exact requirements of these. I saw lots of cluster LCM requirements in #445  but as I remember we agreed in Antwerp not to include these into the RA and we should focus on the interface towards the VNF. 

If we need to sort these out I would put generic subchapters to Chapter 3 to discuss the base of the component selection and subchapters to Chapter 4 for the component selections.

tomkivlin:
Should we have a 3.4 Container Storage Services, to capture the architecture of the storage that might be provided from an NFVI to a CaaS/CNF?

Issue #462, opened 2019-10-25
[RM ch8] Updates for perf measurement (section 8.9.1.1)


xavier-grall:
Perf metrics has been moved from chapter 4 to chapter 8 in a new perf measurement section which requires to be updated to detail how perf metrics can ne measured.
A commit is proposed for that purpose:
https://github.com/xavier-grall/CNTT/commit/f30e9de72115e83c644842fe494c1b2f8a22d7ae

markshostak:
Just to clarify, performance measurements that are taken with external instrumentation (e.g., Spirent, Ixia, etc.) and used to characterize the infra, were moved to RM CH-8. Performance measurements (PMs) that are generated by the infra itself on a day to day basis, remain in RM CH-4. Also, be aware that the measurements in RM CH-8 will likely move to RC NFVI.

xavier-grall:
@markshostak 
OK thank you Mark for that clarification.

xavier-grall:
PR #464 for that issue is replaced by PR #579

kedmison:
The most recent PR #579 was merged, and the perf metrics moved to the RC document

@mf4716 @ASawwaf let me know if you think this issue can be closed.

ASawwaf:
@kedmison , for the time being no issue
I preferred to have a details list of metric either in RM ( Annex ) or RA and RC refer to it

kedmison:
closing this issue, as linked content appears to have been merged to other documents (RC) via other PRs ( #579 )

PR #464, opened 2019-10-28
[RM ch08] 8.9-perf-measurement

xavier-grall:
That PR addresses issues #462 (also including issue #428) and replacing PR #434

### DIFF
--- a/doc/ref_model/chapters/chapter08.md
+++ b/doc/ref_model/chapters/chapter08.md
@@ -874,8 +874,8 @@ They should allow to validate the performance of any software and/or hardware NF
 
 Of course, they can also be used for other purposes, as:
 
-- fine tuning of software and/or hardware NFVI configuration (e.g., the number of cores dedicated to the dpdk vswitch)
-- comparison of the performance of different software or hardware technologies (e.g., dpdk vswitch vs hardware-offloaded vswitch)
+- fine tuning of software and/or hardware NFVI configuration (e.g., the number of cores dedicated to the DPDK vSwitch)
+- comparison of the performance of different software or hardware technologies (e.g., DPDK vSwitch vs hardware-offloaded vSwitch)
 - assessing the performance impact of specific features (e.g., with or without encapsulation) 
 
 
--- a/doc/ref_model/chapters/chapter08.md
+++ b/doc/ref_model/chapters/chapter08.md
@@ -858,69 +858,61 @@ For additional insight, or deeper understanding and reading of IOPS, refer to th
 This section will be used to plan for future offerings.
 
 <a name="8.9.1"></a>
-### 8.9.1 Performance & Resiliency - Measurements, Testing
-
-
-
-#### 8.9.1.1 Performance Measurements
-
-
+#### 8.9.1.1 NFVI Performance Measurements
 
 **Objectives**
 
-The NFVI performance measurements aim at assessing the performance of a given NFVI implementation on the execution plane (i.e., excluding VIM) by providing it with a set of significative metrics to be measured.
+The NFVI performance measurements aim at assessing the performance of a given NFVI implementation on the execution plane (i.e., excluding VIM) by providing it with a set of significant metrics to be measured.
 
-They should allow to validate the performance of any software and/or hardware NFVI implementation as described in 8.6.6. 
+They should allow validating the performance of any software and/or hardware NFVI implementation as described in section 8.6.6.
 
-Of course, they can also be used for other purposes, as:
-
+Of course, they can also be used for other purposes, such as:
 - fine tuning of software and/or hardware NFVI configuration (e.g., the number of cores dedicated to the DPDK vSwitch)
-- comparison of the performance of different software or hardware technologies (e.g., DPDK vSwitch vs hardware-offloaded vSwitch)
-- assessing the performance impact of specific features (e.g., with or without encapsulation) 
-
+- comparing the performances of different software or hardware technologies (e.g., DPDK vSwitch vs hardware-offloaded vSwitch)
+- assessing the performance impact of specific features (e.g., with or without encapsulation)
 
 
 **Metrics Baseline**
 
-For the purpose of validation, a baseline of the performance metrics is required for being compared with the results of their measurements on the NFVI implementation to be validated.
+For the purpose of validation, a baseline of the performance metrics is required for comparison with the results of their measurements on the NFVI implementation to be validated.
 
-That baseline is a set of threshold values which could be determined  by **measuring the performance metrics on Reference Implementations**.. 
+That baseline is a set of threshold values which could be determined by **measuring the performance metrics on Reference Implementations**.
 
-The validation can then be based on simple pass/fail test results or on a grade (e.g. "class" A, B or C) provided by the combination of pass/fail results for 2 different threshold values of some (or all) metrics.
-
+The validation can then be based on simple pass/fail test results or on a grade (e.g., "class" A, B or C) provided by the combination of pass/fail results for 2 different threshold values of some (or all) metrics.
 
 
 **Metrics Description**
 
-Two metrics categories are considered depending if they are related to either the VNF domain or the NFVI domain itself:
+Two categories of metrics are considered depending on whether they are related to either the VNF domain or the NFVI domain itself:
 
-- Metrics related to the VNF domain are defined from VNF perspective (i.e., per VNFC, per vNIC, per vCPU...) and should interest VNF as well as NFVI actors. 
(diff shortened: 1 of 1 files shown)


xavier-grall:
@CsatariGergely 
"V2V assumes that the two VNF-s are colocated on the same host which is rarely the case."
V2V is rather related to 2 components (VNFC) of a same VNF, which may be less rare.

@acmacm 
"P2P is a baseline or calibration test, usually performed before any test incorporating VNFs."
Right, but only to calibrate the generator/receiver through physical network equipment, and not through any NFVI compute node, isn't it ?
If necessary, what would you suggest to change ?
 
"The set-ups where the the traffic generator and receiver are physically separate from the NFVI and VNF(s) is usually preferred, because it isolates the testing functions from the NFVI and VNFs under test (there are some confounding factors to consider when generating traffic on the host where it is processed and measured), and PVP and PVVP are common deployment scenarios."
Do you mean that V2V and VPV test paths must be removed ?
(in that case, how to measure inter-vnfcs metrics ? should they be considered out of our scope ?)
When a loop function is required in addition to physically separated gen/rec (eg for PVP), is there not also any "confounding factors" ?

"L2 and L3 UDP traffic is sufficient for much of this testing, until the VNFs in the setup require state-full traffic. TCP traffic can't be replayed exchanges, must be a real state machine with realistic RTT for the use case (and that may mean emulating delay)."
Do you mean statefull test traffic should not be considered at all ? even not optionally ?
Indeed, statefull "forwarded" test traffic is now explicitly optional, but it would be mandatory to test NFVI performance for processing L4 statefull functions such as firewalling (eg openstack security groups) or nat (eg openstack floating ip).
 
"VSPERF is strong candidate for the L2 and L3 testing, especially when combined with Functest. Most of the Test traffic generators that VSPERF controls can measure 1-way Latency, so no need to measure RTT, especially for delay variation the 1-way measurements are more useful."
Do you mean that we should replace RTT by 1-way frame transfer (as per ETSI NFV-TST009) ?
1 way latency measurement implying the receiver to be hosted on the NFVI, is there not also any "confounding factors" ?

"Usually, baseline testing starts-out at one flow, and works up over a range of values (100, 500, 1000, 5000) to 10000 as you suggest. Same with packet size, 64, 128 etc. up to the MTU."
Do you mean that we should consider testing different flow numbers and packat sizes (and not only 10k and 300B as proposed for "forwarded" test traffic) ?
In that case, should we consider each of the consequent measured values (throughput, latency...) as reference value for validation purpose ?

acmacm:
> @acmacm
> 
> _"P2P is a baseline or calibration test, usually performed before any test incorporating VNFs."_
> Right, but only to calibrate the generator/receiver through physical network equipment, and not through any NFVI compute node, isn't it ?
The P2P test includes the vSwitch, so it tests for the baseline limits of the infrastructure after choosing the vSwitch from among different possibilities (OVS, OVS w/DPDK, VPP, futureVS). A generator/receiver calibration would test using a loopback cable or other trivial connectivity is a form of calibration - perhaps I should not have used the word calibration // baseline is better. I'm thinking of being able to repeat this test before adding any VNFs to compare with past performance. 
> If necessary, what would you suggest to change ?
I think the PVP and PVVP test setups are mentioned in a following the table?  It's hard to see the code and the comment box at the same time... requires another step!
> 
> _"The set-ups where the the traffic generator and receiver are physically separate from the NFVI and VNF(s) is usually preferred, because it isolates the testing functions from the NFVI and VNFs under test (there are some confounding factors to consider when generating traffic on the host where it is processed and measured), and PVP and PVVP are common deployment scenarios."_
> Do you mean that V2V and VPV test paths must be removed ?
> (in that case, how to measure inter-vnfcs metrics ? should they be considered out of our scope ?)
I think the VNF vendors prefer that we leave the intra-VNF networking alone, but they will speak for themselves on this topic (at some point). Meanwhile, V2V might be useful for finding differences between logical networking strategies or unexpected bottlenecks, but the careful eye on the limitations of virtual test system limitations is needed, and will sometimes be exceeded.
> When a loop function pseudo-VNFC (VM/Container) is required in addition to physically separated gen/rec (eg for PVP), is there not also any "confounding factors" ?
Certainly, additional factors that influence the results are introduced with loop function VNF, but we are able to test the logical interfaces between the vSwitch and the VNF with minimal affect from the loop function VNF.
> 
> _"L2 and L3 UDP traffic is sufficient for much of this testing, until the VNFs in the setup require state-full traffic. TCP traffic can't be replayed exchanges, must be a real state machine with realistic RTT for the use case (and that may mean emulating delay)."_
> Do you mean statefull test traffic should not be considered at all ? even not optionally ?
> Indeed, statefull "forwarded" test traffic is now explicitly optional, but it would be mandatory to test NFVI performance for processing L4 statefull functions such as firewalling (eg openstack security groups) or nat (eg openstack floating ip).
For firewall or other L4 testing "...must be a real state machine with realistic RTT for the use case (and that may mean emulating delay)."
> 
> _"VSPERF is strong candidate for the L2 and L3 testing, especially when combined with Functest. Most of the Test traffic generators that VSPERF controls can measure 1-way Latency, so no need to measure RTT, especially for delay variation the 1-way measurements are more useful."_
> Do you mean that we should replace RTT by 1-way frame transfer (as per ETSI NFV-TST009) ?
Yes, please.
> 1 way latency measurement implying the receiver to be hosted on the NFVI, is there not also any "confounding factors" ?
No, 1-way through the PVP path is a one directional measurement of the NFVI and the VNF through the physical interfaces. From physical traffic generator to physical traffic receiver.
> 
> _"Usually, baseline testing starts-out at one flow, and works up over a range of values (100, 500, 1000, 5000) to 10000 as you suggest. Same with packet size, 64, 128 etc. up to the MTU."_
> Do you mean that we should consider testing different flow numbers and packat sizes (and not only 10k and 300B as proposed for "forwarded" test traffic) ?
Yes please, because the performance will vary significantly over packet size and number of flows.
> In that case, should we consider each of the consequent measured values (throughput, latency...) as reference value for validation purpose ?
This may be where need to automatically run a series of tests, but when we understand the workload profiles and traffic profiles of the VNFs, more targeted tests are possible, I think.  In the end, we want the test to maximally useful to describe the performance of the NFVI in the dimensions we care most about. There's more to learn here, don't get me wrong! I don't have all the answers!

doc/ref_model/chapters/chapter08.md
```diff
@@ -860,50 +860,199 @@ This section will be used to plan for future offerings.
...
 
+**Objectives**
+
+The NFVI performance measurements aim at assessing the performance of a given NFVI implementation on the execution plane (i.e., excluding VIM) by providing it with a set of significative metrics to be measured.
+
+They should allow to validate the performance of any software and/or hardware NFVI implementation as described in 8.6.6. 
+
+Of course, they can also be used for other purposes, as:
+
+- fine tuning of software and/or hardware NFVI configuration (e.g., the number of cores dedicated to the dpdk vswitch)
```
CsatariGergely: nits: DPDK, vSwitch

xavier-grall: ok done thanks

rabi-abdel:
@xavier-grall can you please address @CsatariGergely review request?

xavier-grall:
@acmacm 

OK for one-way latency and stateful traffic: modifications done in the last edition (for latency in "traffic profile" section, for stateful, the note is removed)

Concerning P2P, imho, we should not consider it in the context of an NFVI validation automatization since it is not always applicable (eg for sriov) and when applicable, it would require either to change the vswitch configuration of the deployed NFVI to test or to have a specific vswitch configuration available for that test on deployed NFVI (which seems better to avoid).

Concerning V2V/VPV, NFVI actors may need to provide E/W performance values to VNF actors, so we should keep them as possible topologies (moreover, they are largely used by functest vmtp and shaker).

Concerning frame sizes and flows number, I am quite doubful about the interest and efficiency in multiplying the measurement for the same network metrics in the context of an automatized NFVI validation, so I only add a note on that point.

rabi-abdel:
@xavier-grall do you mind merging this change with the overall cleanup one here: https://github.com/cntt-n/CNTT/pull/579 i.e. add the changes you have in here into the https://github.com/cntt-n/CNTT/pull/579 by committing changes in there. is that OK? this way we can close this one and reflect your proposed changes in here in there.

xavier-grall:
@rabi-abdel 
It is ok for me, but:
- it seems there is not enough approval for that PR (at least in its current version...)
- i do not know how to proceed

rabi-abdel:
> @rabi-abdel
> It is ok for me, but:
> 
> * it seems there is not enough approval for that PR (at least in its current version...)
> * i do not know how to proceed

I suggest if you move your changes into there (to #579) we will bring all needed approvals and resolve any outstanding issues.

rabi-abdel:
since this is moved to the other PR. can we close?

xavier-grall:
of course (done right now)

Issue #378, opened 2019-10-04
[RM Ch04] Remove Measurements Taken by External Instruments


markshostak:
Per analysis and agreement in 2/10/2019 RMCH-4 weekly meeting, remove all measurements (currently identified as metrics), that are taken by external test instrumentation. These measurements are to be inserted into chapter 8 (V&V) by Issue #379. See meeting minutes for further details.

markshostak:
Fixed by PR #520. Pending review.

markshostak:
Fixed by PR @520, which has been merged. Closing. -M

PR #471, opened 2019-10-29
[RM Ch04] Remove PM taken by external instruments

xavier-grall:
Resolving #378 (including sections & tables renumbering)

### DIFF
--- a/doc/ref_model/chapters/chapter04.md
+++ b/doc/ref_model/chapters/chapter04.md
@@ -491,15 +491,10 @@ n100, n200, n300, n400, n500, n600 | N | Y | N
 <p align="center"><b>Table 4-21:</b> Mapping of NFVI Capabilities to Instance Types</p>
 
 <a name="4.2.6"></a>
-### 4.2.6 Instance Performance Measurement Mapping
+### 4.2.6 One stop shop
 
-_**Comment:** To be worked on._
-
-<a name="4.2.7"></a>
-### 4.2.7 One stop shop
-
-<a name="4.2.7.1"></a>
-#### 4.2.7.1 Naming convention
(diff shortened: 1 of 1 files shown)


markshostak:
@xavier-grall Hi Xavier, Thank you for all of the effort you put in this commit, it is obvious you were doing a meticulous job. There is an issue, in that the plan for the PMs, as had been discussed in the last several RMCH-2/4/5 meetings, isn't reflected in the commit. Notably, we're not removing PMs altogether, we're only removing the _specific_ PMs that were relocated to RC/RMCH-8. I have an edit open and was thinking I'd just combine mine with yours, but after careful review, I'm afraid we need to revert. If you'd like to discuss/align, we can set up a time on Thr or Fri, just let me know. Thanks, -Mark

xavier-grall:
@markshostak 
In my opinion, only PMs of section 4.1.3.1 and 4.1.5.1 have to move in chapter 8, since they are related to testing.
And PMs of section 4.1.3.4, which are rather related to monitoring capabilities, should stay in chapter 4.
Also, that section could be renamed "performance **monitoring** capabilities" or " **utilization monitoring** capabilities".

markshostak:
@xavier-grall 
"only PMs of section 4.1.3.1 and 4.1.5.1 have to move in chapter 8" - Agreed, and those are the only tables that moved.
"section could be renamed "performance monitoring capabilities"" - Which section, specifically? The only section I see that qualifies, is Section 4.1.2.3 (using original section numbering).
Thx, -M

xavier-grall:
@markshostak sorry it is 4.1.4.4 (but not 4.1.3.4) that could be renamed

xavier-grall:
replaced by #520

PR #527, opened 2019-10-31
[RI1 Core] Add figure files via upload

fuqiao123:
add typical RI figure for the RI Chpter 2.


markshostak:
@fuqiao123 Looks good. I have a few questions: 
- Regarding the name of the control nodes, as those nodes handle a lot more than just the VIM, I'd suggest a more typical/conventional name (e.g., "Control [Node]")
- Do we have dedicated SDN-C nodes? Why not use the VIM/Control nodes for this function?
- Is a mgt spine required? If it is, does it need to be duplex? Does the mgt leaf need to be duplex?

ASawwaf:
@fuqiao123 @markshostak , I believe that there is no need to define the spine / leaf function, IP Fabric ( Spine / leaf ) architecture should able to carry all types of services (Services, Storage and Management ) and this types of traffic shall be disaggregated by VXLAN ( VTEP IDs)  
So my recommendation to update the diagram with normal fabric as described above 

for 2nd point, for Typical implementation, for sure we need to have dedicated SDN-C, and this should br reflected on both RA , RM

fuqiao1:
> @fuqiao123 Looks good. I have a few questions:
> 
> * Regarding the name of the control nodes, as those nodes handle a lot more than just the VIM, I'd suggest a more typical/conventional name (e.g., "Control [Node]")
> * Do we have dedicated SDN-C nodes? Why not use the VIM/Control nodes for this function?
> * Is a mgt spine required? If it is, does it need to be duplex? Does the mgt leaf need to be duplex?

Hi, Mark. thank you for the comments.
Agree with the suggestion for the Control node. I will change it accordingly.
As far as I know, current SDN-C node requires huge resource, and for multiple reasons, we can't reuse the control nodes. I am not an expert for SDN, so I will let the SDN experts talk at this.
Agree that we do not need separate spine leaf structure for manage/storage/service. This figure is just an example. I do not actually require such feature in the above text. But since it may cause confusion, I will change the figure accordingly.

fuqiao1:
> @fuqiao123 @markshostak , I believe that there is no need to define the spine / leaf function, IP Fabric ( Spine / leaf ) architecture should able to carry all types of services (Services, Storage and Management ) and this types of traffic shall be disaggregated by VXLAN ( VTEP IDs)
> So my recommendation to update the diagram with normal fabric as described above
> 
> for 2nd point, for Typical implementation, for sure we need to have dedicated SDN-C, and this should br reflected on both RA , RM

See the above comments for Mark. I agree we don't need to require spine-leaf. This figure is just an example. I would change the figure accordingly to avoid confusion

fuqiao123:
I will close this pull requst, since this figure is uploaded in a wrong directory. I have just create a new PR #527. Please provide your comments in that PR.

ASawwaf:
> > @fuqiao123 @markshostak , I believe that there is no need to define the spine / leaf function, IP Fabric ( Spine / leaf ) architecture should able to carry all types of services (Services, Storage and Management ) and this types of traffic shall be disaggregated by VXLAN ( VTEP IDs)
> > So my recommendation to update the diagram with normal fabric as described above
> > for 2nd point, for Typical implementation, for sure we need to have dedicated SDN-C, and this should br reflected on both RA , RM
> 
> See the above comments for Mark. I agree we don't need to require spine-leaf. This figure is just an example. I would change the figure accordingly to avoid confusion

thanks

PR #618, opened 2019-11-15
[RM Ch07] Add ETSI NFV to the list of common standards

glimonta:
Fixes #431

PR #619, opened 2019-11-15
[RM Ch07] Modifications to Section 7.10.4

glimonta:
Improve the description of the behaviour of the TPM when taking boot time
integrity measurements.


rabi-abdel:
@glimonta thanks a lot for creating those PR. since this PR and PR: #618 are impacting the same file, is it possible to use the other PR to reflect the Changs you are trying to make here? this is to void conflicts of contents (when we have two PRs changing the same file).

glimonta:
@rabi-abdel Sure, I'll move the changes to this PR (both this and [620](https://github.com/cntt-n/CNTT/pull/620)).

PR #620, opened 2019-11-15
[RM Ch07] Add new section 7.10.5 on Runtime Integrity Measurements

glimonta:
Renumbered the existing 7.10.5 NFVI & VIM section to 7.10.6 and added a
new section 7.10.5 that addresses runtime integrity measurements that can be taken in a platform using the TPM.

The renumbering is because the runtime integrity section should
naturally come after the boot time integrity section.


rabi-abdel:
@glimonta can you please merge those changes into #618? so we have only one PR for all the changes you are proposing to Chapter 7.

glimonta:
The changes were moved to PR #618

PR #648, opened 2019-11-18
[RA1 Ch05] Precise Glance versions

### LINKED ISSUE #647: [RA1 Ch05] Improve Image API section
v2.5 should be precised
https://docs.openstack.org/api-ref/image/versions/index.html#api-versions

collivier:
v2.5 is more precised than v2.0
https://docs.openstack.org/api-ref/image/versions/index.html#api-versions

Fixes #647

### DIFF
--- a/doc/ref_arch/openstack/chapters/chapter05.md
+++ b/doc/ref_arch/openstack/chapters/chapter05.md
@@ -39,10 +39,11 @@ Security compliance and PCI-DSS: https://docs.openstack.org/keystone/train/admin
 
 ### 5.2.2 Glance
 
-| **OpenStack Service** | **Link for API and CLI** | **API/Client Minimum (Baseline) Version** |
-|------------------|----------------------------------------------------|-------------------|
-| Imaging: Glance | https://docs.openstack.org/api-ref/image/v2/index.html#images | Version 2.0 |
-| Imaging: Glance | https://docs.openstack.org/python-glanceclient/latest/ | Version 2.0 |
+| **OpenStack Service** | **Minimal API Version** |
+|-----------------------|-------------------------|
+| Image: Glance         | 2.5                     |
+
+Image Service Versions: https://docs.openstack.org/api-ref/image/versions/index.html#version-history
 
 ### 5.2.3. Cinder
 


collivier:
I don't know if we do list the container and diask formats (as for networking overlays).
If I'm not wrong VIO (VMware in OpenStack) doesn't support qcow2 but it's quite easy to switch from one format to another via qemu utils (see VIO testing in Functest)

Else I could list the default testing values
container_formats = ami,ari,aki,bare,ovf,ova
disk_formats = ami,ari,aki,vhd,vmdk,raw,qcow2,vdi,iso,vhdx

deebhatia:
> I don't know if we do list the container and disk formats (as for networking overlays).
> If I'm not wrong VIO (VMware in OpenStack) doesn't support qcow2 but it's quite easy to switch from one format to another via qemu utils (see VIO testing in Functest)
> 
> Else I could list the default testing values
> container_formats = ami,ari,aki,bare,ovf,ova
> disk_formats = ami,ari,aki,vhd,vmdk,raw,qcow2,vdi,iso,vhdx

@collivier Seeing the test cases in tempest glance v2, test cases uses the format dependent on the order in which they are provided. See [here](https://github.com/openstack/tempest/blob/5642a1ad4a533398264a9dec60910d85e9493cd1/tempest/api/image/v2/test_images.py#L46L47) and [here](https://github.com/openstack/tempest/blob/5642a1ad4a533398264a9dec60910d85e9493cd1/tempest/api/image/v2/test_images.py#L177L187). We should keep it configurable from end user perspective to provide the valid set of values if functest allows with default values as you have mentioned above.

collivier:
deebhatia: agree (then I don't modify the pull request). Functest doesn't override the containers and the disk formats and then keeps the default values here. We provide a tempest conf yaml file which could be modified on purpose. In context of Compliance, we shouldn't have so much customized values. I will list them in the compliance chapter (volume_device_name, ..).

Please find below a few configs from OPNFV installers:
https://git.opnfv.org/releng/tree/jjb/functest/functest-alpine.sh#n133

deebhatia:
@collivier Okay to keep it as it is. For VIO, it [supports](https://docs.vmware.com/en/VMware-Integrated-OpenStack/4.1/com.vmware.openstack.admin.doc/GUID-D50D0804-C0EB-49BF-BC92-9577502BA83B.html) on the fly conversion of image to supported format. So, we should be good.

collivier:
@deebhatia If I'm not wrong, few tests raises timeouts when converting the image on-the-fly. Functest provides 2 helpers to use the right config and to convert the image as first step:
- https://git.opnfv.org/functest/tree/functest/ci/convert_images.sh
- https://git.opnfv.org/functest/tree/functest/ci/config_patch.yaml#n201

deebhatia:
If we are converting it as a prerequisite, then we should also give the correct disk format and container format to import. Right now if I run, will tempest run it with `vmdk` format or `ami` ?

rabi-abdel:
will wait for some time for this to cool down b4 merge. I see still more questions from @deebhatia

collivier:
@rabi-abdel we are diverging to Functest supporting VIO testing. Nothing related to this pull request.

@deebhatia This part has been covered by third-parties (I could only manage the dev part but I tested anyway the vmdk part via KVM which supports this format too). Our Functest scenarii/scenarios publish the vmdk images in glance as they are and the tempest driver writes the correct ids in tempest.conf.

PR #653, opened 2019-11-19
[RM Ch05] Remove Confusion about SR-IOV.

### LINKED ISSUE #652: [RM Ch05] Remove SR-IOV notes
As spotted by @ASawwaf we need to remove reference to SR-IOV in the RM to remove confusion. 
e.g. 

nfvi.net.acc.cfg.002 | Support of HW offload | Yes/No | e.g. support of SR-IOV, SmartNic.
-- | -- | -- | --

and 

SR-IOV (Single-Root Input/Output Virtualisation): Configure PCIe ports to support SR-IOV.

rabi-abdel:
Fixes #652

### DIFF
--- a/doc/ref_model/chapters/chapter05.md
+++ b/doc/ref_model/chapters/chapter05.md
@@ -169,12 +169,14 @@ This section will detail NFVI SW profiles and associated configurations for the 
 
 | .conf | Feature | Type  | Basic | Network Intensive | Compute Intensive |
 |------------------|----------------|----------------|----------------|----------------|----------------|
-| nfvi.net.cfg.001 | vNIC interface | IO virtualisation | virtio1.1 |  virtio1.1 |  virtio1.1 |
+| nfvi.net.cfg.001 | vNIC interface | IO virtualisation | virtio1.1 |  virtio1.1* |  virtio1.1* |
 | nfvi.net.cfg.002 | Overlay protocol | Protocols  | VXLAN, MPLSoUDP, GENEVE, other |  VXLAN, MPLSoUDP, GENEVE, other |VXLAN, MPLSoUDP, GENEVE, other |
 | nfvi.net.cfg.003 | NAT | Yes/No  | Y | Y | Y |
 | nfvi.net.cfg.004 | Security Group | Yes/No  | Y | Y | Y |
 | nfvi.net.cfg.005 | SFC support | Yes/No  | N | Y | Y |
 | nfvi.net.cfg.006 | Traffic patterns symmetry | Yes/No  | Y | Y | Y |
+
+*[VNF Transtion Guidelines.](../chapters/appendix-a.md) may allow for intermediate alternative interfaces.
 
 <p align="center"><b>Table 5-11:</b> Virtual Networking features and configuration for the 3 types of SW profiles.</p>
 
@@ -223,7 +225,7 @@ The host profile and capabilities include:
 1. **Local Disk Capacity**: is the # of local disks and teh capacity of the disks installed on the physical server.
 1. **SMT/HT (SMT: Simultaneous Multithreading/ HT: Hyper Threading)**: Enabled on all physical servers. Gets multiple threads per physical core. Always ON. Configured in the host.
 1. **NUMA (Non-Uniform Memory Access)**: Indicates that vCPU will be on a Socket that is aligned with the associated NIC card and memory. Important for performance optimized VNFs. Configured in the host.
-1. **SR-IOV (Single-Root Input/Output Virtualisation)**: Configure PCIe ports to support SR-IOV (without exposing SR-IOV VFs to workload via PCI-Pass through).
+1. **SR-IOV (Single-Root Input/Output Virtualisation)**: Configure PCIe ports to enable SR-IOV (without exposing SR-IOV VFs to workload via PCI-Pass through).
 1. **smartNIC (aka Intelligent Server Adaptors)**: Accelerated virtual switch using smartNIC
 1. **Cryptography Accelerators**: such as AES-NI, SIMD/AVX, QAT.
 1. **Security features**: such as TRusted Platform Module (TPM).
--- a/doc/ref_model/chapters/chapter05.md
+++ b/doc/ref_model/chapters/chapter05.md
@@ -101,7 +101,7 @@ The following sections detail the NFVI SW profile features per type of virtual r
 | .conf | Feature | Type  | Description |
 |------------------|----------------|----------------|------------------------------------------------------------------------------------------------|
 | nfvi.net.acc.cfg.001 | vSwitch optimisation | Yes/No and SW Optimisation | e.g. DPDK. |
-| nfvi.net.acc.cfg.002 | Support of HW offload | Yes/No | e.g. support of SR-IOV, SmartNic. |
+| nfvi.net.acc.cfg.002 | Support of HW offload | Yes/No | e.g. support of SmartNic. |
 | nfvi.net.acc.cfg.003 | Crypto acceleration | Yes/No |  |
 | nfvi.net.acc.cfg.004 | Crypto Acceleration Interface |Yes/No | |
 
@@ -181,7 +181,7 @@ This section will detail NFVI SW profiles and associated configurations for the 
 | .conf | Feature | Type  | Basic | Network Intensive | Compute Intensive |
 |------------------|----------------|----------------|----------------|----------------|----------------|
 | nfvi.net.acc.cfg.001 | vSwitch optimisation | YeS/No and SW Optimisation | N | Y, DPDK | Y, DPDK |
-| nfvi.net.acc.cfg.002 | Support of HW offload | YeS/No | N | Y, support of SR-IOV and  SmartNic |Y, support of SR-IOV and  SmartNic |
+| nfvi.net.acc.cfg.002 | Support of HW offload | YeS/No | N | Y, support of SmartNic |Y, support of SmartNic |
 | nfvi.net.acc.cfg.003 | Crypto acceleration | Yes/No | N  | Y | Y |
 | nfvi.net.acc.cfg.004 | Crypto Acceleration Interface | Yes/No | N  | Y | Y |
 
@@ -223,7 +223,7 @@ The host profile and capabilities include:
 1. **Local Disk Capacity**: is the # of local disks and teh capacity of the disks installed on the physical server.
 1. **SMT/HT (SMT: Simultaneous Multithreading/ HT: Hyper Threading)**: Enabled on all physical servers. Gets multiple threads per physical core. Always ON. Configured in the host.
 1. **NUMA (Non-Uniform Memory Access)**: Indicates that vCPU will be on a Socket that is aligned with the associated NIC card and memory. Important for performance optimized VNFs. Configured in the host.
-1. **SR-IOV (Single-Root Input/Output Virtualisation)**: Configure PCIe ports to support SR-IOV.
+1. **SR-IOV (Single-Root Input/Output Virtualisation)**: Configure PCIe ports to support SR-IOV (without exposing SR-IOV VFs to workload via PCI-Pass through).
 1. **smartNIC (aka Intelligent Server Adaptors)**: Accelerated virtual switch using smartNIC
 1. **Cryptography Accelerators**: such as AES-NI, SIMD/AVX, QAT.
 1. **Security features**: such as TRusted Platform Module (TPM).
--- a/doc/ref_model/chapters/chapter05.md
+++ b/doc/ref_model/chapters/chapter05.md
@@ -40,6 +40,7 @@ Depending on the requirements of VNFs, a VNFC will be deployed with a NFVI insta
 
 The following sections detail the NFVI SW profile features per type of virtual resource. The list of these features will evolve over time.
 
+<a name="5.1.1"></a>
 ### 5.1.1 Virtual Compute
 
 **Table 5-1** and **Table 5-2** depict the features related to virtual compute.
@@ -60,7 +61,7 @@ The following sections detail the NFVI SW profile features per type of virtual r
 
 <p align="center"><b>Table 5-2:</b> Virtual Compute Acceleration features.</p>
 
-<a name="5.2"></a>
+<a name="5.1.2"></a>
 ### 5.1.2 Virtual Storage
 
 **Table 5-3** and **Table 5-4** depict the features related to virtual storage.
@@ -82,6 +83,7 @@ The following sections detail the NFVI SW profile features per type of virtual r
 
 <p align="center"><b>Table 5-4:</b> Virtual Storage Acceleration features.</p>
 
+<a name="5.1.3"></a>
 ### 5.1.3 Virtual Networking
 
 **Table 5-5** and **Table 5-6** depict the features related to virtual networking.
@@ -169,14 +171,14 @@ This section will detail NFVI SW profiles and associated configurations for the 
 
 | .conf | Feature | Type  | Basic | Network Intensive | Compute Intensive |
 |------------------|----------------|----------------|----------------|----------------|----------------|
-| nfvi.net.cfg.001 | vNIC interface | IO virtualisation | virtio1.1 |  virtio1.1* |  virtio1.1* |
+| nfvi.net.cfg.001 | vNIC interface | IO virtualisation | virtio1.1 |  virtio1.1 or NIC-specific* |  virtio1.1 |
 | nfvi.net.cfg.002 | Overlay protocol | Protocols  | VXLAN, MPLSoUDP, GENEVE, other |  VXLAN, MPLSoUDP, GENEVE, other |VXLAN, MPLSoUDP, GENEVE, other |
 | nfvi.net.cfg.003 | NAT | Yes/No  | Y | Y | Y |
 | nfvi.net.cfg.004 | Security Group | Yes/No  | Y | Y | Y |
 | nfvi.net.cfg.005 | SFC support | Yes/No  | N | Y | Y |
 | nfvi.net.cfg.006 | Traffic patterns symmetry | Yes/No  | Y | Y | Y |
 
-*[VNF Transtion Guidelines.](../chapters/appendix-a.md) may allow for intermediate alternative interfaces.
+*For enabling SR-IOV VFs to be directly passed to VNFC, NIC vendor specific interface is transiently allowed but not recommended, and it should be prohibited when mature enough solutions will be available for providing the same efficiency level (especially regarding CPU and energy consumption).
 
 <p align="center"><b>Table 5-11:</b> Virtual Networking features and configuration for the 3 types of SW profiles.</p>
 
@@ -225,7 +227,7 @@ The host profile and capabilities include:
 1. **Local Disk Capacity**: is the # of local disks and teh capacity of the disks installed on the physical server.
 1. **SMT/HT (SMT: Simultaneous Multithreading/ HT: Hyper Threading)**: Enabled on all physical servers. Gets multiple threads per physical core. Always ON. Configured in the host.
 1. **NUMA (Non-Uniform Memory Access)**: Indicates that vCPU will be on a Socket that is aligned with the associated NIC card and memory. Important for performance optimized VNFs. Configured in the host.
-1. **SR-IOV (Single-Root Input/Output Virtualisation)**: Configure PCIe ports to enable SR-IOV (without exposing SR-IOV VFs to workload via PCI-Pass through).
+1. **SR-IOV (Single-Root Input/Output Virtualisation)**: Configure PCIe ports to enable SR-IOV.
 1. **smartNIC (aka Intelligent Server Adaptors)**: Accelerated virtual switch using smartNIC
 1. **Cryptography Accelerators**: such as AES-NI, SIMD/AVX, QAT.
 1. **Security features**: such as TRusted Platform Module (TPM).
--- a/doc/ref_model/chapters/chapter05.md
+++ b/doc/ref_model/chapters/chapter05.md
@@ -178,7 +178,7 @@ This section will detail NFVI SW profiles and associated configurations for the 
 | nfvi.net.cfg.005 | SFC support | Yes/No  | N | Y | Y |
 | nfvi.net.cfg.006 | Traffic patterns symmetry | Yes/No  | Y | Y | Y |
 
-*[VNF Transtion Guidelines.](../chapters/appendix-a.md) might have other interfaces (such as SR-IOV VFs to be directly passed to VNFC) or NIC-specefic drivers on guest machines transiently allowed until mature enough solutions are available with a similar efficiency level (for example regarding CPU and energy consumption).
+*[VNF Transtion Guidelines.](../chapters/appendix-a.md) might have other interfaces (such as SR-IOV VFs to be directly passed to VNFC) or NIC-specific drivers on guest machines transiently allowed until mature enough solutions are available with a similar efficiency level (for example regarding CPU and energy consumption).
 
 <p align="center"><b>Table 5-11:</b> Virtual Networking features and configuration for the 3 types of SW profiles.</p>
 
--- a/doc/ref_model/chapters/chapter05.md
+++ b/doc/ref_model/chapters/chapter05.md
@@ -171,14 +171,14 @@ This section will detail NFVI SW profiles and associated configurations for the 
 
 | .conf | Feature | Type  | Basic | Network Intensive | Compute Intensive |
 |------------------|----------------|----------------|----------------|----------------|----------------|
-| nfvi.net.cfg.001 | vNIC interface | IO virtualisation | virtio1.1 |  virtio1.1 or NIC-specific* |  virtio1.1 |
+| nfvi.net.cfg.001 | vNIC interface | IO virtualisation | virtio1.1 |  virtio1.1* |  virtio1.1 |
 | nfvi.net.cfg.002 | Overlay protocol | Protocols  | VXLAN, MPLSoUDP, GENEVE, other |  VXLAN, MPLSoUDP, GENEVE, other |VXLAN, MPLSoUDP, GENEVE, other |
 | nfvi.net.cfg.003 | NAT | Yes/No  | Y | Y | Y |
 | nfvi.net.cfg.004 | Security Group | Yes/No  | Y | Y | Y |
 | nfvi.net.cfg.005 | SFC support | Yes/No  | N | Y | Y |
 | nfvi.net.cfg.006 | Traffic patterns symmetry | Yes/No  | Y | Y | Y |
 
-*For enabling SR-IOV VFs to be directly passed to VNFC, NIC vendor specific interface is transiently allowed but not recommended, and it should be prohibited when mature enough solutions will be available for providing the same efficiency level (especially regarding CPU and energy consumption).
+*[VNF Transtion Guidelines.](../chapters/appendix-a.md) might have other interfaces (such as SR-IOV VFs to be directly passed to VNFC) or NIC-specefic drivers on guest machines transiently allowed until mature enough solutions are available with a similar efficiency level (for example regarding CPU and energy consumption).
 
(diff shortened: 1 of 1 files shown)


rabi-abdel:
@ASawwaf here it is as discussed.

rabi-abdel:
@karinesevilla I clarified your point about SR-IOV used within NFVI without bypassing NFVI. I have removed vendor specific interfaces from the list. please review

rabi-abdel:
@pgoyal01 kept SR-IOV as we are not opposing to the technology as long as it is used within NFVI without exposing VFs to workload and having to require custom drivers. please review

doc/ref_model/chapters/chapter05.md
```diff
@@ -223,7 +223,7 @@ The host profile and capabilities include:
 1. **Local Disk Capacity**: is the # of local disks and teh capacity of the disks installed on the physical server.
 1. **SMT/HT (SMT: Simultaneous Multithreading/ HT: Hyper Threading)**: Enabled on all physical servers. Gets multiple threads per physical core. Always ON. Configured in the host.
 1. **NUMA (Non-Uniform Memory Access)**: Indicates that vCPU will be on a Socket that is aligned with the associated NIC card and memory. Important for performance optimized VNFs. Configured in the host.
-1. **SR-IOV (Single-Root Input/Output Virtualisation)**: Configure PCIe ports to support SR-IOV.
+1. **SR-IOV (Single-Root Input/Output Virtualisation)**: Configure PCIe ports to support SR-IOV (without exposing SR-IOV VFs to workload via PCI-Pass through).
```
pgoyal01: Maybe change in wording: Configure multiple Virtual Functions (VF) on the PCIe ports on SR-IOV enabled cards.

doc/ref_model/chapters/chapter05.md
```diff
@@ -181,7 +181,7 @@ This section will detail NFVI SW profiles and associated configurations for the
 | .conf | Feature | Type  | Basic | Network Intensive | Compute Intensive |
 |------------------|----------------|----------------|----------------|----------------|----------------|
 | nfvi.net.acc.cfg.001 | vSwitch optimisation | YeS/No and SW Optimisation | N | Y, DPDK | Y, DPDK |
-| nfvi.net.acc.cfg.002 | Support of HW offload | YeS/No | N | Y, support of SR-IOV and  SmartNic |Y, support of SR-IOV and  SmartNic |
+| nfvi.net.acc.cfg.002 | Support of HW offload | YeS/No | N | Y, support of SmartNic |Y, support of SmartNic |
```
pgoyal01: Since HW offload is required for Network Intensive and Compute intensive, what are the options in the absence of SmartNICs?

doc/ref_model/chapters/chapter05.md
deebhatia: In addition to this, we should define `SR-IOV` in glossary. We can mention there that PCI-passthrough is not part of it.

rabiabdel: @pgoyal01 the NFVI doesn't configure VF ports, it enables SRIOV on SRIOV devices.

rabiabdel: @deebhatia if it has to be used on different place in RM, we can move it to glossary. I think it should be OK.

doc/ref_model/chapters/chapter05.md
rabiabdel: SRIOV is not a HW offload technology it is a mechanism to expose a hardware device into multiple ones. A other example of hardware offload is a HW acceleration card.

ASawwaf: @pgoyal01 , first why HW offload is required for Network Intensive and Compute intensive?
as per my previous response and to make more clear 

for Compute Intensive (ch02: for VNFCs that require predictable computing performance and low network latency ) : from practical speaking no need to have HW offload as we can achieve this profile by OVS or OVS+DPDK with minimal vCPU allocated to PMDs

for Network Intensive: (high network throughput and low network latency.) which required one of the below Tech :
1-SRIOV
2- OVS+DPDK (a large number of Core be allocated to PMDs )
3- HW offloading cards  (Marketing name SmartNIC ) or FPGA to offload the allocated core on the NIC card 

So the above 3 option is available and Yes I admit that SRIOV is proved technology for high performance but it has many drawbacks and I  considered it as fake virtualization and against what we looking for,

So my recommendation to remove it totally.
I hope it is clear 

@markshostak @rabi-abdel @karinesevilla @petorre

xavier-grall:
@rabi-abdel @ASawwaf
As I understand, SRIOV should be excluded from CNTT reference mainly because it requires NIC-specific drivers (and not only the virtio driver) to be implemented by VNFC, thus loosing the hardware abstraction, and that seems a quite relevant reason.
Consequently, for providing VNFC with high throughput and/or low latency, only 2 solutions are currently available: DPDK and virtio-compliant smartNIC.
So does it mean that **only DPDK vSwitchs** should be considered in a first step (despite they are very CPU and energy consumer and require complex tuning) ?
Or should **virtio-compliant smartNIC** also be considered for vSwitch offloading (despite such NICs are quite rare) ? for a future release ?
And what about **non-virtio-compliant smartNIC with DPDK-based adaptation** (despite it is a quite complex engineering) ? to be avoided ?

I fully agree with @walterkozlowski concerning "the current, immature state of technology" and that SRIOV and DPDK "usage is not desirable, and the VNF vendors should work with operators on evolving alternatives", and I would even precise that DPDK-based VNFC is also not desirable (and not only DPDK-based NFVI network connectivity/switching) for resource consumption reasons (CPU and energy): can it be the purpose of CNTT in future releases ? the issue is to have standard open interfaces for enabling data plane VNFC to run on adequat programmable hardware.

Note: @rabi-abdel SRIOV is not itself a HW offload technology, but applied to PCI device as NIC in the context of virtualisation (especially NFVI), it can be considered as a real HW offloading solution because it allows to offload some processing applied to workload network traffic: VLAN tagging, L3/L4 segmentation/fragmentation/checksumming, L2 anti-spoofing, and possibly also, depending on NIC, VLAN trunking, VXLAN encapsulation, mirroring...

rabi-abdel:
@karinesevilla @xavier-grall I think we are talking cross purposes :) I am not opposing to SR-IOV, I am opposing to PCI-Pass Through. SmartNICs can use SR-IOV but don't have to use PCIe-Pass Through.

In respect to this PR, there are three changes I made:

1. remove specific vendor drivers: I believe we all are on agreement on this should be removed, it is against CNTT principles to have vendor specific drivers in Guest machines. 
2. Clarified SR-IOV definition: I have not removed it, I clarified that this relates to enabling SR-IOV and not pass VF to workload via PCIe Pass Through. are we all OK with this?
3. I removed SR-IOV as being an example of HW Offload: I still think keeping it will cause confusion as people will quickly assume we mean PCIe-Pass Though which we all agree it is bad idea. The implementation of SmartNIC can always use SR-IOV which is something we are not saying no to. if we add "Acceleration Card" as another example, will this satisfy your needs?

Please let me know

walterkozlowski:
I concur, Rabi. And, as we all seem to agree, new technologies and improved performance chipsets keep coming. I suggest add a remark there that we are conscious of this fact and the technologies explicitly mentioned are examples of currently available solutions, and that CNTT are going to keep updating the model accordingly.

Walter Kozlowski
Architecture & Strategy Principal
Converged Cloud & Infrastructure Principal
Networks and IT
Telstra
+61 407 724 612
________________________________

xavier-grall:
@rabi-abdel 

>    1. remove specific vendor drivers: I believe we all are on agreement on this should be removed, it is against CNTT principles to have vendor specific drivers in Guest machines.

**OK for a near future target** when smartNIC-based techno will be matured enough (and available on RI labs), **but not for now** because, basic SRIOV VF is the most pragmatic and resource-efficient solution for providing network connectivity to data plane VNFC, thus requiring NIC vendor specific drivers in Guest machines.
Otherwise, it would implies that only DPDK can currently be used for providing network connectivity to workloads requiring high throughput and/or low latency, thus also implying the CNTT NI profile to consume many CPU and energy resources (in addition to a complex tuning).

>    2. Clarified SR-IOV definition: I have not removed it, I clarified that this relates to enabling SR-IOV and not pass VF to workload via PCIe Pass Through. are we all OK with this?

**No, neither for now, nor for future**, because it seems acceptable that a VF is passed to workload via PCI Pass Through **provided that the PCI device supports a generic driver as virtio-net for VF** (which I named virtio-compliant smarNIC in my previous comment, as Mellanox Virtio or Intel vDPA).
What would be the issue


I think we need much more opinions on that fundamental point

petorre:
More generally than trying to list SR-IOV with current implementations, vSwitches/vRouters and smart/acceleration adapters, looking at separation between platform and application it comes to two categories of Connectivity options:
1. Respecting layer boundary: Exposing generic interface (like virtio) from vSwitches/vRouters running on CPU and/or adapters with acceleration function. This avoids HW definition in applications.
2. Not respecting layer boundary: Requiring vendor-specific drivers for HW adapters directly exposed to VNF/CNF. This increases validation/integration cost and makes onboarding/LCM harder. Main example is current SR-IOV NICs using vendor-specific drivers. It can also limit/change security choices and lose functionality provided by vSwitches/vRouters.

walterkozlowski:
Thanks, Petar. Spot on.
Regards
Walter

Walter Kozlowski
Architecture & Strategy Principal
Converged Cloud & Infrastructure Principal
Networks and IT
Telstra
+61 407 724 612
________________________________

rabi-abdel:
> More generally than trying to list SR-IOV with current implementations, vSwitches/vRouters and smart/acceleration adapters, looking at separation between platform and application it comes to two categories of Connectivity options:
> 
> 1. Respecting layer boundary: Exposing generic interface (like virtio) from vSwitches/vRouters running on CPU and/or adapters with acceleration function. This avoids HW definition in applications.
> 2. Not respecting layer boundary: Requiring vendor-specific drivers for HW adapters directly exposed to VNF/CNF. This increases validation/integration cost and makes onboarding/LCM harder. Main example is current SR-IOV NICs using vendor-specific drivers. It can also limit/change security choices and lose functionality provided by vSwitches/vRouters.

echoing Walter, spot on.

rabi-abdel:
@xavier-grall @karinesevilla 

To address your valid concerns:

In terms of point 1. 

We are working on "transitional plan" within CNTT (with the help from @petorre ) to @xavier-grall point to east the transition from where we are now to cloud native. to your point @xavier-grall if the transition plan allows for vendor specific driver today, it must be clear in the plan when this will no longer be allowed. in my views CNTT must never allow it because this is completely against the fundamental principles of it. (the transitional plan will clarify the adoption strategy).

In terms of point 2.

There are many technologies such as DPDK relays or virtue-forwarder which will enumerate SR-IOV directly from host and still expose virtue-net to VMs. we don't have to by-pass the kernel. now there is the one use case which the NIC itself implement virtio driver (which will still limit the migration and I don't think it is a preferred solution) but we can re-word the definition of SR-IOV to make it focused on the enablement of SR-IOV capabilities in the hardware without being very perceptive on the implementation details. does this sound ok for you?

xavier-grall:
@rabi-abdel 

In terms of point 1.
We, Karine and me, understand that you propose that CNTT allows SR-IOV VF PCI pass through with vendor specific drivers for a temporary period (transition plan), but that it must not be the target model which requires the full HW abstraction (ie "respecting layer boundary” as per @petorre) and thus, no vendor specific driver to be implemented by VNFC.
So we agree that proposal with the following principle to be written (possibly re-worded): **allowing SR-IOV VF PCI pass through for VNF network connectivity as long as a solution with minimal CPU resource and without vendor specific driver is not available** (ie defined/validated/verified by CNTT), ideally based on a virtio-compliant smartNIC enabling full vSwitch datapath offload.
For clarification (if needed), we also think that DPDK vSwitch should be possible (eg as a RA option) because it paves the way of vSwitch offloading by smartNIC with much more features for NFVI networking which must be functionally tested in RI/RC.

In terms of point 2.
OK with that point, but it seems included in our position for point 1.

Maybe some VNF vendors also have an opinion to share: @kedmison ? @ulikleber ?

rabi-abdel:
> @rabi-abdel
> 
> In terms of point 1.
> We, Karine and me, understand that you propose that CNTT allows SR-IOV VF PCI pass through with vendor specific drivers for a temporary period (transition plan), but that it must not be the target model which requires the full HW abstraction (ie "respecting layer boundary” as per @petorre) and thus, no vendor specific driver to be implemented by VNFC.
> So we agree that proposal with the following principle to be written (possibly re-worded): **allowing SR-IOV VF PCI pass through for VNF network connectivity as long as a solution with minimal CPU resource and without vendor specific driver is not available** (ie defined/validated/verified by CNTT), ideally based on a virtio-compliant smartNIC enabling full vSwitch datapath offload.
> For clarification (if needed), we also think that DPDK vSwitch should be possible (eg as a RA option) because it paves the way of vSwitch offloading by smartNIC with much more features for NFVI networking which must be functionally tested in RI/RC.
> 
> In terms of point 2.
> OK with that point, but it seems included in our position for point 1.
> 
> Maybe some VNF vendors also have an opinion to share: @kedmison ? @ulikleber ?

@xavier-grall last commit hopefully addresses your concern. I have also created an issue to make sure we include the statement you suggested in the transitional guidelines in Appendix-A (https://github.com/cntt-n/CNTT/issues/678)

doc/ref_model/chapters/chapter05.md
```diff
@@ -169,19 +169,21 @@ This section will detail NFVI SW profiles and associated configurations for the
...
 |------------------|----------------|----------------|----------------|----------------|----------------|
-| nfvi.net.cfg.001 | vNIC interface | IO virtualisation | virtio1.1 |  virtio1.1, i40evf (Intel driver for VF SR-IOV) |  virtio1.1, i40evf (Intel driver for VF SR-IOV) |
+| nfvi.net.cfg.001 | vNIC interface | IO virtualisation | virtio1.1 |  virtio1.1* |  virtio1.1* |
 | nfvi.net.cfg.002 | Overlay protocol | Protocols  | VXLAN, MPLSoUDP, GENEVE, other |  VXLAN, MPLSoUDP, GENEVE, other |VXLAN, MPLSoUDP, GENEVE, other |
 | nfvi.net.cfg.003 | NAT | Yes/No  | Y | Y | Y |
 | nfvi.net.cfg.004 | Security Group | Yes/No  | Y | Y | Y |
 | nfvi.net.cfg.005 | SFC support | Yes/No  | N | Y | Y |
 | nfvi.net.cfg.006 | Traffic patterns symmetry | Yes/No  | Y | Y | Y |
 
+*[VNF Transtion Guidelines.](../chapters/appendix-a.md) may allow for intermediate alternative interfaces.
```
deebhatia: Small typo, `Transtion` -> `Transition`. Remove the period at the end of `Guidelines`.

walterkozlowski:
We have been going through this for a long time by now.  We need to close it out.  I still have problems with the list "The host profile and capabilities include:".  Items 1 - 4 (number of CPU sockets, number of cores/CPU, RAM, local disk capacity) seem to be obvious and mandatory "capabilities", while the rest 5 - 10 seem to be optional, hence for example HT can be on or off, and our SR-IOV may be there or not.  I suggest, marking capabilities 1- 4 as mandatory, the rest as optional, i.e. allowed within the limitations specified (like SR-IOV without exposing SR-IOV VFs to workload via PCI-Pass through) but making clear they are not mandated.

markshostak:
@rabi-abdel @xavier-grall @karinesevilla 

\1) **Non-Technical:** Given traditional SRIOV (i.e. SRIOV over PT) is a key, cost-effective (at the expense of having to load NIC-specific drivers) technology, as Karine already pointed out, **I think it's very important we _explicitly_ state we have a transition plan**, hopefully that has a roadmap to similar performance, at a comparable price point, to move from traditional SRIOV to a virtio solution. 

We want to make CNTT easy to adopt, and I'm concerned simply eradicating all references to "SR-IOV" or banishing it to a transition guide will scare potential adopters away if they use SRIOV, and probably panic existing member that are using it. If we want to move to something else, let's call it out by name and say SRIOVoPT is in vogue today, but our target architecture is virtio and SRIOV is optionally available today in CNTT, but will be deprecated when (fill in when, _here_). Some hate SRIOV, but for others it's very important, and in my mind, simply erasing it makes us look like we're summarily dismissing what others currently consider critical to their business. 

\2) **Technical:** It's been stated we don't have an issue with SRIOV, per se. Does that mean our issue is with PT, or, more likely the NIC-specific drivers? My concern here is that SmartNICs frequently leverage PT, and even SRIOV, so does that mean the "SRIOV" prohibition will extend to SmartNICs as well? What about other acceleration/offload technologies that use an ABI over PT (e.g. vGPU)?

In identifying the root of the SRIOV concern, I currently see 3 possibilities. Let me know if i missed one:
1) "SR-IOV"
2) PCIe PassThrough (PT)
3) Any technology that requires loading hardware-specific drivers in a guest VM

My money is currently on \#3. Comments?

Technology like XVIO seem to offer a compromise, which appears to use PT and SRIOV as transport, but is integrated w/ virtio so it allows features such as VM Live Migration and an OVS-level of granularity in control of packet flow, yet offers performance near that of SRIOVoPT.

@rabi-abdel / all, Would this type of solution be acceptable, despite leveraging some verboten technology? IMHO we need to identify and document what the root concern/criteria is, otherwise CNTT prohibitions will be applied subjectively and reflect poorly.

Note: I haven't seen enough information to tell if XVIO is a proprietary solution and exactly how it works, so just think of it as a model for this discussion.

xavier-grall:
@markshostak 

The concern is to avoid any solution that breaks the hardware abstraction principle (by requiring the VNF to implement any vendor specific software: driver…).
This is quite relevant, but the current available solutions may provide that abstraction with a high CPU and energy resource consumption, so that breaking the abstraction is the pragmatic choice for efficiency.
 
XVIO is the Netronome solution relying on a DPDK-based software (namely virtio-forwarder) for providing an abstraction of Netronome NIC hardware, especially when used for vSwitch offload (and it consumes nearly the same CPU resource as a simple L2 forwarding by OVS-DPDK).
So, it does not require vendor specific driver to be implemented by VNF, but only to be requested as a specific vNIC type because it implies a specific backend.
It is related to non-virtio-compliant smartNIC with DPDK-based adaptation which I mention in a previous comment.

markshostak:
> 
> 
> The concern is to avoid any solution that breaks the hardware abstraction principle (by requiring the VNF to implement any vendor specific software: driver…).

So, \#3, above.

>
> This is quite relevant, but the current available solutions may provide that abstraction with a high CPU and energy resource consumption, so that breaking the abstraction is the pragmatic choice for efficiency.

If breaking the abstraction is the pragmatic choice, we should be careful about prohibiting it. Perhaps that's the criteria for the when to deprecate SRIOVoPT.

> 
> XVIO is the Netronome solution relying on a DPDK-based software (namely virtio-forwarder) for providing an abstraction of Netronome NIC hardware...

I'm assuming that implies it requires vendor-specific s/w drivers, but we're ok with that, provided they're relegated to the host. Correct?

Does it use PT at all?

Thanks,
-Mark

rabiabdel:
It might be best to discuss this during TSC. I am sure we can get to a text that satisfies everybody and move on! @xavier-grall will you be able to make it to the call Wednesday to get to a conclusion on this?

xavier-grall:
@rabi-abdel  - I have just commited a text proposal in https://github.com/cntt-n/CNTT/tree/rabi-abdel-patch-3 :
- we agree to remove SRIOV as a HW offload function and the Intel specific driver example (i40e)
- we even think that SRIOV VF passthrough must not be used for CI profile and thus be restricted to NI profile
- But, we would prefer not to banish VF passthrough in an annex (which is quite empty for now...)


@markshostak 
> I'm assuming that implies it requires vendor-specific s/w drivers, but we're ok with that, provided they're relegated to the host. Correct?

Yes
 
> Does it use PT at all?

No, there is no PT in XVIO case

rabi-abdel:
@xavier-grall generally I am ok with your text. I have edit it slightly in a way that we don't have to go back to RM and change the text when we have a mature solution. I hope this text will give you enough assurance that our plan is certainly not to punish VNF vendors. instead we need to work out a transitional period with them. hence many of those discussion will need to occur and be document in the annex and hence my slight modification. let me know/approve if you are ok with the proposed text so we can move on :)

doc/ref_model/chapters/chapter05.md
```diff
@@ -169,19 +171,21 @@ This section will detail NFVI SW profiles and associated configurations for the
...
 |------------------|----------------|----------------|----------------|----------------|----------------|
-| nfvi.net.cfg.001 | vNIC interface | IO virtualisation | virtio1.1 |  virtio1.1, i40evf (Intel driver for VF SR-IOV) |  virtio1.1, i40evf (Intel driver for VF SR-IOV) |
+| nfvi.net.cfg.001 | vNIC interface | IO virtualisation | virtio1.1 |  virtio1.1* |  virtio1.1 |
 | nfvi.net.cfg.002 | Overlay protocol | Protocols  | VXLAN, MPLSoUDP, GENEVE, other |  VXLAN, MPLSoUDP, GENEVE, other |VXLAN, MPLSoUDP, GENEVE, other |
 | nfvi.net.cfg.003 | NAT | Yes/No  | Y | Y | Y |
 | nfvi.net.cfg.004 | Security Group | Yes/No  | Y | Y | Y |
 | nfvi.net.cfg.005 | SFC support | Yes/No  | N | Y | Y |
 | nfvi.net.cfg.006 | Traffic patterns symmetry | Yes/No  | Y | Y | Y |
 
+*[VNF Transtion Guidelines.](../chapters/appendix-a.md) might have other interfaces (such as SR-IOV VFs to be directly passed to VNFC) or NIC-specefic drivers on guest machines transiently allowed until mature enough solutions are available with a similar efficiency level (for example regarding CPU and energy consumption).
```
deebhatia: Small typo. `NIC-specefic` -> `NIC-specific`.

xavier-grall: done

xavier-grall:
Thank you Rabi, ok for us
If you add me as reviewer, I will approve

rabiabdel:
> Thank you Rabi, ok for us
> If you add me as reviewer, I will approve

Just did.

kedmison:
@rabi-abdel wrote: 
> [...] we don't have to go back to RM and change the text when we have a mature solution.

I think it would be good to in fact do exactly that; go back to RM and change the text once we have a mature solution.
We are specifying a permitted API (SR-IOV), but we effectively want to mark it as deprecated right now and indicate that it is subject to replacement/removal in future (e.g. a future generation of profile).  For now though, we need to be clear about which profiles are able to use it, and which ones do not.

A generational aspect to the profiles would allow us to define the current profiles to permit SR-IOV, and to indicate that it is deprecated already, pending production-grade replacements that meet our abstraction (and other) requirements.  

Then, an update of of the RM doc to specify that the next generation of profiles do not support SR-IOV, but do support <insert-alternative-technology-here>, can be added to indicate what VNF vendors should expect.    This would be a clear and transparent way of indicating that SR-IOV is permitted, but also not intended for long-term use, while also giving indications that vendors should launch the programs necessary to investigate, test, and improve acceptable SR-IOV replacement technologies.

markshostak:
xavier wrote:
> The concern is to avoid any solution that breaks the hardware abstraction principle (by requiring the VNF to implement any vendor specific software: driver…).

How do we explain retaining other technologies that require vendor-specific software drivers, such as vGPU or even just extended CPU instruction sets that require VNF compilation for vendor-specific target micro-architectures? These also break the h/w abstraction principle, so it seems inconsistent.

rabi-abdel:
> @rabi-abdel wrote:
> 
> > [...] we don't have to go back to RM and change the text when we have a mature solution.
> 
> I think it would be good to in fact do exactly that; go back to RM and change the text once we have a mature solution.
> We are specifying a permitted API (SR-IOV), but we effectively want to mark it as deprecated right now and indicate that it is subject to replacement/removal in future (e.g. a future generation of profile). For now though, we need to be clear about which profiles are able to use it, and which ones do not.
> 
> A generational aspect to the profiles would allow us to define the current profiles to permit SR-IOV, and to indicate that it is deprecated already, pending production-grade replacements that meet our abstraction (and other) requirements.
> 
> Then, an update of of the RM doc to specify that the next generation of profiles do not support SR-IOV, but do support , can be added to indicate what VNF vendors should expect. This would be a clear and transparent way of indicating that SR-IOV is permitted, but also not intended for long-term use, while also giving indications that vendors should launch the programs necessary to investigate, test, and improve acceptable SR-IOV replacement technologies.

HI @kedmison the details of what VNF to expect will be in the annex to set the transitional plan. in there it will indicate what interfaces will be allowed and when. I hope that clarifies it.

rabi-abdel:
Once this approved, we can immediately start looking at the transitional plan. as mentioned, we will work out the plan based on what mature technologies available. @markshostak if you are OK with this plan, please approve to move forward as this requires your approval as WSL.

rabi-abdel:
since more progress made on this, I suggest to remove Technical Steering Tag and try to get consensus in here. unless people feel it requires "technical steering" discussion?

rabi-abdel:
> xavier wrote:
> 
> > The concern is to avoid any solution that breaks the hardware abstraction principle (by requiring the VNF to implement any vendor specific software: driver…).
> 
> How do we explain retaining other technologies that require vendor-specific software drivers, such as vGPU or even just extended CPU instruction sets that require VNF compilation for vendor-specific target micro-architectures? These also break the h/w abstraction principle, so it seems inconsistent.

@markshostak: let us not mix acceleration technology with interfaces, if the GPU or FPGA or any other HW Acceleration techniques is providing network acceleration functionality, the text here applies to those too .  if it is to provide Compute acceleration, RM clearly saying "to be worked on".

rabi-abdel:
> @rabi-abdel @xavier-grall @karinesevilla
> 
> \1) **Non-Technical:** Given traditional SRIOV (i.e. SRIOV over PT) is a key, cost-effective (at the expense of having to load NIC-specific drivers) technology, as Karine already pointed out, **I think it's very important we _explicitly_ state we have a transition plan**, hopefully that has a roadmap to similar performance, at a comparable price point, to move from traditional SRIOV to a virtio solution.
> 
> We want to make CNTT easy to adopt, and I'm concerned simply eradicating all references to "SR-IOV" or banishing it to a transition guide will scare potential adopters away if they use SRIOV, and probably panic existing members that are using it. If we want to move to something else, let's call it out by name and say SRIOVoPT is in vogue today, but our target architecture is virtio and SRIOV is optionally available today in CNTT, but will be deprecated when (fill in when, _here_). Some hate SRIOV, but for others it's very important, and in my mind, simply erasing it makes us look like we're summarily dismissing what others currently consider critical to their business.
> 
> \2) **Technical:** It's been stated we don't have an issue with SRIOV, per se. Does that mean our issue is with PT, or, more likely the NIC-specific drivers? My concern here is that SmartNICs frequently leverage PT, and even SRIOV, so does that mean the "SRIOV" prohibition will extend to SmartNICs as well? What about other acceleration/offload technologies that use an ABI over PT (e.g. vGPU)?
> 
> In identifying the root of the SRIOV concern, I currently see 3 possibilities. Let me know if i missed one:
> 
> 1. "SR-IOV"
> 2. PCIe PassThrough (PT)
> 3. Any technology that requires loading hardware-specific drivers in a guest VM
> 
> My money is currently on #3. Comments?
> 
> Technology like XVIO seem to offer a compromise, which appears to use PT and SRIOV as transport, but is integrated w/ virtio so it allows features such as VM Live Migration and an OVS-level of granularity in control of packet flow, yet offers performance near that of SRIOVoPT.
> 
> @rabi-abdel / all, Would this type of solution be acceptable, despite leveraging some verboten technology? IMHO we need to identify and document what the root concern/criteria is, otherwise CNTT prohibitions will be applied subjectively and reflect poorly.
> 
> Note: I haven't seen enough information to tell if XVIO is a proprietary solution and exactly how it works, so just think of it as a model for this discussion.

@markshostak Valid point, this is why RA 1 should propose those technologies and come up with one (XVIO is netronome specific though but there are very similar open projects solutions to do the same). I am confident we will get to a solution but it might take time to get it matured, upstreamed, supported by vendors, during this time, we will have to look at the transitional plan and what we will have to do to ease the transition.

Issue #403, opened 2019-10-16
[RI1 Ch04] Lab Requirements


rabi-abdel:
This issue to start initial content for Lab Requirements (specify in details the minimum specs of hardware needed for lab - including compute nodes, network, storage - for each of the profiles. this also should specify the topology of the lab expected.

cntt-n:
@fuqiao1@outlook.com

mf4716:
@rrajesh70 @fuqiao123 - we have initial requirements in Chapter 4 already.  This needs to be reviewed with our Infra WG, CNTT RA, and OPNFV teams to ensure we have sufficient requirements to meet the needs of each profile and we need a topology diagram as well.  @rrajesh70 - this is closely tied to RI Labs efforts.  Please assign as needed.

lylavoie:
@rabi-abdel and @mf4716 , I think this issue can also be closed / resolved by pull #654 .

mf4716:
status?

Issue #564, opened 2019-11-06
[RI1 Ch04] Lab Hardware Specification


lylavoie:
We need to finalize the hardware roadmap and specification for RI 1, so labs can acquire and setup hardware.  To facilitate this discussion, a wiki page has been created here: https://wiki.opnfv.org/display/pharos/CNTT+RI+Lab+Infrastructure+Discussion, along with a set of proposals / agreements for the "labs release 1", i.e. what is included in the RI 1 specification materials.

lylavoie:
@rabi-abdel Can you apply the right RI1 labels to this, I don't have write access, so I can't add labels...Thanks

rabiabdel:
> @rabi-abdel Can you apply the right RI1 labels to this, I don't have write access, so I can't add labels...Thanks

invite to you on it's way so you can do it in the future

mf4716:
RI Ch4 is reserved for Lab Requirements - we can tie the discussion/disposition to this chapter.  https://cntt-n.github.io/CNTT/doc/ref_impl/cntt-ri (RI)

mf4716:
@lylavoie - can we get an update?

lylavoie:
@mf4716 This issue should have been closed / resolved by #654, so I'm closing this.

PR #654, opened 2019-11-20
[RI1 Ch04] Update Reference Implementation Lab Hardware

lylavoie:
Per the discussions on the [OPNFV Infra Wiki](https://wiki.opnfv.org/display/INF/CNTT+RI+Lab+Infrastructure+Discussion) and the Infra WG meetings, these changes aim to update the hardware requirements for Release 1 of the Reference Implementation to those needed to support the Airship installer and other basic OpenStack deployments.

### DIFF
--- a/doc/ref_impl/cntt-ri/chapters/chapter04.md
+++ b/doc/ref_impl/cntt-ri/chapters/chapter04.md
@@ -14,7 +14,7 @@
 ## 4.1 Introduction
 This chapter provides the laboratory hardware requirements needed to deploy the reference implementation.  These requirements represent a minimum set of requirements, where different labs may have hardware or capabilities exceeding these minimum specifications.  
 
-The lab provides a set of bare-metal servers, supported by a dedicated network infrastructure.  Users should be provided access to the servers, their operating systems, and out-of-band management (i.e. IPMI, IDRAC, ILO, etc.).  
+The lab provides a set of physical servers, supported by a dedicated network infrastructure.  Users should be provided access to the servers, their operating systems, and out-of-band management (i.e. IPMI, IDRAC, ILO, etc.).  
 
 <a name="4.2"></a>
 ## 4.2 Lab Requirements
@@ -27,10 +27,10 @@ A CNTT compliant *Pod* shall meet the following requirements.
 2. Six (6) physical servers, serving as either compute or controllers
 3. A configured network topology allowing for: Out-of-Band Management, Admin, Public, Private, and Storage Networks
 
-The Jump Host / Test Host system will be utilized to install and manage the operation of the *Pod*.  For example, the system can be used to host the installer software used to deploy and configure OpenStack on the other bare-metal servers.  Once installed, the system may be used to run test automations or it may be used to generate test traffic intended to measure the performance of a VNF operating within the *Pod*.  To support these operations, the bare-metal server provided for the Jump / Test Host shall meet the same hardware requirements as compute / controller, as outline below.  In this context, the system may be considered as a miscellaneous use  or general purpose system.
+The Jump Host / Test Host system will be utilized to install and manage the operation of the *Pod*.  For example, the system can be used to host the installer software used to deploy and configure OpenStack on the other physical servers.  Once installed, the system may be used to run test automations or it may be used to generate test traffic intended to measure the performance of a VNF operating within the *Pod*.  To support these operations, the physical server provided for the Jump / Test Host shall meet the same hardware requirements as compute / controller, as outline below.  In this context, the system may be considered as a miscellaneous use  or general purpose system.
 
 <a name="4.2.1"></a>
-### 4.2.1 Bare-metal Server Requirements
+### 4.2.1 Physical Server Requirements
 Each server shall meet the following minimum specifications:
 
 - **CPU**
@@ -44,15 +44,15 @@ Each server shall meet the following minimum specifications:
   - 4x 25 Gbps Ethernet Ports, implemented as two separate dual port NICs
   - Out-of-band Management Port
 
-Note 1: At least 1 network interface must be capable of performing PXE boot and that network must be available to both the Jump / Test Host and each bare-metal server.
+Note 1: At least 1 network interface must be capable of performing PXE boot and that network must be available to both the Jump / Test Host and each physical server.
 
 <a name="4.2.2"></a>
 
 ## 4.2.2 Lab Network Hardware & Topology
 
-Labs that are hosting multiple *Pods* should utilize a leaf / spine topology when interconnecting *Pods* or bare-metal servers.   This is especially important in the cases where the bare-metal servers constituting a *Pod* are not located in the same physical rack or are not connected to the same leaf switch. At least one leaf switch will be provided for each *Pod*, with interface speeds matching the above server specifications.  In this context, the reference to the single leaf refers to the logical appearance of switch, compared to topologies providing 2 or more leaf switches to provide high availability.
+Labs that are hosting multiple *Pods* should utilize a leaf / spine topology when interconnecting *Pods* or physical servers. This is especially important in the cases where the physical servers constituting a *Pod* are not located in the same physical rack or are not connected to the same leaf switch. At least one leaf switch will be provided for each *Pod*, with interface speeds matching the above server specifications.  In this context, the reference to the single leaf refers to the logical appearance of switch, compared to topologies providing 2 or more leaf switches to provide high availability.
 
-Leaf switches must provide interfaces matching the bare-metal server specifications above and northbound (spine connections) of 100 Gbps connections. Spine switches must provide the corresponding 100 Gbps interfaces to each leaf switch.  The minimum requirement is one spine switch.
+Leaf switches must provide interfaces matching the physical server specifications above and northbound (spine connections) of 100 Gbps connections. Spine switches must provide the corresponding 100 Gbps interfaces to each leaf switch.  The minimum requirement is one spine switch.
 
 The minimum networking configuration must provide at least VLANs to partition the various networks required for the reference implementation deployment and separation of each *Pod*, if multiple *Pods* are hosted within the labs.   
 
--- a/doc/ref_impl/cntt-ri/chapters/chapter04.md
+++ b/doc/ref_impl/cntt-ri/chapters/chapter04.md
@@ -17,7 +17,7 @@ This chapter provides the laboratory hardware requirements needed to deploy the 
 The lab provides a set of bare-metal servers, supported by a dedicated network infrastructure.  Users should be provided access to the servers, their operating systems, and out-of-band management (i.e. IPMI, IDRAC, ILO, etc.).  
 
 <a name="4.2"></a>
-## 4.2 Lab Requirement
+## 4.2 Lab Requirements
 
 Labs are organized into one or more *Pods*, where the *Pod* provides the complete set of resources [servers and network(s)] necessary to support the reference implementation installation and operation.  
 
@@ -47,14 +47,16 @@ Each server shall meet the following minimum specifications:
 Note 1: At least 1 network interface must be capable of performing PXE boot and that network must be available to both the Jump / Test Host and each bare-metal server.
 
 <a name="4.2.2"></a>
-### 4.2.2 Networking Hardware
 
-The minimum networking configuration must provide at least VLANs to partition the various networks required for the reference implementation deployment.  At least one top of rack switch will be provided for each *Pod*, with interface speeds matching the above server specifications.
+## 4.2.2 Lab Network Hardware & Topology
 
-<a name="4.3"></a>
-## 4.3 Lab Network Topology
+Labs that are hosting multiple *Pods* should utilize a leaf / spine topology when interconnecting *Pods* or bare-metal servers.   This is especially important in the cases where the bare-metal servers constituting a *Pod* are not located in the same physical rack or are not connected to the same leaf switch. At least one leaf switch will be provided for each *Pod*, with interface speeds matching the above server specifications.  In this context, the reference to the single leaf refers to the logical appearance of switch, compared to topologies providing 2 or more leaf switches to provide high availability.
 
-The *Pod* network topology should provide at least 2 networks with preallocated IP addressing schemes for the *out-of-band management* network and the *Public* network.  The *Public* network should be able to access the public Internet.  At least IPv4 addresses must be available, ideally IPv6 addresses should also be supported.
+Leaf switches must provide interfaces matching the bare-metal server specifications above and northbound (spine connections) of 100 Gbps connections. Spine switches must provide the corresponding 100 Gbps interfaces to each leaf switch.  The minimum requirement is one spine switch.
+
+The minimum networking configuration must provide at least VLANs to partition the various networks required for the reference implementation deployment and separation of each *Pod*, if multiple *Pods* are hosted within the labs.   
+
+The *Pod* network topology should provide at least 2 networks with preallocated IP addressing schemes for the *out-of-band management* network and the *Public* network.  The *Public* network must be able to reach / access the public Internet.  At least IPv4 addresses must be available, ideally IPv6 address space should also be supported.
 
 Remote users shall access the lab via a VPN gateway, that shall also provide basic security and separation from the public Internet.  Both the *Public* and *out-of-band management* networks shall be accessible through the VPN connection.
 
--- a/doc/ref_impl/cntt-ri/chapters/chapter04.md
+++ b/doc/ref_impl/cntt-ri/chapters/chapter04.md
@@ -7,56 +7,62 @@
 * [4.1 Introduction](#4.1)
 * [4.2 Lab Requirement](#4.2)
 * [4.3 Lab Topology](#4.3)
-* [4.4 Lab HW Spec](#4.4)
-* [4.5 Lab Use Guidelines](#4.5)
+* [4.4 Lab Use Guidelines](#4.4)
 
 
 <a name="4.1"></a>
-## 4.1 Introdcution
+## 4.1 Introduction
+This chapter provides the laboratory hardware requirements needed to deploy the reference implementation.  These requirements represent a minimum set of requirements, where different labs may have hardware or capabilities exceeding these minimum specifications.  
 
+The lab provides a set of bare-metal servers, supported by a dedicated network infrastructure.  Users should be provided access to the servers, their operating systems, and out-of-band management (i.e. IPMI, IDRAC, ILO, etc.).  
 
 <a name="4.2"></a>
 ## 4.2 Lab Requirement
 
-- Controller Nodes:
-  - 3 x
-    - 2x dual-port 10Gbps NIC.
-    - 2.2GHz 14C/28T.
-    - 256GB RAM.
-    - 10TB HDD.
-    - 3.2TB SSD.
-- Compute Node
-  - 10x
-    - 2x dual-port 25Gbps NIC.
-    - 2.2GHz 24C/48T.
-    - 512GB RAM.
-    - 2TB HDD
-    - Supports 2x Nodes for B/N/C flavors
-- Jump/Baremetal manager
-  - 1x
-    - 2x dual-port 10Gbps NIC.
-    - 2.2GHz 14C/28T.
-    - 256GB RAM.
-    - 10TB HDD
-- Networking
-  - 1x Spine Switch
-    - Total: 32x100G
-  - 2x Leaf Switch
-    - 48 x 25/10G
-    - 6 x 100G
-- 1x 48u Rack
-- Cables and Transcievers
+Labs are organized into one or more *Pods*, where the *Pod* provides the complete set of resources [servers and network(s)] necessary to support the reference implementation installation and operation.  
+
+A CNTT compliant *Pod* shall meet the following requirements.
+
+1. One (1) bare-metal server dedicated as a Jump / Test Host
+2. Six (6) bare-metal servers, serving as either compute or controllers
+3. A configured network topology allowing for: Out-of-Band Management, Admin, Public, Private, and Storage Networks
+
+The Jump Host / Test Host system will be utilized to install and manage the operation of the *Pod*.  For example, the system can be used to host the installer software used to deploy and configure OpenStack on the other bare-metal servers.  Once installed, the system may be used to run test automations or it may be used to generate test traffic intended to measure the performance of a VNF operating within the *Pod*.  To support these operations, the bare-metal server provided for the Jump / Test Host shall meet the same hardware requirements as compute / controller, as outline below.  In this context, the system may be considered as a miscellaneous use  or general purpose system.
+
+<a name="4.2.1"></a>
+### 4.2.1 Bare-metal Server Requirements
+Each server shall meet the following minimum specifications:
+
+- **CPU**
+  - 2x CPU sockets, providing 24 cores each, at 2.2 GHz
+- **Memory**
+  - 512 GB RAM
+- **Storage**
+  - 3.2 TB SSD via SATA 6 Gbps
+  - Storage should present as at least 3 or more disks to the OS, allowing for usage as CEPH storage nodes, or similar.
+- **Network Interfaces** (note 1)
+  - 4x 10Gbps Ethernet Ports
+  - Out-of-band Management Port
+
+Note 1: At least 1 network interface must be capable of performing PXE boot and that network must be available to both the Jump / Test Host and each bare-metal server.
+
+<a name="4.2.2"></a>
+### 4.2.2 Networking Hardware
+
+The minimum networking configuration must provide at least VLANs to partition the various networks required for the reference implementation deployment.  At least one top of rack switch will be provided for each *Pod*, with interface speeds matching the above server specifications.
 
 <a name="4.3"></a>
-## 4.3 Lab Topology
+## 4.3 Lab Network Topology
 
+The *Pod* network topology should provide at least 2 networks with preallocated IP addressing schemes for the *out-of-band management* network and the *Public* network.  The *Public* network should be able to access the public Internet.  At least IPv4 addresses must be available, ideally IPv6 addresses should also be supported.
+
+Remote users shall access the lab via a VPN gateway, that shall also provide basic security and separation from the public Internet.  Both the *Public* and *out-of-band management* networks shall be accessible through the VPN connection.
+
+***EDITORS NOTE: Provide a figure showing the basic lab topology, include the VPN, Pod / Servers, Switch, and management / public networks.***
 
 <a name="4.4"></a>
-## 4.4 Lab HW Spec
 
-
-<a name="4.5"></a>
-## 4.5 Lab Use Guidelines
+## 4.4 Lab Use Guidelines
 
 **SETUP & Maintenance**
 
--- a/doc/ref_impl/cntt-ri/chapters/chapter04.md
+++ b/doc/ref_impl/cntt-ri/chapters/chapter04.md
@@ -34,14 +34,14 @@ The Jump Host / Test Host system will be utilized to install and manage the oper
 Each server shall meet the following minimum specifications:
 
 - **CPU**
-  - 2x x86_64 CPU sockets, providing 24 cores each, 44 hyper-threads, at 2.2 GHz
+  - 2x x86_64 CPU sockets (both populated), providing 24 cores each, 48 simultaneous multi-threads (SMT), at 2.2 GHz
 - **Memory**
   - 512 GB RAM
 - **Storage**
   - 3.2 TB SSD via SATA 6 Gbps
   - Storage should present as at least 3 or more disks to the OS, allowing for usage as CEPH storage nodes, or similar.
 - **Network Interfaces** (note 1)
-  - 4x 10Gbps Ethernet Ports
+  - 4x 25 Gbps Ethernet Ports, implemented as two separate dual port NICs
   - Out-of-band Management Port
 
 Note 1: At least 1 network interface must be capable of performing PXE boot and that network must be available to both the Jump / Test Host and each bare-metal server.
@@ -66,7 +66,7 @@ Remote users shall access the lab via a VPN gateway, that shall also provide bas
 
 **SETUP & Maintenance**
 
-OPNFV will facilitate the need for lab procurment, as required, for projects which come into their front door for verification and validation.
+OPNFV will facilitate the need for lab procurement, as required, for projects which come into their front door for verification and validation.
 
 Individual companies that donated a lab would be responsible for setup and maintenance of a community lab. Labs, once setup, will be shared and posted in a wiki https://wiki.opnfv.org/display/pharos/Community+Labs.
 
@@ -76,14 +76,14 @@ The wiki will contain information such as:
 - Contact person
 - Policy info for use
 - Access instructions, include VPN
-- Lab Topolgy
-- Resrouce Availability
+- Lab Topology
+- Resource Availability
 
 **N+1 Labs Use Strategy**
 
 The "lab" concept could mean one physical POD or multiple PODs in a community lab.  In practice, a person requesting a lab can apply for multiple PODs, where POD 1 can be "Lab 1", POD 2+3 can be "Lab 2", etc, all subject to the availability of PODs. Each "Lab" can be used for different purpose such as "Reference Implementation (RI) Lab 1" (POD 1), "RI Lab 2" (POD 2+3), etc.
 
-The requestor for labs needs to know only the number of PODs that are needed.  Then, apply for use of PODs according to the requirement.
+The requester for labs needs to know only the number of PODs that are needed.  Then, apply for use of PODs according to the requirement.
 
 Talk to the Lab Owners if special hardware or topology requirements are needed, such as special NIC or hardware accelerators or if 8 servers in one POD is a different default setting from the lab settings.  The Lab Owner will check to see if lab config changes can be made to accommodate the test need.
 
--- a/doc/ref_impl/cntt-ri/chapters/chapter04.md
+++ b/doc/ref_impl/cntt-ri/chapters/chapter04.md
@@ -34,7 +34,7 @@ The Jump Host / Test Host system will be utilized to install and manage the oper
 Each server shall meet the following minimum specifications:
 
 - **CPU**
-  - 2x CPU sockets, providing 24 cores each, at 2.2 GHz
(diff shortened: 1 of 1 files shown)


lylavoie:
This pull request should fix Issues #403 and #564

doc/ref_impl/cntt-ri/chapters/chapter04.md
```diff
@@ -7,56 +7,62 @@
...
+
+- **CPU**
+  - 2x x86_64 CPU sockets, providing 24 cores each, 44 hyper-threads, at 2.2 GHz
+- **Memory**
+  - 512 GB RAM
+- **Storage**
+  - 3.2 TB SSD via SATA 6 Gbps
+  - Storage should present as at least 3 or more disks to the OS, allowing for usage as CEPH storage nodes, or similar.
+- **Network Interfaces** (note 1)
+  - 4x 10Gbps Ethernet Ports
```
rabi-abdel: RM kind of assume 2x dual 25Gbps.

doc/ref_impl/cntt-ri/chapters/chapter04.md
```diff
@@ -7,56 +7,62 @@
...
+3. A configured network topology allowing for: Out-of-Band Management, Admin, Public, Private, and Storage Networks
+
+The Jump Host / Test Host system will be utilized to install and manage the operation of the *Pod*.  For example, the system can be used to host the installer software used to deploy and configure OpenStack on the other bare-metal servers.  Once installed, the system may be used to run test automations or it may be used to generate test traffic intended to measure the performance of a VNF operating within the *Pod*.  To support these operations, the bare-metal server provided for the Jump / Test Host shall meet the same hardware requirements as compute / controller, as outline below.  In this context, the system may be considered as a miscellaneous use  or general purpose system.
+
+<a name="4.2.1"></a>
+### 4.2.1 Bare-metal Server Requirements
+Each server shall meet the following minimum specifications:
+
+- **CPU**
+  - 2x x86_64 CPU sockets, providing 24 cores each, 44 hyper-threads, at 2.2 GHz
```
pgoyal01: @lylavoie since this is a bare metal server requirement should hyper-threads be specified?

doc/ref_impl/cntt-ri/chapters/chapter04.md
```diff
@@ -7,56 +7,62 @@
...
+A CNTT compliant *Pod* shall meet the following requirements.
+
+1. One (1) bare-metal server dedicated as a Jump / Test Host
+2. Six (6) bare-metal servers, serving as either compute or controllers
+3. A configured network topology allowing for: Out-of-Band Management, Admin, Public, Private, and Storage Networks
+
+The Jump Host / Test Host system will be utilized to install and manage the operation of the *Pod*.  For example, the system can be used to host the installer software used to deploy and configure OpenStack on the other bare-metal servers.  Once installed, the system may be used to run test automations or it may be used to generate test traffic intended to measure the performance of a VNF operating within the *Pod*.  To support these operations, the bare-metal server provided for the Jump / Test Host shall meet the same hardware requirements as compute / controller, as outline below.  In this context, the system may be considered as a miscellaneous use  or general purpose system.
+
+<a name="4.2.1"></a>
+### 4.2.1 Bare-metal Server Requirements
```
rabi-abdel: There is a need to specify different categories of compute (NI, Basic, CI). each can have different specs. what was wrong with the original spec?

rabi-abdel:
@lylavoie for this to be approved by Wednesday next week, we expect all comments to be resolved and addressed by Tuesday Next week at the latest.

doc/ref_impl/cntt-ri/chapters/chapter04.md
lylavoie: I had a request (via email) that hyper-threads be part of the specification / requirements.  I'm ok either with (i.e. with or without them) as long as we can define the number of CPUs and Cores, we will cover most of the hardware bill of material requirements.

doc/ref_impl/cntt-ri/chapters/chapter04.md
lylavoie: There were not differences in the original specs for what was actually different (from a hardware perspective) for the 3 categories.  So, for release 1 or the ref. imp. documentation I'm proposing this simplification.  As the work matures, and we can clearly identify the differences (i.e. hardware needs), we can update these specifications for release 2, etc.  This allows labs to pace the role out of the hardware, etc.

doc/ref_impl/cntt-ri/chapters/chapter04.md
lylavoie: So, 4 ports of 25Gbps Ethernet?  Do we need to be specific to dual port cards?

markshostak: For 4 ports, I'd assume no more than 2 ports/card, for both bandwidth (i.e. bottleneck on the PCIe side) and redundancy (i.e. failure or OOS mtc on one card, does not impact more than 50% of the links) reasons.

doc/ref_impl/cntt-ri/chapters/chapter04.md
markshostak: To clarify and make specific the original logic, the intent was to specify # of physical cores, and to specify that SMT was equipped on the socket (as opposed to HT).

doc/ref_impl/cntt-ri/chapters/chapter04.md
pgoyal01: @rabi-abdel As a practical matter, IMHO, it is highly unlikely that enterprises in their private clouds will have different hardware for the different categories (B, NI, CI) -- except maybe for CI -- as it would lead to operational inefficiencies and will also partition the "cloud" by category.  Public clouds, on the other hand, have the necessary distribution of workload types that can justify the intended hardware by category variations.

doc/ref_impl/cntt-ri/chapters/chapter04.md
lylavoie: So, do we want to require 4 ports, 2 per card, 25 Gbps, and make that the specific requirement, if yes, I'll re-write this and update the pull request.

doc/ref_impl/cntt-ri/chapters/chapter04.md
markshostak: @rabi-abdel To add to the above, there were two relevant notes:

1) For various reasons, including it being part of the original UNH lab proposal, we've aligned that the lab, at least this lab, will be homogeneous, and therefore all Instance Types (IT) will run on the same server config/SKU, be it the concurrently on the same host (i.e. mixed IT types/host) or independently on a host dedicated to a particular IT.

2) There was a lot of discussion around the "jump" server (aka Jump Host (JH)), which is part of the POD, but outside of the NFVI, serving a supervisory role. The anchor driver for the JH is to host the VPN gateway to allow remote user access to the POD across the public internet, for control plane, OOB mgmt and tenant routeable internal and non-routeable networks (i.e. visibility to all networks in the POD). During the discussion, requirements were also raised for a location to host both instrumentation (e.g., Functest, iPerf, etc.) and instrumentation mgmt (e.g., Ixia client and apps of such ilk). The original concept was, when the JH was envisioned to merely host a VPN, why not make it a VM and put all PODs' VPN servers on the same host, rather than dedicating a 44 core, 2 socket server w/ 1/2 a TB of RAM, to perform a function a Raspberry Pi could handle, thus improving the total number of usable infra servers in the lab.
HTH, -Mark

markshostak:
> 
> 
> @lylavoie for this to be approved by Wednesday next week, we expect all comments to be resolved and addressed by Tuesday Next week at the latest.

@rabi-abdel Point of clarification - Who do we want to be authoritative to declare a given comment "resolved". Must it be the originator of the comment, or anyone on the RI team or other?

Thx, -M

doc/ref_impl/cntt-ri/chapters/chapter04.md
rabi-abdel: 2x dual port in my head means: two cards (one per CPU socket). Each card has two parts of 25G.

markshostak: Makes perfect sense. I vote yes.

doc/ref_impl/cntt-ri/chapters/chapter04.md
rabi-abdel: did you mean to say 44 or was is supposed to be 48 with HT?

doc/ref_impl/cntt-ri/chapters/chapter04.md
rabi-abdel: OK happy with the approach of homogeneous approach for now until the need to adjust.

rabi-abdel:
> > @lylavoie for this to be approved by Wednesday next week, we expect all comments to be resolved and addressed by Tuesday Next week at the latest.
> 
> @rabi-abdel Point of clarification - Who do we want to be authoritative to declare a given comment "resolved". Must it be the originator of the comment, or anyone on the RI team or other?
> 
> Thx, -M

Usually the person who asks for the review. but if the author feels he covered it, it is Off or him to hit the magic button and resolve it.

doc/ref_impl/cntt-ri/chapters/chapter04.md
```diff
@@ -7,56 +7,62 @@
...
+  - 3.2 TB SSD via SATA 6 Gbps
+  - Storage should present as at least 3 or more disks to the OS, allowing for usage as CEPH storage nodes, or similar.
+- **Network Interfaces** (note 1)
+  - 4x 10Gbps Ethernet Ports
+  - Out-of-band Management Port
+
+Note 1: At least 1 network interface must be capable of performing PXE boot and that network must be available to both the Jump / Test Host and each bare-metal server.
+
+<a name="4.2.2"></a>
+### 4.2.2 Networking Hardware
```
rabi-abdel: we specified Spine/Leaf switches in the org specs as this is the model assumed in RA 1. I know there are contributions in the pipelines to do that. can we add more information here and can we make sure we adopt Spine/Leaf architectures and have spec the networking devices?

doc/ref_impl/cntt-ri/chapters/chapter04.md
```diff
@@ -7,56 +7,62 @@
...
+
+The minimum networking configuration must provide at least VLANs to partition the various networks required for the reference implementation deployment.  At least one top of rack switch will be provided for each *Pod*, with interface speeds matching the above server specifications.
 
 <a name="4.3"></a>
-## 4.3 Lab Topology
+## 4.3 Lab Network Topology
```
rabi-abdel: is it possible to provide a topology diagram?

markshostak: @rabi-abdel We can, but I don't think it should be critical path to approval of the server h/w config. Thoughts?

doc/ref_impl/cntt-ri/chapters/chapter04.md
markshostak: @rabi-abdel By networking devices, you're referring to the leaf and spine switches? Are you envisioning make and model, or physical description (i.e. ports/speeds). In the interest of neutrality, I'd recommend the latter.

@ASawwaf Included in the leaf/spine spec, we need to align on the/any required APIs we require. For example, must provide open/public APIs for <list>, where $list might include NC, P4, OF or whatever. 

Again, networking should not gate server approval, except for NICs, which has already been resolved (above), and vice versa.

doc/ref_impl/cntt-ri/chapters/chapter04.md
lylavoie: @markshostak I will change this to SMT, to avoid the proprietary term.
@rabi-abdel  Yes, this should be 48 (I promise I really can do basic math when not distracted)

doc/ref_impl/cntt-ri/chapters/chapter04.md
lylavoie: It's likely not practical to require the switches support a superset of SDN protocols.  It would be hard to fine a switch supporting both P4 and OpenFlow for example.  I'm not sure how we provide flexibility to the RA/RM definitions here, while documenting the requirements for lab hardware and implementation.  Unless the Reference Implementation will make a specific choice, and the lab is then required to support that choice.

rabi-abdel: @markshostak it is the latter

doc/ref_impl/cntt-ri/chapters/chapter04.md
rabi-abdel: as long as the networking devices are more specified, I think a diagram Is not very important.

doc/ref_impl/cntt-ri/chapters/chapter04.md
markshostak: Agreed, and what I listed was to illustrate what I meant by APIs. I'm guessing the fabric will use distributed control, so really only NC would be relevant, if that, but the point is we need to align on what we require. I don't have a specific preference at this time, but Ahmed will.

lylavoie: On the leaf / spine topology.  I'm struggling with how to fit that into the context of the Pod definition.  For example, the Pod, as defined only has 7 total servers.  To provide an independent leaf / spine per pod would really require 2 leaf and 2 spine switches, which doesn't scale.  We could provide guidance for labs like LaaS to ensure the network(s) supporting multiple Pods meets the leaf / spine topology requirement.

rabi-abdel: @ASawwaf agree with you on the need for this, but for the purpose of this PR, I think it is ok to only specify high level specs with the acknowledgment that we will be specifying them in more details as the specs matures and evolve.

doc/ref_impl/cntt-ri/chapters/chapter04.md
markshostak: @pgoyal01 Can we mark this sub-thread as Resolved?

doc/ref_impl/cntt-ri/chapters/chapter04.md
markshostak: @rabi-abdel Are you referring to the API thing, when you say only specify high level specs?

rabi-abdel: @markshostak Yes API but pithing drilling to too much into details. (i.e supporting BGP protocols, encapsulation, etc). @ASawwaf probably can help specifying that.

rabi-abdel: @lylavoie I don't think the intent here is to provide an independent leaf/spine to each pod. the pod is a special use case in my views. but the lab as a whole, we require a spine/lead architecture. I hope that makes sense.

lylavoie: @rabi-abdel that makes complete sense, and I can add text to make that clear.  I just don't want to set an expectation that each pod (which is kinda of a "unit of access" or a lab deliverable to the users) gets an independent set of lead/spine switches, as that would be unrealistic.

markshostak: @lylavoie Lincoln, I agree with Rabi. A reasonable question for us to run down for you is, is a single spine sufficient or are 2 required. However, the driver for that decision is irrespective of the number of PODs. The intent is to have a shared fabric and definitely **not** to have a dedicated physical leaf/spine for _each_ POD. Whether to have 2 spines and 2 leafs in total, or 1 and 1 (provided the leaf has sufficient ports for all of the servers), is where I see the questions. 

To that point, as is probably obvious, the driver for the NIC config we've discussed (2x dual 25Gb (4 ports)), is to have 1 port from each NIC go to one leaf and the other port from the same NIC go to the other leaf. That would imply we should have a min of 2 physical leaf switches (more, added in pairs, if required to provide the number of ports needed for the number of servers in the infra).

Whether to connect the leaf switches with 1 or 2 spines, is a different question. I'll defer to Ahmed @ASawwaf , for direction there.

Irrespective, as Rabi implied, all of the servers would share the fabric described above. Coexistence will be provided by isolation and encapsulation protocols. We can pursue this further in a dedicated fabric discussion, which is out of scope here, but we'd like to hear how LaaS and/or your IOL currently provides isolation, if there's any encapsulation and if you're fabric is L2 or L3 based. From there we can determine the actual fabric topology and design (and get Rabi his diagram), and what, if any enhanced support will be required from the LaaS automation. I can envision ways we can leverage additional LaaS features, and ways we could get by without.

Does that answer your question/clarify the vision?

So, we owe you:
1) number of spines
2) if leaf switches need to be paired
@ASawwaf Can you give us your input on the 2 above? Thanks

lylavoie: @markshostak with respect to LaaS, we current provide isolation at Layer 2 using VLANs. 

For the question about "paired" leaf switches, I could go either way.  It might make sense, if we are going to start testing HA or failing a link, etc.  It's not really practical to test failing a switch in this environment, if the switch is supporting multiple pods and you're "neighbors" aren't expecting to test that case at that minute.  

Same for number of spine switches.  I think could be fine to role out with one for release 1, and add more as a requirement, if we see the need.

ASawwaf: @lylavoie @markshostak @rabi-abdel 

The answers for the above 2 points will be on 
1-  # of servers  
2- Use case

for lab purpose, if we have one RACK only, we can leave with Single SW covers both Spine/Leaf
but this SW should have enough capacity 

To have the right decision 
@liyingyjy can share what we have or what can we secure in terms of 
- number of servers 
- Which server vendor that we have / or will have
- which NIC card vendor 
- Which SW vendors

@liyingyjy for VLAN separation, it is not a recommendation to have VLAN inside the Fabric 



My comments and recommendation 
-
- We need to lock down the use case or SUT that should covers ( i.e Just VNF onboarding to match the profile or VNF performance, or ...
- To have a Small mimic of what the operator should implement 

- # of server : 3 Server ( OSP + SDN-C ) + 1 Server ( Jump+Installer ) + 3 CEPH + Computer/workload : at least to 2 per profile ( 6 Server ) 
- Sequential VNF testing may be a challenge due to the limitation of workload servers 
- NIC Card either be 10G / 25G based on the use case ( for scalability 25G ) , Just want to unify the Fabric ( either 10G or 25G )
- The server should be NIC-NUMA aligned for VNF at least 
- We can host the Openstack controller with SDN -C ( as @markshostak mentioned before ) 
- RAM : Should be balanced between CPU ( no matter if it is 32G / 16G ) but should be balanced
- CEPH : we need to have min 3 CEPH Cluster, my recommendation to go with SSD for High IOPS req.
- We need to cover Jump server, Installer Server and Tools servers ( Servers or VM )

ASawwaf: I don't know why bold font appears  :):)

markshostak: My 2 pence - let's focus on the immediate objective - **align on the server hardware config, so we can approve it on Wednesday's TSC call** and once that's locked, for extra credit we may provide some guidelines on the fabric. 

I agree with all of the statements above, but there's no way we can resolve them in time for the TSC call. The latest server configuration proposed seems like it meets our gross requirements. No one I'm aware of has identified an [open] issue with the current config. @kedmison identified a risk WRT clock speed, in that the minimum speed (or if clock speed will be specified) hasn't been determined yet, hence if these servers end up not meeting the final clock spec, that could be an issue. I'd like to know the impact if this risk is realized. My working assumption is even if they're not quite as fast as the spec, they'll still be faster than if we miss the window and don't get servers. @lylavoie also needs to weigh in with if this is something UNH is comfortable with. Kelvin, Lincoln, thoughts? Thx, -M

mf4716:
There are a few open conversations which need closure if @markshostak @rabi-abdel if you can please review and close, if possible.

rabiabdel:
@lylavoie is it possible please try to address as much as possible from the points above before Wednesday? I understand not all points can be addressed though on time for this.

markshostak:
> 
> 
> There are a few open conversations which need closure if @markshostak @rabi-abdel if you can please review and close, if possible.

Sure. Which are you referring to?

mf4716:
hi @markshostak - there are two open conversation which require resolution: 1) the ask about a topology diagram, and 2) h/w config.  Open in GitHub for details.

fuqiao1:
I see this lab requirement still use the 7-node structure of Pharos pods, with 1 jump server, and 6 control/compute nodes. Do we think this number is sufficient enough for future testing and verification? shall we expend the concept of Pod to any number of servers according to the requriement of testing? We can add discription of server number in the installer discription file to specify this?

doc/ref_impl/cntt-ri/chapters/chapter04.md
lylavoie: I've added text for the switches (i.e. leaf / spine).

For the CPU topics, I'm happy with the current definition, as the minimum specs for release 1.  It's challenging to use other specifications, while still remaining general in the requirements (i.e. non product specific).

rabi-abdel:
Happy with the content so far! offcourse this can evolve, but as a minimum requirements, this seems to me in a good shape. @markshostak @kedmison @pgoyal01 @iangardner22 @ASawwaf @fuqiao123 can you all approve (this is a minimum spec to unlock NUH and spec can evolve to add more details in future.

doc/ref_impl/cntt-ri/chapters/chapter04.md
```diff
@@ -7,60 +7,68 @@
...
-## 4.3 Lab Topology
+## 4.2 Lab Requirements
 
+Labs are organized into one or more *Pods*, where the *Pod* provides the complete set of resources [servers and network(s)] necessary to support the reference implementation installation and operation.  
 
-<a name="4.4"></a>
-## 4.4 Lab HW Spec
+A CNTT compliant *Pod* shall meet the following requirements.
+
+1. One (1) bare-metal server dedicated as a Jump / Test Host
```
iangardner22: Where will the "Foundation" node be hosted - e.g. the cloud builder (Director/Undercloud in the case of Red Hat)?

markshostak: Hi Ian, That's an installer-specific question. So far we've been using Airship (1.x), which does not require a dedicated node for the purpose, it uses any (i.e. you pick it) of the future control nodes to perform that function. In the case of RH (OOO), my understanding is the foundation node is ephemeral, so are you thinking it requires a dedicated node?, or it could just as easily be a VM on the jump host. Either way, that's a decision that the person implementing the particular RI would need to make.

markshostak:
@kedmison @fuqiao123 @ASawwaf @rabi-abdel and all, We believe we've addressed all of the open issues relevant to the minimum server configuration. Are we ready to approve this PR and approve the server config on tomorrow's TSC call?

rabiabdel:
> @kedmison @fuqiao123 @ASawwaf @rabi-abdel and all, We believe we've addressed all of the open issues relevant to the minimum server configuration. Are we ready to approve this PR and approve the server config on tomorrow's TSC call?

Yes we should be able to approve this during TSC call tomorrow. All others, please make sure to provide your comments here if you have any in here before tomorrow to allow Lincoln to address them b4 tomorrow's TSC call.

doc/ref_impl/cntt-ri/chapters/chapter04.md
ASawwaf: @markshostak @iangardner22 , for RH tripleO , The undercloud node, it can be VM based on any serevr as it is just used for deployment ( Just :  8-core , with minimum 16 GB of RAM and 100GB available disk)

ASawwaf: @markshostak @liyingyjy , my recomedation not to mention Bare Metal , to be able to utilize the servers for any VM deployment ( undercloud , Tools ,..)

markshostak: Replaced Bare Metal w/ Physical

doc/ref_impl/cntt-ri/chapters/chapter04.md
```diff
@@ -7,60 +7,68 @@
...
+<a name="4.2.1"></a>
+### 4.2.1 Physical Server Requirements
+Each server shall meet the following minimum specifications:
+
+- **CPU**
+  - 2x x86_64 CPU sockets (both populated), providing 24 cores each, 48 simultaneous multi-threads (SMT), at 2.2 GHz
+- **Memory**
+  - 512 GB RAM
+- **Storage**
+  - 3.2 TB SSD via SATA 6 Gbps
```
kedmison: 3.2TB is under-provisioned as per the size templates in the RM. 
A host of this CPU spec could support (ignoring, for now, reserved CPUs) 96 ".small" tenants for CPU and RAM, but only about 50 ".small" tenants for storage, when factoring in a 3x replication for ceph storage.
If the ".small" vCPU to disk ratio changed from 1:20 to 1:10, then there would be enough disk in the 3.2 TB, leaving ~150GB for the host disk, roughly.  This is just a simple calculation and ceph experts should weigh in here too.

doc/ref_impl/cntt-ri/chapters/chapter04.md
```diff
@@ -7,60 +7,68 @@
...
-
-<a name="4.3"></a>
-## 4.3 Lab Topology
+## 4.2 Lab Requirements
 
+Labs are organized into one or more *Pods*, where the *Pod* provides the complete set of resources [servers and network(s)] necessary to support the reference implementation installation and operation.  
 
-<a name="4.4"></a>
-## 4.4 Lab HW Spec
+A CNTT compliant *Pod* shall meet the following requirements.
```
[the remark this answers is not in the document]
markshostak: We can loosen that up. It should apply to minimum lab requirements. Are you ok if we update this after the go/no-go?

doc/ref_impl/cntt-ri/chapters/chapter04.md
```diff
@@ -7,60 +7,68 @@
...
+
+1. One (1) physical server dedicated as a Jump / Test Host
+2. Six (6) physical servers, serving as either compute or controllers
```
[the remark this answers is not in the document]
markshostak: This is considered the minimum configuration. UNH uses LaaS, so the size can be specified when the (v)Pod is instantiated. We can clarify it by specifying this is the minimum config. Are you ok if we update this after the go/no-go?

doc/ref_impl/cntt-ri/chapters/chapter04.md
```diff
@@ -7,60 +7,68 @@
...
+3. A configured network topology allowing for: Out-of-Band Management, Admin, Public, Private, and Storage Networks
+
+The Jump Host / Test Host system will be utilized to install and manage the operation of the *Pod*.  For example, the system can be used to host the installer software used to deploy and configure OpenStack on the other physical servers.  Once installed, the system may be used to run test automations or it may be used to generate test traffic intended to measure the performance of a VNF operating within the *Pod*.  To support these operations, the physical server provided for the Jump / Test Host shall meet the same hardware requirements as compute / controller, as outline below.  In this context, the system may be considered as a miscellaneous use  or general purpose system.
+
+<a name="4.2.1"></a>
+### 4.2.1 Physical Server Requirements
+Each server shall meet the following minimum specifications:
+
+- **CPU**
+  - 2x x86_64 CPU sockets (both populated), providing 24 cores each, 48 simultaneous multi-threads (SMT), at 2.2 GHz
```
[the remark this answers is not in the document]
markshostak: This is the _minimum_ configuration, so it's Broadwell. If we make it Cascade, it will disqualify the only running labs we currently have, which are Broadwell-based.

doc/ref_impl/cntt-ri/chapters/chapter04.md
```diff
@@ -7,60 +7,68 @@
...
+- **Memory**
+  - 512 GB RAM
```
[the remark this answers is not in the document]
markshostak: As it turns out, the memory is a given. The question that came up yesterday is, how is the memory distributed? We're following up on that, and can add some informative language describing how UNH configured the memory, and possibly some general guidelines on distributing memory across the various controllers and buses, but I don't think we should be prescriptive at this level. I suggest we refer readers to their server manufacturer's documentation for information on how to optimize the performance of their specific machine.

doc/ref_impl/cntt-ri/chapters/chapter04.md
```diff
@@ -7,60 +7,68 @@
...
 
-<a name="4.4"></a>
-## 4.4 Lab HW Spec
+A CNTT compliant *Pod* shall meet the following requirements.
+
+1. One (1) physical server dedicated as a Jump / Test Host
+2. Six (6) physical servers, serving as either compute or controllers
+3. A configured network topology allowing for: Out-of-Band Management, Admin, Public, Private, and Storage Networks
+
+The Jump Host / Test Host system will be utilized to install and manage the operation of the *Pod*.  For example, the system can be used to host the installer software used to deploy and configure OpenStack on the other physical servers.  Once installed, the system may be used to run test automations or it may be used to generate test traffic intended to measure the performance of a VNF operating within the *Pod*.  To support these operations, the physical server provided for the Jump / Test Host shall meet the same hardware requirements as compute / controller, as outline below.  In this context, the system may be considered as a miscellaneous use  or general purpose system.
```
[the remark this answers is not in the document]
markshostak: Agreed, running the test host instrumentation as an Infra workload, which I believe is what you're discussing, is an option and is not precluded by the language above. As far as sharing the same OS, there's no hard requirement there (i.e. you can put it in a VM), but generally speaking, I don't see a large risk here. There's a good amount of portability in this particular space, and additionally, I believe (I'll double check) you have flexibility in the OS you choose for the jump host.

doc/ref_impl/cntt-ri/chapters/chapter04.md
markshostak: Agreed. Ahmed also raised this issue, and we're working to address. Our thinking is, worst case, if we can't resolve it before the approval, we'll try to amend it subsequent to the approval. Failing that, it will have to be deferred to next year. Again, our conclusion is it's sub-optimal, but there's still quite a bit we can do with it, as in, quite possibly almost everything we want to test, so we rather not make this critical path to the lab, if we can avoid it.

doc/ref_impl/cntt-ri/chapters/chapter04.md
iangardner22: @markshostak @ASawwaf - points noted and yes I am happy for it to be a VM - just may be we name the "jumphost" something like "utility host" and then we can use it for anything including jump, Director, Puppet-master, Ansible etc...

markshostak:
> 
> 
> All good - happy with the comments provided. Just one suggestion on the naming of "jump host" could be "utility host" and that covers a wider set of use cases

Thanks for the comments. I agree on the jump host naming. We actually inherit that from an upstream project (Pharos), so probably best to keep consistent w/ their naming, but if you want to drive it, I certainly wouldn't object! :-)

doc/ref_impl/cntt-ri/chapters/chapter04.md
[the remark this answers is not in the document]
kedmison: I think we should insert some of the caveats from the governance call here; potentially points 3,4,5.  The document shouldn't imply that these specs are yet ready for use.

From the [Governance call minutes on 2019-11-21](https://wiki.lfnetworking.org/pages/viewpage.action?pageId=25363522) 

> AI - Approved by Governance with caveats:
> 1. This requires Technical Steering discussion and approval of specs.
> 2. This is a UNH owned system.
> 3. UNH would need to understand there is risk because RI1 specs not finalized, and may change.
> 4. CNTT not legally bound to specs.
> 5. Stipulate needs to only be used at UNH.

doc/ref_impl/cntt-ri/chapters/chapter04.md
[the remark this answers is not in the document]
kedmison: Yes, we can update after go/no-go.  
It has implications for network infrastructure and cabling, as the servers are going to be physically grouped in pods with less than full non-blocking to other pods.  Deploying larger pods will require lab techs to re-deploy network cables, potentially switches, etc.

doc/ref_impl/cntt-ri/chapters/chapter04.md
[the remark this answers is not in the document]
kedmison: I think I am a bit confused. I thought this was to specify something that $FirstLabLocation would purchase, not to align with an existing hardware specification.

doc/ref_impl/cntt-ri/chapters/chapter04.md
```diff
@@ -7,60 +7,68 @@
...
 
+Labs are organized into one or more *Pods*, where the *Pod* provides the complete set of resources [servers and network(s)] necessary to support the reference implementation installation and operation.  
 
-<a name="4.4"></a>
-## 4.4 Lab HW Spec
+A CNTT compliant *Pod* shall meet the following requirements.
+
+1. One (1) physical server dedicated as a Jump / Test Host
+2. Six (6) physical servers, serving as either compute or controllers
+3. A configured network topology allowing for: Out-of-Band Management, Admin, Public, Private, and Storage Networks
```
markbeierl: I realize this is already closed, but just noticed: we need to standardize on what we call the networks.  The terminology here differs from RA Ch 4, S 4.2.3.2 (https://github.com/cntt-n/CNTT/blob/master/doc/ref_arch/openstack/chapters/chapter04.md#4232-high-level-logical-network-layout)

mf4716: @markbeierl, hey Mark, for small edits, you can open an issue & PR, make the change and we can merge quickly.  For this RI Ch04 change, we just make the change, reference the RA Ch4 alignment, and we're done.  If it turns out we need dialogue we can keep the issue open & have that post-Snezka.

markbeierl: Do we have these networks and what goes over them defined anywhere already?

mf4716: @markbeierl, only the [RA Ch4 S 4.2.3.2](https://github.com/cntt-n/CNTT/blob/master/doc/ref_arch/openstack/chapters/chapter04.md#4232-high-level-logical-network-layout) which you referenced, and what's published on the Intel POD wikis for [POD10](https://wiki.opnfv.org/display/pharos/Intel+POD10) and [POD15](https://wiki.opnfv.org/display/pharos/Intel+POD15).  

Team, is/are there another source(s) for this info?

PR #665, opened 2019-11-21
[RM Ch02] Test Labler (testing only)

rabi-abdel:


rabi-abdel:
it didn't work. bummer

PR #675, opened 2019-11-22
[RA2 Ch3] Clarification of the compute part of RA2 Ch3 ToC

CsatariGergely:
This change contains some clarifications to the ToC part of
RA2 Ch3.


rabiabdel:
@tomkivlin I am sorry your review got dismissed by Github because I added a dummy commit to test the labeler. You need to review again. It is a new configuration we added to github.

rabiabdel:
@CsatariGergely I googled it and the reason is you are using your local fork for this Pull request which causes an access token issue (instead of creating a branch directly of the master). You can ignore the error for now but it might be a good practice to create a branch of the CNTT master for the PR (this way you will always be on sync anyway)

tomkivlin:
> @tomkivlin I am sorry your review got dismissed by Github because I added a dummy commit to test the labeler. You may want to review/approve again. It is a new configuration we added to github.

No problem.

PR #681, opened 2019-11-22
[GOV REA Ch08] Recruitment and Adoption Framework & Strategy

### LINKED ISSUE #680: [Gov Ch08] Recruiting, Engagement, & Adoption
**Scope**
Recruit operators & suppliers  to bring diverse perspective and contributions into the community
Drive membership engagement activities to address community resource demands
Engage industry leaders with Common NFVI value proposition to drive adoption rate
Drive adoption roadmap with VNF Vendors

rabi-abdel:
Fixes #680 
Needed before https://github.com/cntt-n/CNTT/issues/750 can be fixed


rabi-abdel:
> 8.2.2 VNF Evolution generation and 8.3 CNF Evolution Generation have the same contents (and rightly so), so I propose to merge both sections as VNF/CNF Evolution Generation
> Typo in line below Fig 3: "Less and less" should read "less and less"
> I propose to stop using "compromises/exceptions" and stick to "exceptions", which is an industry common term.
> We need to explain also who and how decides whether an exception is "big" in the definition of "non-compliant".

@walterkozlowski fixed the spelling mistake and used exceptions only as suggested. in regards to the CNF/VNF: the reason why they are separate is that they might have different evolution generations (likely they will). in case they ended up being the same, I suggest we come back and merge them.

doc/gov/chapters/chapter09.md
```diff
@@ -27,7 +27,7 @@
...
+It is vitally important for CNTT to have working solution from infrastructure vendors and mature VNFs/CNFs designs from application vendors that is compliant to CNTT specifications. It is also understood that, in some areas, the industry might not have solutions that are fully aligned with CNTT requirements. Therefore, an evolution or transitional strategy, comprised of one or more exceptions and/or transitions is required to address technology that does not presently conform to CNTT mandates, and hence requires explicit direction to prescribe how the situation will be treated in the present, as well as in the future.
```
pgoyal01: Typo: "... vendors that is compliant to CNTT ..." -- "are"

doc/gov/chapters/chapter09.md
```diff
@@ -27,7 +27,7 @@
...
+It is vitally important for CNTT to have working solution from infrastructure vendors and mature VNFs/CNFs designs from application vendors that is compliant to CNTT specifications. It is also understood that, in some areas, the industry might not have solutions that are fully aligned with CNTT requirements. Therefore, an evolution or transitional strategy, comprised of one or more exceptions and/or transitions is required to address technology that does not presently conform to CNTT mandates, and hence requires explicit direction to prescribe how the situation will be treated in the present, as well as in the future. 
 
 This transition plan will inform application designers how RC and ultimately OVP will react when encountering such technologies during the qualification process, including flagging warnings and potentially errors which could prevent issuance of an OVP badge.
```
pgoyal01: Documented and allowed exceptions should not prevent the issuance of the "right" badge (consistent, compliant, ...).

doc/gov/chapters/chapter09.md
```diff
@@ -0,0 +1,121 @@
...
+
+<a name="8.2.1"></a>
+### 8.2.1 Rationale
+
+It is vitally important for CNTT to have working solution from infrastructure vendors and mature VNFs/CNFs designs from application vendors that is compliant to CNTT specifications. It is also understood that, in some areas, the industry might not have solutions that are fully aligned with CNTT requirements. Therefore, an evolution or transitional strategy, comprised of one or more exceptions and/or transitions is required to address technology that does not presently conform to CNTT mandates, and hence requires explicit direction to prescribe how the situation will be treated in the present, as well as in the future. 
+
+This transition plan will inform application designers how RC and ultimately OVP will react when encountering such technologies during the qualification process, including flagging warnings and potentially errors which could prevent issuance of an OVP badge.
+
+The transitional plan has the following elements (as shown in **Figure 3** below):
+- **VNF/CNF Evolution Generations**: Defines the evolution metric from Non-Conformant to Fully Conformant.
```
kedmison: I think this VNF/CNF Evolution Generations concept is can be supported through a mechanism we have and one that I believe we should have:
1. Versions of the RM/RA documents themselves and 
1. leveraging/introducing generations of profiles, which are needed anyway as new generations of hardware get introduced.  

When wanting to change the compliance statement for a feature (e.g. move SR-IOV from conformant to non-compliant), a new document generation can be introduced that removes or marks this capability differently.  When changes that are visible to the VNF are required to make this transition happen, a new profile can be defined during this new document creation.

This enables RC teams to know exactly what to test and award status to for compliance to RM/RA version x, and to mark compliance differently for VNFs a document version x+1.  
e.g. this VNF is RA-1 V3 compliant, but RA1 V4 non-compliant, because it uses SR-IOV.

Example use case: A viable SR-IOV alternative has been identified, but needs a new driver in the guest to accomplish the task.
Pre-conditions: 
- Existing profile B1 has identified SR-IOV as a viable option.
- Existing workloads run on B1 instances (e.g. B1.large)

Steps:
- CNTT updates and issues documentation to identify new profile B2, dropping allowance for SR-IOV, and adding new driver option as SR-IOV alternative. Existing profile B1 remains defined, but marked as deprecated.
- VNF vendors find that existing VNFs are now meeting a stronger warning during VNF validation, indicating pending deprecation of B1 instances.  
- VNF vendors being migration of VNFs to use/support both B1 and B2 instances, so that VNFs can be deployed in operators that have executed the transition and those that have not yet done so.
- Operators deploy compute nodes meeting B2 specification, while leaving existing B1 nodes in place to serve the existing workloads.
- Operators indicate their intent and readiness to no longer accept B1 based workloads and thus the need to migrate VNFs using B1 to using B2 instead.
Operators receive/acquire new VNF versions from VNF vendors, that leverage B2 instances and thus have migrated away from B1 instance.
- As B1 workloads are migrated to B2, operators can re-configure and re-deploy (where possible) the B1 compute nodes to meet the B2 configuration needs.

Key aspects: 
- allows gradual migration/evolution of VNF workloads from B1 to B2 while continuing to place the right workload on the matching compute nodes
- allows for operators to deploy B2 configurations on somewhat differing schedules

Changes that would support this approach:
- identifying in this document that CNTT document versions can be used to indicate permitted exceptions, and when these exceptions begin to be deemed non-compliant
- adding support for profile generations in RM

doc/gov/chapters/chapter09.md
```diff
@@ -0,0 +1,121 @@
...
+* [8.2 Transitional Plan](#8.2)
+  * [8.2.1 Rationale](#8.2.1)
+  * [8.2.2 VNF Evolution Generations](#8.2.2)
+  * [8.2.3 CNF Evolution Generations](#8.2.3)
+  * [8.2.4 Policies](#8.2.4)
+* [8.3 Adoption Strategy](#8.3)
+* [8.4 Adoption Roadmap](#8.4)
+  * [8.4.1 CNTT Roadmap](#8.4.1)
+  * [8.4.2 VNF Adoption Roadmap](#8.4.2)
+  * [8.4.3 CNF Adoption Roadmap](#8.4.3)
```
markshostak: How this relates to Adoption is a bit unclear to be. Is this the right/most intuitive word?

doc/gov/chapters/chapter09.md
```diff
@@ -0,0 +1,121 @@
...
+
+<a name="8.4.2"></a>
+#### 8.4.2 VNF Adoption Roadmap
+
+| Generation | Details | Date | Notes |
+|------------|----------------|------------|-------|
+| Gen1 | As per Technical Specification |  |  |
+| Gen2 | As per Technical Specification |  |  |
+| Gen3 | As per Technical Specification |  |  |
+| Gen4 | As per Technical Specification |  |  |
```
markshostak: Having multiple phases of generations for VNFs with the same name (e.g., Gen1, Gen2, etc.), pertaining to the same rls of CNTT, but differentiated by the technology they pertain to, seems very confusing. For example, Gen 1 for SRIOV might correspond to CNTT Rls 3, while Gen 1 for feature x-y-z may pertain to CNTT Rls 4 and so on. To keep it simple, I'd consider aligning to a normalized chronology, referenced by each phase of given technology.

hdamker:
Actually the additional concept of "generations" for VNFs/CNFs in addition of CNTT releases and profiles seems to be quite complex to me and I tend to agreee with https://github.com/cntt-n/CNTT/pull/681#discussion_r359305823 that this can lead to arbitrary combinations of exceptions. Keep in mind that the use of an exceptions by a VNF/CNF implies also that the NFVI operator has to support these exceptions on infrastructure layer (with all consequences e.g. to operational processess).

The proposal by @kedmison in https://github.com/cntt-n/CNTT/pull/681#discussion_r359135599 is worth a discussion ... which might be a clearer way to communicate from NFVI operator to VNF/CNF vendors which "exceptional" profiles are supported or not supported.

BTW: There is the principle in RM Chapter 1.3. 

> "VNFs that are designed to take advantage of NFVI accelerations shall still be able to run without these accelerations, however with the understanding that there will be potential performance impacts."

 ... therefore my understanding would be that "exceptional" profiles might provide e.g. additional acceleration options for VNFs not adapted to preferred profiles.

rabi-abdel:
Due to the complex of this PR and the number of files impacted, I will close this PR and create a different one targeting the framework. (https://github.com/cntt-n/CNTT/pull/893). comments added here will still be reflected in there.

PR #693, opened 2019-11-26
[RA2 Ch3]: Container networking initial content

### LINKED ISSUE #676: [RA2 Ch3]: Container Networking Services
Add content to Container Networking Services chapter of Ch3

CsatariGergely:
This change adds describes the requirements for container networking.

Fixes #676


iawells:
This section will need to address the problems that container networking will solve as well as solutions and that isn't mentioned in the breakdown in the change Gergely proposes.  It's not a question of 'the CNI does this' but one of 'these are the features we require *and* this is how the CNI measures up, and therefore we also need XXX'.

iawells:
The reason I say this is because of this first comment:

> Becouse `req.inf.ntw.01` requires the architecture to support CNI and `req.inf.ntw.16` requires the capability to attach several network interfaces to the pods the architecture must suport a CNI metaplugin/CNI multiplexer.

Firstly: the requirements are wrong.  We do not only need to support multiple interfaces in the same namespace.  The requirements are broader than that.  We need to support multiple interfaces, yes, but they may be presented as passthrough interfaces not attached to the namespace if they're being used for PCI passthrough; and they may not be in the same routing domain (network VRF) however they ar presented, so putting them in the same network namespace is wrong.

Secondly, the conclusion is also wrong.  The CNI is not a good presentation model for providing such interfaces as the CNI is intended to provide access to a single 'network' by its definition (connects a container to a 'network' per the CNI definition and the network is always 'default' according to the way k8s calls CNIs).  We are generally asking for multiple interfaces because packets are treated differently depending on the interface being used, and, for instance, SRIOV interfaces will not all be attached to the same 'network' and we will normally want to use several of them with different forms of attachment.

I would rather we discussed what would be appropriate to solve our problems here (which need stating differently, and I realise on closer reading I was wrong in the previous comment, and that I should look at the requirement reference and fix that) and ensure we're solving the correct problem before we choose anything that is a solution.

Speaking as someone who creates platforms and writes VNFs, I don't think the choice made here is a good fit for delivering CNF networking.  It's what has been done to date, but in all cases where I've seen it it requires a ton of container privilege to use, and that would prevent the platform from being usable for multiple VNFs simultaneously.

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -59,13 +59,31 @@
...
+
+The low latency, high throughput networks for handling the user plane traffic require the capability to use an user space networking technology, like VPP or DPDK. In case of DPDK the usage of SR-IOV VF-s is recommended.
+
+As `req.inf.ntw.14` mandates the architecture must enable the integration of different SDN solutions via their respective CNI integration.
+
+The architecture must support networking for telecom equipments in an environment where the networks of the CNF-s are set up by the network adminisrators of the telecom operator. This is why, as `req.inf.ntw.10` requires, the architecture must provide a set of abstract management API-s to manage the network connectivity of the CNF pods.
+The API must support multiple tennants and must require elevated acces rights to manipulate infrastructure related API objects as these operations require reconfiguration of the physical network infrastructure.
+
+To fullfill the requirements of `req.inf.acc.02` the architecture must support the usage of device plugins via the Device Plugin API, also the alignement of the devices and the CPU topology must be supported using the Topology Manager.
```
tomkivlin: typo alignement > alignment

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -59,13 +59,31 @@
...
 ### 3.2.2 Container Networking Services
 
-> This chapter should discuss networking related considerations of container network services, like:
-> * Ingress and any limitations on protocol or raw interface support
-> * Dependencies on external loadbalancers
-> * Regular service mesh related considerations
-> * Network Service Mesh related considerations
-> * Number of interfaces and IP addresses to be supported by a pod
-> * IPv6 single and dual stack needs
+Becouse `req.inf.ntw.01` requires the architecture to support CNI and `req.inf.ntw.16` requires the capability to attach several network interfaces to the pods the architecture must suport a CNI metaplugin/CNI multiplexer.
```
tomkivlin: typo Becouse > Because
typo suport > support

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -59,13 +59,31 @@
...
+
+A CNI metaplugin/CNI multiplexer is capable to attach several interfaces, using different other CNI plugins to a pod. The different network charastheristics of the interfaces, which require different networking technologies require different CNI plugins.
```
tomkivlin: typo (add second comma) "interfaces, using different other CNI plugins to" > "interfaces, using different other CNI plugins, to"
typo charastheristics > characteristics

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -59,13 +59,31 @@
...
+
+Inter node communication required by `req.inf.ntw.02` must be served by a simple CNI using overlay technology, like Flannel.
+
+There are two types of low latency and high throughput networks required by `req.inf.ntw.04`. Network used for signalling traffic are more demanding than what an overlay network can handle, but still does not need the usage of user space networking. Due to the nature of signalling protocols used this type of netorks require the NAT-less communication stated by `req.inf.ntw.03`. Due to the combination of these two requirements networks with this charactheristics must be served by a CNI plugin with IPVLAN or MACVLAN suport.
```
tomkivlin: "Network used for signalling traffic are more demanding than what an overlay network can handle" - I don't disagree, but we should call out the reasons why, I think.
typo netorks > networks
typo charactheristics > characteristics
typo suport > support
"must be served by a CNI plugin with IPVLAN or MACVLAN suport" - should we describe why IPVLAN and MACVLAN support? What is it about them that makes them work? What are the alternatives that aren't appropriate?

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -59,13 +59,31 @@
...
+
+The low latency, high throughput networks for handling the user plane traffic require the capability to use an user space networking technology, like VPP or DPDK. In case of DPDK the usage of SR-IOV VF-s is recommended.
```
tomkivlin: SR-IOV - do we need to specify that we don't recommend PCI Pass Through?  (not my strong point, simply parroting what I hear @rabi-abdel shouting into the phone every day 😉)

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -59,13 +59,31 @@
...
+
+As `req.inf.ntw.14` mandates the architecture must enable the integration of different SDN solutions via their respective CNI integration.
```
tomkivlin: Should we elaborate on what the impact on the RA might be - i.e. that the SDN solution provides the information required to install the CNI into the cluster, at which point the SDN controller is managing the pod network via Kubernetes APIs?  Or do we expect the CNI pods in the cluster to talk back to the SDN controller instead/as well?

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -59,13 +59,31 @@
...
+
+The architecture must support networking for telecom equipments in an environment where the networks of the CNF-s are set up by the network adminisrators of the telecom operator. This is why, as `req.inf.ntw.10` requires, the architecture must provide a set of abstract management API-s to manage the network connectivity of the CNF pods.
+The API must support multiple tennants and must require elevated acces rights to manipulate infrastructure related API objects as these operations require reconfiguration of the physical network infrastructure.
```
tomkivlin: typo tennants > tenants

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: Should we include a link to the Topology Manager?

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -59,13 +59,31 @@
...
+
+To fullfill the requirements of `req.inf.acc.02` the architecture must support the usage of device plugins via the Device Plugin API, also the alignement of the devices and the CPU topology must be supported using the Topology Manager.
+
+The architecture must support both IPv4, IPv6 and dual stack interfaces of the workloads.
+ > How about IPv6 support of the control plane?
+
+As Kubernetes Ingress and Services have no support for all the protocols needed in telecommunication environments and their capacity is limited the architecture must enable the usage alternative load balancers, like external or built into the application.
```
tomkivlin: Should we list the protocols, or link to a comprehensive list?  A non-Telco reader might ask "like what?".

tomkivlin: typo "the usage" > "the usage of"

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -59,13 +59,31 @@
...
+
+Well known service meshes have no support for protocols required by telecom applications therefore their suport is not required in the architecture.
```
tomkivlin: typo "suport" > "support"

tomkivlin: I get what you mean with this, but I wonder if we should clarify that most "service meshes" are actually "application service meshes" and therefore deal with layer 7 - application - protocols like HTTP only.  New projects looking at network service meshes (i.e. L3/4) could end up being included in the scope of this RA, perhaps?

tomkivlin: Oh, just seen the next line 😆 ...

doc/ref_arch/kubernetes/chapters/chapter03.md
iawells: See above: need not be served by anything involving the CNI, for one, and macvlan support is certainly not a requirement for control communications.  Iimagine a BGP RR that insisted on this specific requirement.  Now imagine it if it was using bonded uplinks...

doc/ref_arch/kubernetes/chapters/chapter03.md
iawells: I don't know that those words are being used to mean what they're supposed to mean (the reason being they're abused on a daily basis, so it's not terribly much a surprise).

SR-IOV: how we split one PCI device into multiple devices, which is an option

PCI passthrough: in the container context, handing a PCI device into a process's direct control.  It may be an SR-IOV derived device.  It might not be.  It might be a device that's not capable of SR-IOV.

It is likely a very bad idea to hand an CNF control of the PF of a device that *does* support SR-IOV, because that gives it the power to make new PCI devices - which will end up with the kernel, and that's almost certainly a really bad idea.  On the other hand, it might make sense to hand a trusted service control of that device so that it can manipulate the VFs for SDN purposes.  I believe (and I think Trevor Cooper knows more about this than me) that the DPDK libraries and tools support controlling PFs in this way.

Also, diverging from the networking discussion for a moment, whatever means of providing passthrough devices we choose might want pulling out from the exclusively networking section since it's also a reasonable thing to do with accelerators and presenting the PCI devices to the container consistently and in a way that's passthrough-friendly should be the same, for usability reasons.

Finally, I would take the comment about the way those devices are used as an instruction to presentation.  The models we see in the wild today tend to stick even PCI NICs into the kernel netns as network devices and they have to be removed from the netns to be consumed by a DPDK application and rebound to a passthrough kernel device driver.  It would be better if we forgot they were networking devices at all and presented them to the container as a vfio device file, since that device can be used with no privilege to the consuming container.

doc/ref_arch/kubernetes/chapters/chapter03.md
iawells: I think 'application service meshes' is indeed required.

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -59,13 +59,31 @@
...
-> * Dependencies on external loadbalancers
-> * Regular service mesh related considerations
-> * Network Service Mesh related considerations
-> * Number of interfaces and IP addresses to be supported by a pod
-> * IPv6 single and dual stack needs
+Becouse `req.inf.ntw.01` requires the architecture to support CNI and `req.inf.ntw.16` requires the capability to attach several network interfaces to the pods the architecture must suport a CNI metaplugin/CNI multiplexer.
+
+A CNI metaplugin/CNI multiplexer is capable to attach several interfaces, using different other CNI plugins to a pod. The different network charastheristics of the interfaces, which require different networking technologies require different CNI plugins.
+
+Inter node communication required by `req.inf.ntw.02` must be served by a simple CNI using overlay technology, like Flannel.
```
msidana: Can we rephrase this to ".. served by using CNI plugins like calico, flannel, weave etc.
The current sentence says "using overlay" which gives a wrong impression that overlay must be used.

doc/ref_arch/kubernetes/chapters/chapter03.md
rabi-abdel: Re ... _It would be better if we forgot they were networking devices at all and presented them to the container as a vfio device file, since that device can be used with no privilege to the consuming container._

Agree.

rabi-abdel: _In case of DPDK the usage of SR-IOV VF-s is recommended_
we need to clarify what that means. if it 100% OK to use SR-IOV to present a NIC as multiple NICs, but it is not ok to expose that VF to container without any abstraction via a standard interface (memif, others ...) we need to pick a way.

doc/ref_arch/kubernetes/chapters/chapter03.md
CsatariGergely: I'm okay to rephase it, but it would be good if we could generalize it somehow. Let's try CNI plugins with basic networking capabilities...

doc/ref_arch/kubernetes/chapters/chapter03.md
CsatariGergely: @tomkivlin the reason is the throughput need and the need for NAT less communication. 

@iawells from what is currently available today I see no other solutions than CNI based ones, but maybe I miss something.

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -59,13 +59,31 @@
...
-> * Dependencies on external loadbalancers
-> * Regular service mesh related considerations
-> * Network Service Mesh related considerations
-> * Number of interfaces and IP addresses to be supported by a pod
-> * IPv6 single and dual stack needs
+Because `req.inf.ntw.01` requires the architecture to support CNI and `req.inf.ntw.16` requires the capability to attach several network interfaces to the pods the architecture must support a CNI metaplugin/CNI multiplexer.
+
+A CNI metaplugin/CNI multiplexer is capable to attach several interfaces, using different other CNI plugins, to a pod. The different network charastheristics of the interfaces, which require different networking technologies require different CNI plugins.
+
+Inter node communication required by `req.inf.ntw.02` must be served by a CNI plugin with basic networking capabilities, like Calico or Flannel.
```
Levovar: I wouldn't use the word "basic". Calico does provide some extra features.
Also I think every CNI should be capable of setting-up interface through which inter-node communication is possible :)

Instead I would mention that there needs to be at least one network available in the cluster which complies with the default K8s networking assumptions, and the CaaS infra components are reachable through it.
And this network can be an overlay network too, as latency does not matter here.

Also I would not mention Flannel, as it is quickly falling out of grace due to it not supporting NetworkPolicies. Calico / Weave / Cilium are generally seem as "better" options for this network

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -59,13 +59,31 @@
...
+
+There are two types of low latency and high throughput networks required by `req.inf.ntw.04`. Network used for signalling traffic are more demanding than what an overlay network can handle, but still does not need the usage of user space networking. Due to the nature of signalling protocols used this type of networks require the NAT-less communication stated by `req.inf.ntw.03`. Due to the combination of these two requirements networks with this characteristics must be served by a CNI plugin with IPVLAN or MACVLAN support.
```
Levovar: technically SR-IOV can be also used for kernel networking

doc/ref_arch/kubernetes/chapters/chapter03.md
Levovar: I don't think there is a clear cut answer to this, and TBH both are possible.
There aren't really many SDN solutions out there.

I prefer the second, cause I don't believe all interfaces must be managed by an SDN. But the first options is also equally viable IMO.

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -59,13 +59,31 @@
...
+
+As `req.inf.ntw.14` mandates the architecture must enable the integration of different SDN solutions via their respective CNI integration.
+
+The architecture must support networking for telecom equipments in an environment where the networks of the CNF-s are set up by the network adminisrators of the telecom operator. This is why, as `req.inf.ntw.10` requires, the architecture must provide a set of abstract management API-s to manage the network connectivity of the CNF pods.
+The API must support multiple tenants and must require elevated acces rights to manipulate infrastructure related API objects as these operations require reconfiguration of the physical network infrastructure.
+
+To fullfill the requirements of `req.inf.acc.02` the architecture must support the usage of device plugins via the Device Plugin API, also the alignment of the devices and the CPU topology must be supported using the [Topology Manager](https://kubernetes.io/docs/tasks/administer-cluster/topology-manager/).
```
Levovar: do we want to include hugepages into this mix?

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -59,13 +59,31 @@
...
+
+The architecture must support both IPv4, IPv6 and dual stack interfaces of the workloads.
+ > How about IPv6 support of the control plane?
```
Levovar: the network manager mainly manages application networks.
i.e. most of the control plane is actually running in host network

so while control plane IPv6 is a valid requirement, IMO it is not related to CNI / network management topic

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -59,13 +59,31 @@
...
+
+As Kubernetes Ingress and Services have no support for all the protocols needed in telecommunication environments (Diameter, SIP, LDAP, etc) and their capacity is limited the architecture must enable the usage of alternative load balancers, like external or built into the application.
```
Levovar: might want to menton Egress too?

Levovar: Also do we want to take a stance on Ingress / Egress / Services?
Option1: we don't need them
Option2: We need them, but we require the architecture to support them only on the "cluster network"
Option3: We need them, and we require the architecture to make them work over all interfaces

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -59,13 +59,31 @@
...
+
+Well known service meshes are "application service meshes" and therefore deal with layer 7 - application - protocols, like HTTP, only therefore their support is not required in the architecture.
```
Levovar: punctuation issue

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -59,13 +59,31 @@
...
+
+As the most recent version of Network Service Mesh is v0.1.0 at the time of specifying RA2 in Snezka release it is not required in the architecture.
```
Levovar: I'm not in favour singling out any one particular solution in an architecture document, especially if it is not even required

doc/ref_arch/kubernetes/chapters/chapter03.md
Levovar: ohmm, bifurcated drivers? using SR-IOV devices for kernel networking purposes?
yes, with _Intel_ NICs for _DPDK_ purposes what you write is true, but it is hardly the only use-case

you keep saying it is not okay to expose VFs for containers through the Devices abstraction layer, but I still haven't seen the exact reasons why

Levovar:
> We need to support multiple interfaces, yes, but they may be presented as passthrough interfaces not attached to the namespace if they're being used for PCI passthrough; and they may not be in the same routing domain (network VRF) however they ar presented, so putting them in the same network namespace is wrong.

Well, this is your opinion ;)
I'm not saying it is good or bad, but I wouldn't go out and present it as a universal truth when this option was never discussed with the community, and no reference implementation of it exists today (or have you put together something since ONS?)

> Secondly, the conclusion is also wrong. The CNI is not a good presentation model for providing such interfaces as the CNI is intended to provide access to a single 'network' by its definition (connects a container to a 'network' per the CNI definition and the network is always 'default' according to the way k8s calls CNIs).

CNI spec supports multiple interfaces FYI.
Still, I generally agree with you that CNI shall not be the interface through which operators / CNFs manage networks.
But I don't think that is written. What Gergő writes is that the architecture must support, read, implement CNI because this is a K8s RA.
But it does not say CNI needs to be the network management interface. And yes I agree, it totally should not be.
Maybe we could clarify this further in the doc.

 > We are generally asking for multiple interfaces because packets are treated differently depending on the interface being used, and, for instance, SRIOV interfaces will not all be attached to the same 'network' and we will normally want to use several of them with different forms of attachment.

True, totally agree, just want to expose this to other readers: this is not an SR-IOV specific attribute. All interfaces, be provisioned with whatever network driver can possibly belong to different VLANs, subnets etc.
IPVLAN / MACVLAN / different bridges etc.

> Speaking as someone who creates platforms and writes VNFs, I don't think the choice made here is a good fit for delivering CNF networking. It's what has been done to date, but in all cases where I've seen it it requires a ton of container privilege to use, and that would prevent the platform from being usable for multiple VNFs simultaneously.

The choices presented in this doc while might not be optimal, and I totally want to discuss this more with you -maybe on a panel ;)- are the choices which are possibly the best _today_.
IMO this RA is for NFVI which one can go out and implement today.
That being said, future definitely can look different.
But again, just to expose this to other readers: this networking paradigm totally does not require extra privileges from CNFs. Or if yes, please point out why?
As someone who is also creating platforms and VNFs :), I can say that at least in our VNFs we could do away without providing them privileges while kinda following what is described so far in this doc.

doc/ref_arch/kubernetes/chapters/chapter03.md
LiuyingCMRI: I think "basic" here means more about realizing network connectivity of inter node. Overlay network can be included. And also I think we have no need to mention a specific CNI plugin here just to describe its capabilities. For network policy should we make another network requirement if CNI needs to support?

tomkivlin:
@CsatariGergely could you review the changes requested above please?

CsatariGergely:
> 
> 
> This section will need to address the problems that container networking will solve as well as solutions and that isn't mentioned in the breakdown in the change Gergely proposes. It's not a question of 'the CNI does this' but one of 'these are the features we require _and_ this is how the CNI measures up, and therefore we also need XXX'.

Here we need to build from what we already have. So if we do not have XXX we should not build on it.

doc/ref_arch/kubernetes/chapters/chapter03.md
CsatariGergely: I've made an attempt to correct this.

doc/ref_arch/kubernetes/chapters/chapter03.md
CsatariGergely: Can we conclude this debate? @rabi-abdel any responses to @Levovar -s comments?

doc/ref_arch/kubernetes/chapters/chapter03.md
CsatariGergely: Why not? If we need it we need it.

doc/ref_arch/kubernetes/chapters/chapter03.md
CsatariGergely: I have no strong opinion on this. Any suggestions?

tomkivlin: I think we definitely need to define that Services are needed for inter-Pod communications, for sure.  For incoming communication, I think we need to avoid describing "how" something is deployed (i.e. "you must use Ingress" or "you must use LoadBalancer Service type", etc.).  I think it could be fair to say something like @CsatariGergely already has, but adding that those external load balancers must be managed via Kubernetes API objects (e.g. cloud-controller for LoadBalancer Services, or CRDs, etc.). That way we're specifying an interface that can be tested against.
What do you think?

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -63,7 +63,7 @@ Because `req.inf.ntw.01` requires the architecture to support CNI and `req.inf.n
 
 A CNI metaplugin/CNI multiplexer is capable to attach several interfaces, using different other CNI plugins, to a pod. The different network charastheristics of the interfaces, which require different networking technologies require different CNI plugins.
 
-Inter node communication required by `req.inf.ntw.02` must be served by a CNI plugin with basic networking capabilities, like Calico or Flannel.
+Inter node communication required by `req.inf.ntw.02` must be served by a CNI plugin which complies with the default K8s networking assumptions.
```
pgoyal01: is it "assumptions" or (default k8s networking) "model?"

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -74,16 +74,14 @@ As `req.inf.ntw.14` mandates the architecture must enable the integration of dif
...
+To fullfill the requirements of `req.inf.acc.02` the architecture must support the usage of device plugins via the Device Plugin API, also the alignment of the devices, CPU topology and hugepages must be supported using the [Topology Manager](https://kubernetes.io/docs/tasks/administer-cluster/topology-manager/).
 
 The architecture must support both IPv4, IPv6 and dual stack interfaces of the workloads.
  > How about IPv6 support of the control plane?
 
-As Kubernetes Ingress and Services have no support for all the protocols needed in telecommunication environments (Diameter, SIP, LDAP, etc) and their capacity is limited the architecture must enable the usage of alternative load balancers, like external or built into the application.
+As Kubernetes Ingress, Egress and Services have no support for all the protocols needed in telecommunication environments (Diameter, SIP, LDAP, etc) and their capacity is limited the architecture must enable the usage of alternative load balancers, like external or built into the application.
```
pgoyal01: Typo: "... capacity is limited the architecture ..." --- "... capacity is limited, the architecture ..."

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -74,16 +74,14 @@ As `req.inf.ntw.14` mandates the architecture must enable the integration of dif
...
 
-Well known service meshes are "application service meshes" and therefore deal with layer 7 - application - protocols, like HTTP, only therefore their support is not required in the architecture.
-
-As the most recent version of Network Service Mesh is v0.1.0 at the time of specifying RA2 in Snezka release it is not required in the architecture.
+Well known service meshes are "application service meshes" and therefore deal with layer 7 - application - protocols (eg.: HTTP), only therefore their support is not required in the architecture.
```
pgoyal01: Suggest change from:
"Well known service meshes are "application service meshes" and therefore deal with layer 7 - application - protocols (eg.: HTTP), only therefore their support is not required in the architecture."

to 

"The well known service meshes are "application service meshes" and deal with the application layer 7 protocols (eg.: HTTP) only. Therefore, their support is not required in the architecture."

doc/ref_arch/kubernetes/chapters/chapter03.md
CsatariGergely: In some cases the telco protocols are still used in inter-Pod communication. A load balancer is a typical example when the LB pod only decides the next hop, but does not do any protocol translation. 
Managing external LB-s via Kubernetes API objects I will add.

doc/ref_arch/kubernetes/chapters/chapter03.md
CsatariGergely: As we agreed on the [2019-12-19 - CNTT - RA2 - All call](https://wiki.lfnetworking.org/display/LN/2019-12-19+-+%5BCNTT+-+RA2+-+All%5D+-+Meeting+Agenda+and+Minutes) I will remove the recommendation for SR-IOV with DPDK. I will also add a note that it is still under discussion along with an other note stating that an infra can still offer the possibility without being CNTT un compliant.

doc/ref_arch/kubernetes/chapters/chapter03.md
CsatariGergely: I agree, but the true question is if we would like to use SR-IOV for kernel networking. From the debates around the usage of SR-IOV in general my conclusion is that we should rely on SR-IOV only in places where we really need to.

doc/ref_arch/kubernetes/chapters/chapter03.md
CsatariGergely: It is more like assumptions. like what Kubernetes assumes to happen on the given network.

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -90,13 +90,34 @@ When it comes to Kubernetes, the important interface that we need to consider fo
...
-> This chapter should discuss networking related considerations of container network services, like:
-> * Ingress and any limitations on protocol or raw interface support
-> * Dependencies on external loadbalancers
-> * Regular service mesh related considerations
-> * Network Service Mesh related considerations
-> * Number of interfaces and IP addresses to be supported by a pod
-> * IPv6 single and dual stack needs
+Because `req.inf.ntw.01` requires the architecture to support CNI and `req.inf.ntw.16` requires the capability to attach several network interfaces to the pods the architecture must support a CNI metaplugin/CNI multiplexer.
+
+A CNI metaplugin/CNI multiplexer is capable to attach several interfaces, using different other CNI plugins, to a pod. The different network charastheristics of the interfaces, which require different networking technologies require different CNI plugins.
```
pgoyal01: Typo: charastheristics -- characteristics

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -90,13 +90,34 @@ When it comes to Kubernetes, the important interface that we need to consider fo
...
+
+Inter node communication required by `req.inf.ntw.02` must be served by a CNI plugin which complies with the default K8s networking assumptions.
```
pgoyal01: Have we specified the kubernetes network model or can we provide a reference?

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -90,13 +90,34 @@ When it comes to Kubernetes, the important interface that we need to consider fo
...
+
+There are two types of low latency and high throughput networks required by `req.inf.ntw.04`. Network used for signalling traffic are more demanding than what an overlay network can handle, but still does not need the usage of user space networking. Due to the nature of signalling protocols used this type of networks require the NAT-less communication stated by `req.inf.ntw.03`. Due to the combination of these two requirements networks with this characteristics must be served by a CNI plugin with IPVLAN or MACVLAN support.
```
pgoyal01: Typo: "... used this type of networks require ..." to either "... used, these type of networks require ..."

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -90,13 +90,34 @@ When it comes to Kubernetes, the important interface that we need to consider fo
...
+
+The low latency, high throughput networks for handling the user plane traffic require the capability to use an user space networking technology.
+
+> Note: An infrastructure can provide the possibility to use SR-IOV with DPDK as an additional feature and still be compliant with CNTT.
+
+> Editors note: The possibility to SR-IOV for DPDK is under discussion.
+
+As `req.inf.ntw.14` mandates the architecture must enable the integration of different SDN solutions via their respective CNI integration.
+
+> Note: SDN solution can manage the pod networks via the Kubernetes API or the CNI integrations of the SDN solution can have communication with the SDN solution.
```
pgoyal01: suggest delete "... can have communication with the SDN solution"

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -90,13 +90,34 @@ When it comes to Kubernetes, the important interface that we need to consider fo
...
+
+The architecture must support networking for telecom equipments in an environment where the networks of the CNF-s are set up by the network adminisrators of the telecom operator. This is why, as `req.inf.ntw.10` requires, the architecture must provide a set of abstract management API-s to manage the network connectivity of the CNF pods.
```
pgoyal01: Typo: equipments -- equipment
         adminisrators -- administrators

Suggest slight rewrite:
"The architecture must support telecom equipment networking where the CNF networks  are set up by the operator's network administrators."
(Note: the networks cannot be set up by the telecom operator but the cloud operator)

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -90,13 +90,34 @@ When it comes to Kubernetes, the important interface that we need to consider fo
...
+The API must support multiple tenants and must require elevated acces rights to manipulate infrastructure related API objects as these operations require reconfiguration of the physical network infrastructure.
+
+To fullfill the requirements of `req.inf.acc.02` the architecture must support the usage of device plugins via the Device Plugin API, also the alignment of the devices, CPU topology and hugepages must be supported using the [Topology Manager](https://kubernetes.io/docs/tasks/administer-cluster/topology-manager/).
```
pgoyal01: suggest change: "... also the alignment of the devices ..." --- "... and the alignment of the devices ..."

rabiabdel:
@CsatariGergely would mind please fix those spelling mistakes spotted by @pgoyal01 above. ity should be merged afterwords.

doc/ref_arch/kubernetes/chapters/chapter03.md
CsatariGergely: It is not the network model, but what Kubernetes assumes about the behavior of a network.

CsatariGergely:
> 
> 
> I approve the changes proposed but I am of the opinion that the whole Chapter in general and Networking section in particular need much more content.
> Specifically, i think that this Chapter requires a diagram with proper layering, where each of the layers can be provided from a different open source project or a different distro vendor. I showed an example of a diagram like this (in a different context) in my review of the Integration section.

Can you please add the link of the diagram also here?

doc/ref_arch/kubernetes/chapters/chapter03.md
Levovar: agree with Gergely. it is a far fetched assumption IMO that all CNF internal communication is stateless, and happens in L7

Issue #592, opened 2019-11-13
[RM] Glossary - EPA definition  


ASawwaf:

CsatariGergely:
I agree that we should define the specific technologies and not the umbrella term EPA.

hdamker:
Required features need to be carefully selected -- we need to be adhere to decisions taken already in RM and RA1 and not make RA2 more convenient for lift&shift of PNFs directly into CNFs.

ASawwaf:
@petorre  , if i define it i will pull the definition from the intel site, so if you can put an extensive definition for EPA it will be great , can you support ?


I will start with: Enhanced Platform Awareness (EPA) represents a methodology and a related suite of changes across multiple layers of the orchestration stack targeting intelligent platform capability, configuration & capacity consumption. EPA features include Huge Pages support, NUMA topology awareness, CPU pinning, integration with OVS-DPDK, support for I/O Pass-through via SR-IOV, and many others.

CsatariGergely:
> 
> 
> @petorre , if i define it i will pull the definition from the intel site, so if you can put an extensive definition for EPA it will be great , can you support ?
> 
> I will start with: Enhanced Platform Awareness (EPA) represents a methodology and a related suite of changes across multiple layers of the orchestration stack targeting intelligent platform capability, configuration & capacity consumption. EPA features include Huge Pages support, NUMA topology awareness, CPU pinning, integration with OVS-DPDK, support for I/O Pass-through via SR-IOV, and many others.

This "many others" what we should close down in this definition.

ASawwaf:
I can list what I know :) 

-Platform Quality of Service (PQOS)
-Advanced Encryption Standard New Instructions (AES-NI)
- VT-d
-Memory Huge-pages
-CPU Frequency scaling/Turbo
-NUMA awareness
-Cache QoS
-Intel Data Direct I/O (DDIO)
-Data Protection Technology (with Advanced Encryption Standard New Instructions) [AES-NI]  
- Platform Protection Technology (with Trusted Execution Technology [TXT])
-SR-IOV
-CPU Pinning 
-  Quick Assist Technologies ( QAT ) 
- Huge Pages 
- DPDK
- IO based NUMA scheduling

@CsatariGergely @tomkivlin is it enough :)

@petorre @trevgc your comment and contribution is appreciated here

hdamker:
Hi,

[...]

Issue #590, opened 2019-11-13
[RM/RA/RI/RC] Document Hardware Objectives, Guidelines and Approach


markshostak:
**Issue:** As discussed in multiple forums, CNTT currently specifies hardware in multiple documents (e.g., RM, RA and RI), making alignment challenging. Further, when RI/CIRV (or CSPs) are procuring hardware, there needs to be an objective mechanism to ensure the candidate configuration(s) will support the CNTT designs.

- Due to business realities (and focus on VNF abstraction), CNTT cannot specify exact h/w requirements (e.g. SKU)
- Due to rate of change in technology, even if CNTT did specify SKUs, the spec would rapidly become stale and maintenance would be impractical
- Therefore, as discussed on RM call, we propose creating a set of Guidelines that
- - **i. Meet the documented CNTT objectives for specifying h/w**
- - **ii. Are decoupled from the technology de jour and therefore do not require constant updating**
- - iii. **Can assist users in identifying h/w that will support CNTT infra designs**
- - iv. **Provide validation (i.e. objectively assess if proposed SKUs are aligned w/ CNTT platform needs)**

**Objectives/Drivers:** Drivers are the key to this process, and will be considered carefully. Please add your proposed objectives.
- Deterministic and predictable performance
- **YOUR PROPOSED DRIVER GOES HERE**
- Reduced operational complexity of infrastructure [Trevor]

**Approach:**  The following has been discussed independently in various forums. The intent here, is to bridge all of the forums to achieve a CNTT-level consensus. This is just a starting point to drive the discussion. It is expected to be modified/enhanced (i.e. this is not edict, it's catalyst). 

As previously discussed, the current thinking is to:
1. Identify CNTT's objectives/drivers for specifying h/w in the first place (i.e. what are we trying to achieve by detailing the h/w)
2. Derive a set of clear, concise and usable Guidelines from the Objectives
3. Identify where the Guidelines should live and PR them to those locations
4. Gain consensus on a series of related questions (see below)

**Related Questions:** These questions have come up in RI/CIRV activities. They may be fully CSP discretionary, a hard CNTT requirement or somewhere in between. Irrespective, a supporting the reason, justification or rationale should be identified for each conclusion. In no particular order:  
A. What/can node types/functions can be combined in production (and why)
  - Compute?
  - Storage?
  - SDN Controller?

B. How to best accommodate hetero and homogeneous server farms 
C. Should CNTT recommend interface speed guidelines? (if so, what are they?)
[...]

PR #699, opened 2019-11-26
[RM Glossary] Enhanced Platform Awareness" (EPA)  definition

ASawwaf:
"Enhanced Platform Awareness" (EPA) represents a methodology to get workload placement understand underlying platform to deliver improved and deterministic application performance, and IO throughput. It enables fine-grained matching of workload requirements to platform capabilities, prior to launching a VM or Container.

### DIFF
--- a/doc/ref_model/chapters/glossary.md
+++ b/doc/ref_model/chapters/glossary.md
@@ -94,3 +94,5 @@ To help guide the reader, this glossary provides an introduction to the terminol
 
 - **Virtualised Infrastructure Manager (VIM)**: responsible for controlling and managing the **NFVI** compute, storage and network resources.
 - **NFV Orchestrator (NFVO)**: manages the VNF lifecycle and **NFVI** resources (supported by the **VIM**) to ensure an optimised allocation of the necessary resources and connectivity.
+
+


ASawwaf:
to added in the Glossary

markshostak:
Hmm... Is this filling a specific gap in our existing h/w profile schema? Where is the matching envisioned to occur? Does this move us father towards abstraction? I'm concerned it may work counter to our abstraction objective.

ASawwaf:
@markshostak, as agreed in the closed issue with a definition, reflected in the glossary 
and we need to work more on RM Ch4 to make sure the specific EPA covered under a profiles 
So my opinion, to open issue / PR for this @markshostak

rabiabdel:
@ASawwaf EPA cab be used to detect the profike type + extension available to allow VIM to understand compute nodes capabilities. Certainly EPA must not used to map workload requirements to specific compute node. It should not be exposed to workloads. We might need to adjust the definition to remove any ambiguity

rabiabdel:
> @markshostak, as agreed in the closed issue with a definition, reflected in the glossary
> and we need to work more on RM Ch4 to make sure the specific EPA covered under a profiles
> So my opinion, to open issue / PR for this @markshostak

@ASawwaf the current definition might not reflect the conclusion of the issue referenced.

ASawwaf:
> > @markshostak, as agreed in the closed issue with a definition, reflected in the glossary
> > and we need to work more on RM Ch4 to make sure the specific EPA covered under a profiles
> > So my opinion, to open issue / PR for this @markshostak
> 
> @ASawwaf the current definition might not reflect the conclusion of the issue referenced.

https://github.com/cntt-n/CNTT/issues/592#issuecomment-558795306 , I grab the final definition that petr added  and we are ok with it 

let's update here

markshostak:
> 
> 
> @rabi-abdel wrote:
_Certainly EPA must not used to map workload requirements to specific compute node capabilities (this all should be bundled within the profile itself) ._ 

Intel wrote*:
"_For workloads requiring particular CPU and/or I/O capabilities, EPA helps Open-
Stack assign virtual machines (VMs) to run on the optimal platforms._"

They sound pretty close to me. Can you help me clarify the difference?

*From: EPA white paper cover page, https://networkbuilders.intel.com/docs/ice-house-openstack-enhanced-platform-awareness.pdf

tomkivlin:
Reading through this, and the associated issue (which has already been closed?? Is that a mistake?), I wonder what the value of having a definition for EPA is - do we need one?
My meaning is, we have already agreed we are not going to refer to EPA in requirements, instead listing the individual features per requirement. So when will we use EPA in the document?

tomkivlin:
Also, there don't seem to be any changes in this PR?

ASawwaf:
> Reading through this, and the associated issue (which has already been closed?? Is that a mistake?), I wonder what the value of having a definition for EPA is - do we need one?
> My meaning is, we have already agreed we are not going to refer to EPA in requirements, instead listing the individual features per requirement. So when will we use EPA in the document?

@tomkivlin  , i Close it intentionally, though that we are all algin 
the conclusion it : we need to put just a definition of EPA in RM glossary and for individual features, it will be covered when needed 

i hope it clear, and sorry for the inconvenience , we can cover it in more details in our RM call

tomkivlin:
@ASawwaf sorry what I meant was, normally an Issue is closed once the PR that address it is merged, not before. But anyway, it's not that important 😄 

> we need to put just a definition of EPA in RM glossary 
My wordy question before was, do we need to do this?  If we're not using the term elsewhere in the documentation, why add it to the glossary?

ASawwaf:
> @ASawwaf sorry what I meant was, normally an Issue is closed once the PR that address it is merged, not before. But anyway, it's not that important 
> 
> > we need to put just a definition of EPA in RM glossary
> 
> My wordy question before was, do we need to do this? If we're not using the term elsewhere in the documentation, why add it to the glossary?

@tomkivlin thanks for clarification for the issue closure process,  regarding do we need to define it or not?

My opinion we need to put it in our glossary, as you mentioned before it is marketing terminology but it heavily used in Openstack , NFVO  ( ONAP / OSM ) , as well we need to define this EPA in any VNF descriptors ( VNF-D ) 

So i believe we can put the above definition, and after that, we can select any specific features during defining RAs

make sense or still not ? :)

markshostak:
> 
> 
> @tomkivlin thanks for clarification for the issue closure process, regarding do we need to define it or not?
> 
> My opinion we need to put it in our glossary, as you mentioned before it is marketing terminology but it heavily used in Openstack , NFVO ( ONAP / OSM ) , as well we need to define this EPA in any VNF descriptors ( VNF-D )

Why are we bringing MANO into this now? That seems to describe exposing h/w specific details to the MANO. Isn't that counter to our objective of driving abstraction, and is starting to sound more like HPA, than EPA.

Typically we add terms when we use them. If this is opening the door to breaking the VNF abstraction, we should think carefully before adding it, especially if we haven't used it yet, as it implies we support breaking abstraction, conceptually. In the SRIOV thread, people are vehemently opposed to breaking the abstraction, so it's unclear why the same logic doesn't seem to apply here.

ASawwaf:
> > @tomkivlin thanks for clarification for the issue closure process, regarding do we need to define it or not?
> > My opinion we need to put it in our glossary, as you mentioned before it is marketing terminology but it heavily used in Openstack , NFVO ( ONAP / OSM ) , as well we need to define this EPA in any VNF descriptors ( VNF-D )
> 
> Why are we bringing MANO into this picture? That seems to describe exposing h/w specific details to the MANO. Isn't that counter to our objective of driving abstraction, and is starting to sound more like HPA, than EPA.
> 
> Typically we add terms when we use them. If this is opening the door to breaking the VNF abstraction, we should think carefully before adding it, especially if we haven't used it yet, as it implies we support breaking abstraction, conceptually. In the SRIOV thread, people are vehemently opposed to breaking the abstraction, so it's unclear why the same logic doesn't seem to apply here.

@markshostak Hello Mark , We are not bringing MANO , my point that "EPA" is well-known terminology and it used in many contexts so we need to define it "if we use it in any context during  drafting RM, RA , RI, RC" if not so no need to define  

what did you think Gents @markshostak @rabiabdel @tomkivlin @petorre

rabi-abdel:
@ASawwaf @markshostak @tomkivlin @petorre 

Since EPA can potentially be sending a wrong message of allowing hardware specific to be exposed all the way to VNFD (at least from the definitions lying around), I would propose waiting until we have to use the term (possibly in RA docs). if/when used in there,  we can define it just there as EPA will only impact RA and not RM. RM defines profiles with bundle of capabilities, no place for EPA in there IMO

In particular when we talk about **dynamic** profile discovery (it is not a must) and extensions supported by each profile (against would be nice but not a must) we can decide to use EPA for this if it made sense (will impact the interface between NFVI and VIM). If we decide to do that in RA, then there will be potentailly a need to upstream our concepts of using EPA to openstack which will be part of R* Dev work stream as a gap identified in open source projects.

Finally, Adding EPA parameters into the VNFD (or what is known as HPA in ONAP) is:
1. against our **principles** of abstraction.
2. out of scope of CNTT since we are not covering MANO aspects of CNTT.

rabi-abdel:
> > @rabi-abdel wrote:
> > _Certainly EPA must not used to map workload requirements to specific compute node capabilities (this all should be bundled within the profile itself) ._
> 
> Intel wrote*:
> "_For workloads requiring particular CPU and/or I/O capabilities, EPA helps Open- Stack assign virtual machines (VMs) to run on the optimal platforms._"
> 
> They sound pretty close to me. Can you help me clarify the difference?
> 
> *From: EPA white paper cover page, https://networkbuilders.intel.com/docs/ice-house-openstack-enhanced-platform-awareness.pdf
> 
> P.S. I also caught this nuance, from the EPA implementation description: "_Ensure the Nova filter scheduler uses the ComputeCapabilitiesFilter and the pci_passthrough_filter._"

@markshostak 
Yes, EPA has two parts:
- Discovery part (can be useful if we decide to allow VIM to dynamically detect profile type and extensions).
- Application part: currency used to map VNF parameters to individual host parameters (again CNTT principles).

I was referring to focus on the discovery part in the context of dynamically discovering profile types and extensions and not the application parts. 

I think it will be hard to differentiate them as there are a lot of definitions out there mixing the two concepts so as I proposed above, I think we can defer this to when we have to use it (@ASawwaf are you ok with this?)

hdamker:
a) There is nothing to review in this PR for now (just two new empty new lines)

b) Looking on the conversation so far there is good chance that this discussion get an "unresolved" label or even get rejected. Therefore I reopened #592 so that it is visible that this issue is not yet done.

c) I searched through the repository and found that there are actual too prominent uses of "EPA" within the documents:

[5.3 NFVI HW profile description](https://github.com/cntt-n/CNTT/blob/83bb391c71e1ae0f1e7432f71390233419170800/doc/ref_model/chapters/chapter05.md#53-nfvi-hw-profile-description): Relatively uncritical as it is only saying "can be considered as".

[RA1 Openstack Requirements 2.3.3 VIM](https://github.com/cntt-n/CNTT/blob/a071465f65a9a3d384e5bc98285a2d6fe83696cc/doc/ref_arch/openstack/chapters/chapter02.md#233-vim): Here is requirement
> req.vim.04 | General | The Architecture **must** support Enhanced Platform Awareness (EPA).
-- | -- | --

From my perspective this requirement is not correct or at least ambiguous (if only internal discovery is meant). @rabiabdel, @pgoyal01 do you want to open an issue against RA1 Chap 2 regarding this EPA requirement?

d) Either we resolve the two occurences of EPA in the documents or we add a definition of EPA (including the perspective of CNTT on EPA) into the Glossary

rabi-abdel:
> a) There is nothing to review in this PR for now (just two new empty new lines)
> 
> b) Looking on the conversation so far there is good chance that this discussion get an "unresolved" label or even get rejected. Therefore I reopened #592 so that it is visible that this issue is not yet done.
> 
> c) I searched through the repository and found that there are actual too prominent uses of "EPA" within the documents:
> 
> [5.3 NFVI HW profile description](https://github.com/cntt-n/CNTT/blob/83bb391c71e1ae0f1e7432f71390233419170800/doc/ref_model/chapters/chapter05.md#53-nfvi-hw-profile-description): Relatively uncritical as it is only saying "can be considered as".
> 
> [RA1 Openstack Requirements 2.3.3 VIM](https://github.com/cntt-n/CNTT/blob/a071465f65a9a3d384e5bc98285a2d6fe83696cc/doc/ref_arch/openstack/chapters/chapter02.md#233-vim): Here is requirement
> 
> > req.vim.04 | General | The Architecture **must** support Enhanced Platform Awareness (EPA).
> > -- | -- | --
> 
> From my perspective this requirement is not correct or at least ambiguous (if only internal discovery is meant). @rabiabdel, @pgoyal01 do you want to open an issue against RA1 Chap 2 regarding this EPA requirement?
> 
> d) Either we resolve the two occurences of EPA in the documents or we add a definition of EPA (including the perspective of CNTT on EPA) into the Glossary

@hdamker really good observation :) There has been few areas of which similar discussions has occurred when we needed to understand CNTT positioning. There was the issue with SR-IOV #652 , the issue with Hardware specific features #590, and now the issue with EPA #592 . My suggestion is once and for all to create section in Chapter 1 to document our position regarding those three issues and any new issues that might appear to remove any confusions when people use those terms throughout the documents. I am happy to have a stab at this to start this PR (we probably can use this one since this is currently empty). @hdamker @pgoyal01 @ASawwaf @markshostak

markshostak:
> 
> 
> My suggestion is once and for all to create section in Chapter 1 to document our position regarding those three issues <snipped....>  I am happy to have a stab at this to start this PR

@rabi-abdel Hi Rabi, Are you referring to our position on exposing h/w-specific details outside of the Infra, or position on something else? Can you give a line or two summarizing what you're envisioning? It's fine with me if you want to take a first stab. 

I'm assuming this doesn't overlap with the text I'm writing about our position on features that require h/w-specific content (i.e. drivers, etc.) within the workload that I'm working on. I see them as similar, but not the same. 

> (we probably can use this one since this is currently empty)

I'd suggest we give it its own PR because we'd have to change the name of this PR to reflect its new purpose, and if we do that, we lose a handle to refer to the already significant content in this PR discussion on the topic of EPA.

Thanks,
-Mark

rabi-abdel:
> > My suggestion is once and for all to create section in Chapter 1 to document our position regarding those three issues <snipped....>  I am happy to have a stab at this to start this PR
> 
> @rabi-abdel Hi Rabi, Are you referring to our position on exposing h/w-specific details outside of the Infra, or position on something else? Can you give a line or two summarizing what you're envisioning? It's fine with me if you want to take a first stab.

**[Rabi] something like:
CNTT positioning towards the following technologies:
- EPA: CNTT as a concept relies on profiles which are a set of fixed configurations that will exist in a given NFVI of that profile. EPA as a technology can be used to discover the profile type of a given profile but will not be used as a way to expose specific configurations to the VNFs or map specific requirements from VNFs to specific hardware configuration.
etc .
etc.**

> 
> I'm assuming this doesn't overlap with the text I'm writing about our position on features that require h/w-specific content (i.e. drivers, etc.) within the workload that I'm working on. I see them as similar, but not the same.
> 

**[Rabi] They are similar and probably can exist in the same location. so it might be a better idea to wait until that is done and then expand it afterwords in a separate PR** 

> > (we probably can use this one since this is currently empty)
> 
> I'd suggest we give it its own PR because we'd have to change the name of this PR to reflect its new purpose, and if we do that, we lose a handle to refer to the already significant content in this PR discussion on the topic of EPA.

**[Rabi] probably wait until your PR about hw specific is done as they are quite similar and related.** 

> 
> Thanks,
> -Mark

rabi-abdel:
will it be OK to close this PR based on the discussion above? defer those discussion to the currently ongoing PR https://github.com/cntt-n/CNTT/pull/731 ?

rabi-abdel:
based on discussion, we will close it and move discussion to #731

PR #703, opened 2019-11-26
RC1 Ch04 - Revised Draft

mf4716:
PR generated for leads to expand and:
- Define RM/RA-1 Openstack requirements
- Map Framework/TestCases to Requirements

Leads will update/add content:  Rajesh, Dan, Ashok

### DIFF
--- a/doc/ref_cert/lfn/chapters/chapter04.md
+++ b/doc/ref_cert/lfn/chapters/chapter04.md
@@ -36,8 +36,12 @@ Note that Each requirement may have one or more test cases associated with it.
 <a name="4.3"></a>
 ## 4.3 Traceability Matrix
 
+- Write content that explains this section defines the mapping, or traceability of RM/RA-1 requirements to test cases.
+
 <a name="4.3.1"></a>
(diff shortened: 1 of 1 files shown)


deebhatia:
@mf4716 Is this intended to trace RM/RA-1 Requirement to a test scenario and §9.2  in #708 to trace the scenario captured here to framework's (e.g. functest, yardstick etc) test case? If yes, the title can be changed to both chapters:
**CH04** - NFVI Test Scenario Traceability to RA Requirements
**CH09** - NFVI Test Case Traceability to Test Scenarios

We can define what a test scenario is, in CH04 itself in case it is confusing.

mf4716:
Deepanshu @deebhatia , yes, you're correct, and that is fine.  The goal with Ch 4 is to Describe, or Identify, Test Cases/Scenarios which map to RA Requirements.  Then, in Ch 9, we map the individual Test Cases (and Suites) to the matrix from Ch 4.  So, yes, please proceed to adjust the titles and verbiage, provided we included an intro or synopsis which describes our intentions with the chapter.

PR #706, opened 2019-11-26
RC1 Ch07 - Initial ToC and Draft

mf4716:
Started initial ToC draft. Leads will take it forward:  Rajesh, Kanagaraj, Yan Yang (confirmed)

	Define RM/RA-1 Openstack requirements
	Map Framework to Requirements


rabiabdel:
will be labeled as "idle" and "force merged" during TSC call if no more approval received b4 then.

PR #721, opened 2019-11-28
[RC-1 Ch05] VNF Platform Requirements

mkr1481:
Could you please review this initial content of ch 05, CNTT RC VNF Platform Requirements.
Thanks.


doc/ref_cert/lfn/chapters/chapter05.md
```diff
@@ -1,21 +1,184 @@
...
+
+NOTE: For certification, only compliance and verification test cases will be
+considered, but in future, it could be extent to validation and Performance
+related testing.
+
+##### Compliance
+
+Perform compliance check based on
+
+-   TOSCA using ETSI SOL004 & SOL001
```
rabi-abdel: TOSCA and HEAT are out of scope of CNTT. it might worth having a quick (including @mf4716 to discuss this)

mf4716: Expectation is to perform Heat based VVP/VTP validations for VNF testing.  This confirms ICE compliance before the VNF (golden, actual, protoype) can be instantiated and removed.  First step is compliance.  Do you @rabi-abdel or @mkr1481 see this differently?

yangyanyj: we shouldn't limit the VNF template, both Tosca and Heat should be considered, because these VNFs can all run on  CNTT compliant NFVI, as a common NFVI, it should carry VNF defined by various modules, otherwise it is not a general and common NFVI

mf4716: agree @yangyanyj - @rabi-abdel , can we resolve conversation?

rabi-abdel: I think the whole compliance part needs to be removed. CNTT will scope will not include compliance to TOSCA or HEAT (this is MANO issue not CNTT). my suggestion is to remove compliance section, approve and merge, then we can discuss it further.

mf4716: @rabi-abdel , the goals were to perform VVP as part of HEAT validation for onboarding / instantiating VNFs.  This would allow for validation that the VNF meets specs outlined by the NFVI.  Agree, out of scope are VNF functional testing, and TOSCA validation, but HEAT via VVP execution, which can be run separately from ONAP, is in scope.  Please contact me if live conversation is better.

PR #726, opened 2019-11-29
[RM Ch05] Modification of CPU pinnnig definition

### LINKED ISSUE #521: [RM Ch05] Clarification of CPU pinning definition
Section 5.1.1 Virtual compute, 
nfvi.com.cfg.003, CPU pinning capability
The definition "Binding of a process to a dedicated CPU" doesn't take into account cores and SMT.
It can be changed by:
"Binding a vCPU to either a physical core or a thread of a core with SMT"

karinesevilla:
Fixes #521 
The initial definition "Binding of a process to a dedicated CPU" doesn't take into account cores and SMT.
It is changed by:
"Binding a vCPU to either a physical core or a thread of a core with SMT"

### DIFF
--- a/doc/ref_model/chapters/chapter05.md
+++ b/doc/ref_model/chapters/chapter05.md
@@ -51,7 +51,7 @@ The following sections detail the NFVI SW profile features per type of virtual r
 |------------------|----------------|----------------|------------------------------------------------------------------------------------------------|
 | nfvi.com.cfg.001 | CPU allocation ratio  | Value | Number of virtual cores per physical core  |
 | nfvi.com.cfg.002 | NUMA awareness | Yes/No  | Support of NUMA at the virtualization layer  |
-| nfvi.com.cfg.003 | CPU pinning capability  | Yes/No | Binding of a process to a dedicated CPU |
+| nfvi.com.cfg.003 | CPU pinning capability  | Yes/No | Binding a vCPU to either a physical core or a thread of a core with SMT |
 | nfvi.com.cfg.004 | Huge Pages  | Yes/No | Ability to manage huge pages of memory |
 
 <p align="center"><b>Table 5-1:</b> Virtual Compute features.</p>
--- a/doc/ref_model/chapters/chapter05.md
+++ b/doc/ref_model/chapters/chapter05.md
@@ -51,7 +51,7 @@ The following sections detail the NFVI SW profile features per type of virtual r
 |------------------|----------------|----------------|------------------------------------------------------------------------------------------------|
 | nfvi.com.cfg.001 | CPU allocation ratio  | Value | Number of virtual cores per physical core  |
 | nfvi.com.cfg.002 | NUMA awareness | Yes/No  | Support of NUMA at the virtualization layer  |
-| nfvi.com.cfg.003 | CPU pinning capability  | Yes/No | Binding a vCPU to either a physical core or a thread of a core with SMT |
+| nfvi.com.cfg.003 | CPU pinning capability  | Yes/No | Binding a vCPU to either a physical core on a non-SMT architecture or a core thread on a SMT architecture  |
 | nfvi.com.cfg.004 | Huge Pages  | Yes/No | Ability to manage huge pages of memory |
 
 <p align="center"><b>Table 5-1:</b> Virtual Compute features.</p>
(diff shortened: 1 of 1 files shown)


rabi-abdel:
@markshostak please review

rabiabdel:
This PR will be labeled as "**idle**" and it will be marked as "**force merged**" during TSC call if no more reviews/approvals received by then. It will then be merged automatically when the 15 days period passed.

doc/ref_model/chapters/chapter05.md
```diff
@@ -51,7 +51,7 @@ The following sections detail the NFVI SW profile features per type of virtual r
 |------------------|----------------|----------------|------------------------------------------------------------------------------------------------|
 | nfvi.com.cfg.001 | CPU allocation ratio  | Value | Number of virtual cores per physical core  |
 | nfvi.com.cfg.002 | NUMA awareness | Yes/No  | Support of NUMA at the virtualization layer  |
-| nfvi.com.cfg.003 | CPU pinning capability  | Yes/No | Binding of a process to a dedicated CPU |
+| nfvi.com.cfg.003 | CPU pinning capability  | Yes/No | Binding a vCPU to either a physical core on a non-SMT architecture or a core thread on a SMT architecture  |
```
markshostak: @karinesevilla @deebhatia Hi, The Issue reported says that the line item description doesn't take into account cores and SMT. However, the change additionally replaces "process" with "vCPU". While technically correct, IMHO that change makes it harder to understand because what the reader cares about is binding the process to the h/w. The fact that the process is bound to a vCPU which is in-turn bound to the h/w adds a level of complexity that doesn't enhance the understanding, and may actually confuse some people. **Can we revert vCPU back to process?**

Also, **can we shorten the physical description** to something more concise, such as "_Binds a process to a physical core or thread_"? My thinking is, if we have to teach the difference between cores, threads, SMP and SMT, the reader probably needs to brush up on more than we can explain in one sentence. Thanks, -Mark

rabiabdel:
@karinesevilla can you please address @markshostak comments above?

pgoyal01:
Let us use a standard definition, say, from [wikipedia](https://en.wikipedia.org%2Fwiki%2FProcessor_affinity&usg=AOvVaw3WCC453UZisketi97evqK6): Processor affinity, or CPU pinning or "cache affinity", enables the binding and unbinding of a process or a thread to a central processing unit (CPU) or a range of CPUs, so that the process or thread will execute only on the designated CPU or CPUs rather than any CPU.

markshostak:
> 
> 
> Let us use a standard definition, say, from [wikipedia](https://en.wikipedia.org%2Fwiki%2FProcessor_affinity&usg=AOvVaw3WCC453UZisketi97evqK6): Processor affinity, or CPU pinning or "cache affinity", enables the binding and unbinding of a process or a thread to a central processing unit (CPU) or a range of CPUs, so that the process or thread will execute only on the designated CPU or CPUs rather than any CPU.

Let's stay in context: It's not appropriate to define a term in a configuration line item. Also, Wikipedia? :-)

doc/ref_model/chapters/chapter05.md
karinesevilla: @markshostak Hi Mark, the intention was to precise the definition and for this reason I am not aligned with the simplication you are proposing. 
vCPU is used in chapter 3 & 4 and more accurate than process is this case. Adding the SMT architecture in the defintion aims to explain why the binding is with physical core or  thread. As SMT wasn't used before , a note can be added to develop the acronym if needed.

markshostak: @karinesevilla Bonjour Karine, I understand. 
To your point, we probably want to add a definition to the glossary, rather than define terms in a table of capabilities. The vCPU definition should explain about physical core and thread in SMT. How about "_Binds a process/vCPU to a physical core or SMT thread_"?

karinesevilla: @markshostak, @deebhatia 
I agree on Mark's compromise

deebhatia: Yeah. Make sense.

rabi-abdel: has this been resolved?

rabiabdel:
once approved by @markshostak will merge.

markshostak:
> 
> 
> once approved by @markshostak will merge.

@rabiabdel @karinesevilla 
>
>
> Rabi wrote: "has this been resolved @karinesevilla? I didn't see a commit resolving it."

I don't see the referenced update yet.

doc/ref_model/chapters/chapter05.md
deebhatia: @markshostak @rabi-abdel Done the description change on behalf of @karinesevilla .

rabiabdel:
@markshostak can you please confirm if https://github.com/cntt-n/CNTT/pull/726/commits/cdc0fa28430b0c558d8afbad0a16574928a07c6d resolved your comment?

markshostak:
Opened Issue #863 to add SMT definition. @deebhatia Would you like to run with this one, please? Thanks, -Mark

PR #745, opened 2019-12-04
[RA2] Adding RA2 relation to any underlying infra.

### LINKED ISSUE #600: [RA1 RA2] Describe relationship of RA2 with underlying infrastructures
This is a follow up from the RA2 Core call on 2019-11-14 on issue #565 . The request was to extend the proposed text with examples which kind of functionality Kubernetes may expect from the underlying infrastructure and by what means Kubernetes can integrate directly with the infrastructure it is running on.

Examples for resources or capabilities of the underlying infrastructure consumed by Kubernetes are, but are not limited to:

- storage (ephemeral/persistent)
- connectivity
- compute

Kubernetes uses a plugin-based mechanism for integration with any type of infrastructure resource providers. Most prominently [Kubernetes Cloud Provider](https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/) collect various integrations towards a particular infrastructure platform e.g. to create volumes or additional worker nodes. The [Kubernetes cluster api](https://github.com/kubernetes-sigs/cluster-api) harmonizes some of the k8s cluster management related integrations with different infrastructure platforms under a common API. With the Container Network Interface (CNI) and the Container Storage Interface (CSI) specifications Kubernetes also provides additional means to integrate with the underlying infrastructure. (Note: the use of CNIs or CSIs does not imply the use of specific platform resources.)

With regards to the relation between RA2 and any other infrastructure (including RA1) we should state as part of RA2 that the implementation of RA2 may require resources from the underlying infrastructure. RA2 can integrate directly with the infrastructure using the above mentioned mechanisms.

We could then also try to simplify the existing figure in the [RA readme](https://github.com/cntt-n/CNTT/blob/master/doc/ref_arch/README.md) to reflect only the interaction mechanisms between RA2 and any underlying infrastructure including RA1.

peterwoerndle:
Fixes #600


doc/ref_arch/README.md
```diff
@@ -47,6 +47,22 @@ RA1 is focussing on an OpenStack Reference Architecture that will support VM-bas
...
 
-RA2 is delivering a Kubernetes Reference Architecture that will be standalone, meaning:
-- If there are any components from RA1 (or any other subsequent RA) that are needed in RA2, they will be explicitly included rather than requiring a Reference Implementation to refer to multiple Reference Architectures
-- The Kubernetes Reference Architecture will also include the Kubernetes cloud providers on which a Reference Implementation can choose to deploy an RA2-conformant Kubernetes
+RA2 is delivering a Kubernetes Reference Architecture that will be standalone, meaning that a Reference Implementation for RA2 can be established solely on the content and requirements present in the RA2 documentation.
+
+Kubernetes can be executed on top of differnt types of underlying infrastrucutre. Examples of those underlying infrastructures are:
```
CsatariGergely: nit: different

peterwoerndle: Fixed in 5fbb5aae1ae52f9d5edc4578b753f03322804632

doc/ref_arch/README.md
```diff
@@ -47,6 +47,22 @@ RA1 is focussing on an OpenStack Reference Architecture that will support VM-bas
...
+- storage (ephemeral/persistent)
+- connectivity
+- compute
+
+RA2 contains concrete provisions on the type of infrastructure resources consumed by Kubernetes.
+
+If the underlying infrastructure provides a dynamic way to allocate resources, Kubernetes can integrate directly with the underlying infrastructure and use it as a infrastructure resource provider. Kubernetes uses a plugin-based mechanism for integration with any type of infrastructure resource providers. Most prominently Kubernetes Cloud Provider collect various integrations towards a particular infrastructure platform e.g. to create volumes or additional worker nodes. The Kubernetes cluster api harmonizes some of the k8s cluster management related integrations with different infrastructure platforms under a common API. With the Container Network Interface (CNI) and the Container Storage Interface (CSI) specifications Kubernetes also provides additional means to integrate with the underlying infrastructure. (Note: the use of CNIs or CSIs does not imply the use of specific platform resources.)
+
```
TamasZsiros: What seems to be assumed here is that the OS is inherently part of the RA2 box (since the box beneath is virtual or real "hardware"). I think we should be explicit in that inclusion (at least point it out in the text). The reason why I bring this up is because I believe we need to make it explicit that the choice of OS (in RIx of RA2) needs to be independent from the choice of OS used to generate the docker images.

peterwoerndle: Yes, the OS (for the Kubernetes distribution) should be part of an RI of RA2. The relation to the OS used to create the docker images for the workloads (I assume that's what you mean) is not part of the relation to the underlying infra. Thus I wonder if this should rather be described in RA2 rather than in the readme for both RAs.
If there is a need to have a more detailed picture of RA2, underlying infra and the workload for RA2 I'm willing to create a PR for that as well.

TamasZsiros: Ok, let me check how/where this would be logical to be included in RA2. Maybe it's there already....

peterwoerndle:
@CsatariGergely can you please check if the change your requested has been properly addressed in 5fbb5aa.

rabi-abdel:
@TamasZsiros @CsatariGergely can you please double check that your request has been reflected correctly and resolve if was.

rabiabdel:
Force Merged as there were no objection from @CsatariGergely for more than 48 after his request was resolved.

Issue #750, opened 2019-12-04
[RM Ch-0xA] Create VNF Evolution Framework


markshostak:
This issue creates a framework to support management of the "evolution" of the CNTT VNF operating environment. As the Infra evolves, its supporting technology will change; CNTT will want to promote adoption of newer technology and potentially deprecate or even expedite deloading of older technology.

The Evolution Framework can potentially be envisioned as a roadmap. The VNF phases, let's call them "A", "B", "C", etc., as they elapse, will synchronously promote and/or deprecate various features/functions/capabilities/etc. These features may be unrelated, except that the timing of their application will be governed by VNF Evolution roadmap. 

The elapsing Evolution phases can then be used to "clock" these features into and out of vogue, within the CNTT context, and ostensibly the CNTT-RC Certification requirements, as well. 

This Issue is an action resulting from the review of, and also supports PR #749 .

rabi-abdel:
@markshostak I am happy to have a stab at this if you assign to me.

ASawwaf:
@markshostak I can support in this as well

markshostak:
@rabi-abdel @ASawwaf @mf4716 After some time to think on it, I'd propose:
- VNF Evolution phases should be identified by a number (1, 2, 3, etc.)
- VNF Evolution phases should be directly tied to CNTT releases. In other words, phases increment explicitly with each CNTT release, as opposed to incrementing based on other criteria, such as by date (i.e. quarter, month or arbitrary date)

I'm thinking the state of various features, capabilities, compromises, etc. can then be tied to the various Phases. For example, support for a given capability could be specified as optional in Phase 1, but become mandatory in Phase 3. This allows suppliers to know that a capability will be required in the future, as well as when.

As Operator upgrades are hard to mandate, I mainly see features becoming mandatory, deprecated, etc. driving the results of RC testing. For example, in the case of NEW capabilities, Capability x for NFVI is identified as optional in Phase 1, but mandatory in Phase 3:
- If an NFVI is submitted to OVP in Phase 1 without Capability x, it will PASS, but produce a "transitional" warning, indicating that Capability x will be required in Phase 3
- If an NFVI is submitted to OVP in Phase 3 without Capability x, it will FAIL

Similarly:
- If a VNF is submitted to OVP in Phase 1 that is dependent on Capability x, it will PASS, but produce a warning that it is using an optional Capability and may not be compatible with all Infra
[...]

PR #731, opened 2019-12-02
[Technical Steering] CNTT Overall Technical Direction and position

rabi-abdel:
Fixes #730

PR #754, opened 2019-12-05
[Tech Steering, RM, RA1, RA2] Transition plan

rabi-abdel:
Fixes https://github.com/cntt-n/CNTT/issues/750

This is continuation for https://github.com/cntt-n/CNTT/pull/731


doc/ref_model/chapters/appendix-a.md
```diff
@@ -1,28 +1,25 @@
...
 
 ## Table of Contents
-* [A.1 Goals](#A.1)
-* [A.2 Intro and Terminology](#A.2)
-* [A.3 VNF Maturity Levels](#A.3)
-* [A.4 Links](#A.3)
+* [A.1 Guidelines for VNF Vendors](#A.1)
+* [A.2 Guidelines for CNF Vendors](#A.2)
+* [A.3 VNF Transition Plan](#A.3)
+* [A.4 CNF Transition Phases](#A.4)
```
pgoyal01: @rabi-abdel should A.4 be consistent with A.3 -- plans versus phases

doc/ref_model/chapters/appendix-a.md
```diff
@@ -1,28 +1,25 @@
...
-1. Provide guidance to VNF or more generally Application vendors on how to consume CNTT Reference Model and Architectures
-2. Provide usable definitions of maturity levels for VNF software architecture between Physical-to-Virtual migration and “Cloud Native”.
-
-The goal is not to be prescriptive on how to re-architect existing or architect new applications but rather staying within scope of focusing on interface and interaction between applications and platform.
+## A.1 Guidelines for VNF Vendors
 
 <a name="A.2"></a>
-## A.2 Intro and Terminology
-(Summary status and trends of ETSI NFV and Cloud. Decoupling application from platform. Terminology used/introduced.)
+## A.2 Guidelines for VNF Vendors
```
pgoyal01: @rabi-abdel Guidelines for CNF Vendors

doc/ref_model/chapters/appendix-a.md
```diff
@@ -1,28 +1,25 @@
...
 
 <a name="A.3"></a>
-## A.3 VNF Maturity Levels
-(And how they map to CNTT RAs)
+## A.3 Guidelines for VNF Vendors
```
pgoyal01: @rabi-abdel A.3 VNF Transition Plan (or Phases)

doc/ref_model/chapters/appendix-a.md
```diff
@@ -1,28 +1,25 @@
...
 
 <a name="A.4"></a>
 ## A.4 Links
```
pgoyal01: @rabi-abdel A.4 CNF VNF Transition Phases (or Plans)

doc/ref_arch/kubernetes/chapters/appendix-a.md
```diff
@@ -1,21 +1,21 @@
...
-# Appendix A - Guidance for VNF-to-CNF Transition
+# Appendix A - Guidelines
 <p align="right"><img src="../figures/bogo_ifo.png" alt="scope" title="Scope" width="35%"/></p>
 
 ## Table of Contents
-* [A.1 TBC](#A.1)
-* [A.2 TBC](#A.2)
-* [A.3 TBC](#A.3)
-* [A.4 TBC](#A.3)
+* [A.1 Guidance for NFVI Vendors](#A.1)
```
tomkivlin: Depending on the scope of "NFVI" I'm not sure this is relevant here - do you mean guidance for the providers of the Kubernetes cluster(s)? Does this, as written, imply that the Kubernetes vendor and the NFVI/VIM vendor are the same?

I see the Appendix catering for:
- How cloud native are the CNFs (similar to the discussions in the TUG re. gold, silver, bronze CNFs)?
- For each level of cloud nativeness, what are the features/capabilities of Kubernetes that they need/don't need and how does that affect the compliance of a Kubernetes cluster that is provided for each level?
- For operators who are already running VNFs (and PNFs), what are the considerations and potentially options for adding CNFs into that mix - e.g. relationship with a VIM (or not), relationship with underlay SDN controllers (or not), etc. etc.

doc/ref_arch/kubernetes/chapters/appendix-a.md
```diff
@@ -1,21 +1,21 @@
...
+* [A.2 CNF Transition Plan](#A.2)
+* [A.3 Guidance for VNF-to-CNF Transition](#A.3)
+* [A.4 TBC](#A.4)
 
 <a name="A.1"></a>
-## A.1 TBC
+## A.1 Guidance for NFVI Vendors
```
tomkivlin: See above comment for line 6.

doc/ref_model/chapters/appendix-a.md
rabi-abdel: @pgoyal01 yes. fixed.

doc/ref_model/chapters/appendix-a.md
rabi-abdel: @pgoyal01 Phases. yes fixed

doc/ref_arch/kubernetes/chapters/appendix-a.md
rabi-abdel: Agree on all of the above. reworded NFVI to make it more generic and the text will explain what was meant. I will continue to add more details for the framework and diagrams to better clarify my thoughts.

doc/ref_arch/kubernetes/chapters/appendix-a.md
rabi-abdel: Yes, NFVI is removed and left to the text underneath to explain.

markshostak:
@rabi-abdel Hi Rabi, As we're also using this PR to resolve Issue #750 , please see the comments I added to that Issue last night.

Also, Nick recommended the language "Phases of Evolution", so please consider his input when contemplating terminology.

doc/ref_model/chapters/appendix-a.md
rabi-abdel: should be fixed now

doc/ref_model/chapters/appendix-a.md
rabi-abdel: renamed to Generations.

doc/ref_arch/kubernetes/chapters/appendix-a.md
rabi-abdel: @tomkivlin regarding:
Please look at latest edits and let me know what you think.

doc/ref_arch/openstack/chapters/appendix-a.md
```diff
@@ -0,0 +1,25 @@
...
+* [A.3 TBC](#A.3)
+* [A.4 TBC](#A.4)
+
+<a name="A.1"></a>
+## A.1 Guidance Vendors
+
+<a name="A.2"></a>
+## A.2 RA Transitional Generations
+
+CNF Generation definitions are defined in [RM Appendix](../../../ref_modle/chapters/appendix-a#A.3)
```
pgoyal01: "VNF Generation ..." not "CNF Generation ..."

doc/ref_arch/kubernetes/chapters/appendix-a.md
```diff
@@ -1,21 +1,25 @@
...
 
 <a name="A.1"></a>
-## A.1 TBC
+## A.1 Guidance for Vendors
 
 <a name="A.2"></a>
-## A.2 TBC
+## A.2 RA Transitional Generations
+
+CNF Generation definitions are defined in [RM Appendix](../../../ref_modle/chapters/appendix-a#A.4)
```
pgoyal01: Typo: "../../../ref_modle/chapters/appendix-a#A.4" (also in other places) -- should be "../../../ref_model/chapters/appendix-a#A.4"

doc/ref_model/chapters/appendix-a.md
```diff
@@ -1,80 +1,80 @@
...
 ## Table of Contents
-* [A.1 Goals](#A.1)
-* [A.2 Intro and Terminology](#A.2)
-* [A.3 VNF Evolution Phases](#A.3)
-* [A.4 Links](#A.3)
-* [A.5 Hardware-Dependent Coding Policies](#A.4)
-* [A.6 Miscellaneous](#A.5)
+* [A.1 Guidelines for VNF Vendors](#A.1)
+* [A.2 Guidelines for CNF Vendors](#A.2)
+* [A.3 VNF Transition Generations](#A.3)
```
pgoyal01: A.3 and A.4 should be "Transitional" to be consistent with usage elsewhere

doc/ref_model/chapters/appendix-a.md
```diff
@@ -1,80 +1,80 @@
...
+**Non-Conformant**: (Non-Conformant): None are aligned.
 
-Solving the problems associated with implementing the Abstraction Principle is a work in progress. CNTT has not solved all of the associated problems, nor has the industry. As technology evolves and more designs incorporate cloud native concepts, these problems will be addressed. This appendix is specifically intended to provide CNTT policies to manage these situations as they exist today,   and their exceptions and transitions, as the technology around and supporting the Abstraction Principle matures.
+**Gen1**: (Consistent): Going to the right direction but still lots of work to do.
```
pgoyal01: Should we classify Generations as Consistent | Conformant etc. or should we define Generations with software versions and allowed exceptions. We can also define degree of compliance with labels such as Conformant | Consistent etc. 

For example, RA-1 Generation 1 (as it stands) can be based on OpenStack Pike version and the API versions specified in the current RA-1 Chpater05; other software versions for libvirt etc. can also be specified for completeness. Additionally, RA-1 Generation 1 can specify an acceptable exception for the use of SR-IOV and qualify for "Compliant" while without SR-IOV to qualify for "Fully Compliant." 

RA-1 Generation 2 could be based on OpenStack Stein release, for example, with suitable changes to Chapter 5 and other software versions.

doc/ref_arch/openstack/chapters/appendix-a.md
rabi-abdel: should be fixed now

doc/ref_arch/kubernetes/chapters/appendix-a.md
rabi-abdel: should be fixed now

doc/ref_model/chapters/appendix-a.md
rabi-abdel: This is great point. I do suggest we get some fundamental agreement on these. I will add to next TSC call.

doc/ref_arch/kubernetes/chapters/appendix-a.md
```diff
@@ -1,21 +1,25 @@
...
-## A.2 TBC
+## A.2 RA Transitional Generations
+
+CNF Generation are defined by [Adoption Governance WS](../../../doc/gov/chapters/chapter08.md#844-cnf-transition-generations)
+
+> This will contains architectural choices that are suitable for each generation based on the Exceptions (deviation) agreed on in RM A.4
 
 <a name="A.3"></a>
-## A.3 TBC
+## A.3 Guidance for VNF-to-CNF Transition
```
CsatariGergely: This is a bit misleading as CNF-s are also VNF-s. I would rather use Guideance for cloud native architecture or something like this. 
Also, I'm not sure if it is part of the scope of CNTT to discuss VNF internal architecture.

rabiabdel:
On hold until https://github.com/cntt-n/CNTT/pull/681 is progressed and approved.

doc/ref_model/chapters/appendix-a.md
```diff
@@ -1,73 +1,50 @@
...
 
 ## Table of Contents
-* [A.1 Goals](#A.1)
-* [A.2 Intro and Terminology](#A.2)
-* [A.3 VNF Evolution Phases](#A.3)
-* [A.4 Links](#A.3)
-* [A.5 Hardware-Dependent Coding Policies](#A.4)
-* [A.6 Miscellaneous](#A.5)
+* [A.1 Guidelines for VNF Vendors](#A.1)
+* [A.2 Guidelines for CNF Vendors](#A.2)
```
ulikleber: I am not sure there should be different guidelines for VNF or CNF vendors. Also a CNF is a VNF.

doc/ref_model/chapters/appendix-a.md
```diff
@@ -1,73 +1,50 @@
...
+* [A.3 VNF Generations Details](#A.3)
+* [A.4 CNF Generations Details](#A.4)
```
ulikleber: As above. Different for VNF and CNF??

doc/ref_arch/kubernetes/chapters/appendix-a.md
tomkivlin: I don't agree that CNFs are VNFs - we surely need to distinguish between modern cloud native software (CNFs) and the current standard virtualised software (VNFs). I think it is helpful to keep them separate and show a transition from today to the future where both exist.

doc/ref_model/chapters/appendix-a.md
tomkivlin: As above, I disagree that a CNF is a VNF.  I think it is much more useful to consider them as separate epochs of network software, if you like, in the same way that VNF is a step up from PNF in terms of software architecture.

tomkivlin: The architecture of CNFs and the way they are developed, delivered and operated must be different from VNFs - that surely means they are different.

doc/ref_arch/kubernetes/chapters/appendix-a.md
ulikleber: Our definition for CNF:
Cloud native Network Function (CNF): An implementation of a Virtual Network Function (VNF, as defined by ETSI GS NFV 003) that adheres to the CNCF Cloud Native Definition or a VNF that is transitioning to cloud native.
If we drop that, we deviate completely from ETSI and have to write a new NFVO.

tomkivlin: Ah OK, I hadn't noticed that.  Worth saying that this is not the definition being shaped in the CNCF TUG any more - we've moved away from CNF as being a type of VNF and defining as a distinct type of it's own.
https://docs.google.com/document/d/1LO6G4sWHJKKW77z7gj2Aqt3IF3dRJL5q6Q1wXI4Kj8g/edit?pli=1#heading=h.mk8qm2iyy17n

tomkivlin: > If we drop that, we deviate completely from ETSI and have to write a new NFVO.

Not sure I see the logic here - why would you need a different NFVO?

ulikleber: It will be very difficult using "CNF" if the definition by CNCF TUG is changing. We really derived our definition from there. Are our references to the CNCF TUG whitepaper now outdated? 
I think it is important to say that a CNF is a VNF. NFVO uses templates for the orchestration. A VNF has a VNFD that specifies important characteristics for e.g. scalability and reliability. NS and the related templates are defined to orchestrate VNFs, their networking and connection to PNFs. We should not try to invent that again. It is true that not everything in VNFD easily maps to container technology. ETSI NFV is working on how to do it.

tomkivlin: > It will be very difficult using "CNF" if the definition by CNCF TUG is changing. We really derived our definition from there. Are our references to the CNCF TUG whitepaper now outdated?

To be clear, it has changed but TUG is looking to fix on a definition and continue to change it.

> We should not try to invent that again. 

I disagree with this and I would hope that NFV v4 from ETSI will reflect this, and if that means the NFVO needs to change to be able to orchestration network services that consist of VNFs and CNFs then that's ok, for me - again, that is something I expect the next iteration of ETSI NFV to define, rather than looking for existing models.  

I strongly feel that the whole point of cloud native is to start again from scratch / build from the ground up / etc. to take advantage of cloud principles and technologies.  This affects everything from the design through to the operating models for development, delivery and operations.  It has to be distinct from a "virtualised" application which relies on existing processes otherwise we will continue to see the "lift-and-shift" approach that has largely take place with NFV - i.e. we will end up with Pods that are requesting 32vCPU, and years between upgrades etc. instead of a true reassessment of the data model of each application and decomposition of functionality based on that to allow for the realisation of cloud native benefits.

CsatariGergely: I agree with @ulikleber that we should not redefine the whole MANO stack for cloud native VNF-s and state, that a CNF is a VNF. It is not about the internal architecture of the applications or the resource usage of the smallest manageable resource unit, but about the high level LCM of the apps.
I believe that both ALT4.6, ALT4.7 and ALT4.8 of the CNCF TUG CNF definitions also reflects this and do not try to define a different way of network level orchestration for CNFs from the way how VNF-s are orchestrated.

doc/ref_model/chapters/appendix-a.md
pgoyal01: @tomkivlin I think you would agree that VNFs (targeted for VMs) can also be developed using cloud native principles -- microservices, stateless, ... -- and processes such as "immutability" used (although expensive in time to do so). The real difference being in the deployment and orchestration of the NFs

tomkivlin: @pgoyal01 I agree CNFs can consist of VMs, sure - and bare metal components,  etc.. I disagree that VNFs can be developed using cloud native principles. A network function developed using cloud native principles is surely a CNF?

doc/ref_arch/kubernetes/chapters/appendix-a.md
tomkivlin: To be clear, I don't think MANO needs completely redefining. But o you point @CsatariGergely I do feel that the high level LCM of a VNF (non-immutability, imperative config mgmt, months/years between updates, etc.) is, and must be, different to a cloud native NF (100% immutable, declarative config mgmt, days/weeks between updates, etc.). Otherwise, what's the point, as I say we're just lifting and shifting from Physical to VM to container with none of the benefits of cloud native?

doc/ref_model/chapters/appendix-a.md
markshostak: @tomkivlin My understanding of our position is that ultimately we want to see cloud native principles baked into both VNFs and CNFs, but all else being equal, an NF on a VM is referred to as a "VNF" and an NF in a container is referred to as a "CNF", as they have different implications.

tomkivlin: @markshostak ah ok, this is where I differ, perhaps I'm alone in CNTT but I want to make sure my position is clear, even if we don't agree.  
For me, a CNF could consist of containers, VMs, physical machines, serverless functions, etc. Cloud native is more about how software is architected, developed and LCM'd than what it runs on.  Hence why RA2 scope includes ability for Kubernetes to orchestrate VMs. And I would also not expect a VNF to adopt cloud native principles - if that were the case, what's the difference between VNF and CNF?

pgoyal01: @tomkivlin key differences: declarative specifications (hw, sw, etc.) and workloads deployed in clusters orchestrated by kubernetes. In this scenario CNF is "conatinerised" NFs that are "fully cloud-native." 

VNFs do not require being "cloud native" but doesn't preclude it.

ulikleber: I would then say that a VNF that is cloud native can be deployed as part of a NFV NS using NFVO's Os-Ma-nfvo reference point.
My question is what would we do if we had a cloud native NF that is not a VNF, so it currently cannot be part of a NFV NS and thus cannot be managed by the Os-Ma-nfvo reference point?

rabi-abdel:
to minimise the scope of this PR to effectively take inputs on board, and based on the change we agreed on during TSC, I will close this PR and move the discussion to the new PR just for the appendixes (https://github.com/cntt-n/CNTT/pull/895). I will still take the input from here into there

PR #786, opened 2019-12-12
[RM Ch04] Update/Refine Storage Extensions

### LINKED ISSUE #588: [RM Ch04] Available storage extensions should remove hard-coded capacity
The [Available Storage Extensions](https://github.com/cntt-n/CNTT/blob/master/doc/ref_model/chapters/chapter04.md#4231-available-storage-extensions) table currently has a capacity field.  VNFs should not have a fixed palette of choices for size here when dealing with remote block storage.
Instead, the storage extensions table in 4.19 should be sufficient, with the possible (to be discussed) inclusion of a capability indicating the maximum possible allocation in each supported class.

**Rationale:**
When examining OpenStack APIs, one can make a 'volume' request for a particular storage back-end, and can also specify the size of the volume at the same time.  There is no OpenStack API that limits the available storage extension to a combination of a particular back-end and a particular size.  
In Kubernetes, one finds similar attributes in the PersistentVolumeClaim 'storageClassName' and the storage quantity specified by 'resources.requests.storage'.

markshostak:
Fixes #588 

Refer to Issue #588 for rationale and discussion supporting this PR.

### DIFF
--- a/doc/ref_model/chapters/chapter04.md
+++ b/doc/ref_model/chapters/chapter04.md
@@ -424,11 +424,11 @@ Persistent storage is associated with VNFs via Storage Extensions. The size of a
 
 Note, CNTT documentation uses GB and GiB to refer to a Gibibyte (2<sup>30</sup> bytes), except where explicitly stated otherwise.
 
-.conf |Read IO/s |Write IO/s Read |Throughput (MB/s) |Write Throughput (MB/s) |Max Ext Size
+.conf |Read IO/s |Write IO/s | Read Throughput (MB/s) |Write Throughput (MB/s) |Max Ext Size
 ---|---|---|---|---|---
 .bronze |Up to 3K |Up to 15K |Up to 180 |Up to 120 |16TB
 .silver |Up to 60K |Up to 30K |Up to 1200 |Up to 400 |1TB
-.gold |Up to 680K |Up to 360K |Up to 2650 |Up to 1400 |300GB
+.gold |Up to 680K |Up to 360K |Up to 2650 |Up to 1400 |1TB
 
 <p align="center"><b>Table 4-19:</b> Storage Performance Profiles</p>
 
--- a/doc/ref_model/chapters/chapter04.md
+++ b/doc/ref_model/chapters/chapter04.md
@@ -420,7 +420,9 @@ n100, n200, n300, n400, n500, n600 |100, 200, 300, 400, 500, 600 Gbps
 <a name="4.2.3"></a>
 ###  4.2.3 Storage Extensions
 
-Multiple persistent storage extensions can be attached to virtual compute instances for data storage. Each extension can be configured with the required performance category.
+Persistent storage is associated with VNFs via Storage Extensions. The size of an extension can be specified explicitly in increments of 100GB, ranging from a minimum of 100GB to a maximum of **(TBD)**. Extensions are configured with the required performance category, as per Table 4-19. Multiple persistent Storage Extensions can be attached to virtual compute instances.
+
+Note, CNTT documentation uses GB and GiB to refer to a Gibibyte (2<sup>30</sup> bytes), except where explicitly stated otherwise.
 
 .conf |Read IO/s |Write IO/s Read |Throughput (MB/s) |Write Throughput (MB/s)
 ---|---|---|---|---
@@ -432,11 +434,11 @@ Multiple persistent storage extensions can be attached to virtual compute instan
 
 Note, performance is based on a block size of 256KB or larger.
 
+<!---
 <a name="4.2.3.1"></a>
-
 #### 4.2.3.1 Available Storage Extensions
 The following table defines persistent storage extensions that can be provided to VNFs for data storage. More than one storage extension can be provided to a single VNFC. The option selected determines both the size and the performance of the extension.
-
+//
 | .conf | capacity | Read IOPS | Write IOPS | Read Throughput (MB/s) | Write Throughput (MB/s) |
 |----------|----------|------------|------------|------------------------|-------------------------|
 | .bronze1 | 100GB | Up to 3K | Up to 15K | Up to 180 | Up to 120 |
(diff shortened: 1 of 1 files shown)


markshostak:
@kedmison @tomkivlin Given our large maximum size, what do you think about setting maximum sizes based on the performance grade? For example, if you want 16TiB, you can't use NVMe. We currently have 3 categories (gold, silver, iron, etc.). We could associate a maximum extension size with each of those 3, so we can conserve the scarce high-speed storage, yet allow people to go to town on the rotating media. Thoughts?

tomkivlin:
Yes that makes sense to me

markshostak:
I've added the initial max sizes, as follows:
- Gold (300GB) // Based on my experience with the SSD storage options we discussed for UNH's lab and our previous maximum ext size
- Bronze (16TB) // Based on our consensus on the overall max extension size from Issue #588 
- Silver (1TB) // Based on Kelvin's comment in 588

All, Please review/APPROVE the PR or let me know if you think we need to make any changes to the sizes or anything else. Thanks, -Mark

kedmison:
I had been thinking of each of these (gold, silver, bronze) storage extensions as remote extensions;  the locally attached storage I had seen attached in the 'd' column of the pre-defined compute flavours. 

Thus, I was envisioning these systems providing the gold/silver/bronze storage extensions as independently scalable of the size of the compute instances, and each able to reach a high maximum volume size.   Is the lower size suggestion because we are envisioning a particular implementation constraint?

doc/ref_model/chapters/chapter04.md
```diff
@@ -420,23 +420,25 @@ n100, n200, n300, n400, n500, n600 |100, 200, 300, 400, 500, 600 Gbps
...
-.bronze |Up to 3K |Up to 15K |Up to 180 |Up to 120
-.silver |Up to 60K |Up to 30K |Up to 1200 |Up to 400
-.gold |Up to 680K |Up to 360K |Up to 2650 |Up to 1400
+Note, CNTT documentation uses GB and GiB to refer to a Gibibyte (2<sup>30</sup> bytes), except where explicitly stated otherwise.
+
+.conf |Read IO/s |Write IO/s Read |Throughput (MB/s) |Write Throughput (MB/s) |Max Ext Size
+---|---|---|---|---|---
+.bronze |Up to 3K |Up to 15K |Up to 180 |Up to 120 |16TB
+.silver |Up to 60K |Up to 30K |Up to 1200 |Up to 400 |1TB
+.gold |Up to 680K |Up to 360K |Up to 2650 |Up to 1400 |300GB
```
tomkivlin: I think it would be good to articulate the rationale for the maximum sizes.  The way I'd been thinking of it was more along the lines of:

|.conf|IOPS per GB|Throughput per GB|
|.bronze|low|low|
|.silver|medium|
|.gold|high|

My thinking being that a NF vendor would choose an instance profile, and then select a storage extension (gold/silver/bronze) and then a size - what I'm not seeing here at the moment is what drives a maximum value. i.e. how would it affect the certification of a VNF/CNF as part of CNTT/OVP?

kedmison: I think we do need to have some performance numbers specified, to create a testable baseline and to create a performance expectation that can be offered to VNF vendors.  

To that end, having IOPS and throughput specified is useful.  
Where I do not agree is in having such low limits for the maximum volume size, for the highest IOPS size.

I think the minimum volume size should be 1TB for the silver and gold tiers.  I agree with keeping the 16TB for the bronze tier.

doc/ref_model/chapters/chapter04.md
```diff
@@ -420,23 +420,25 @@ n100, n200, n300, n400, n500, n600 |100, 200, 300, 400, 500, 600 Gbps
...
+Persistent storage is associated with VNFs via Storage Extensions. The size of an extension can be specified explicitly in increments of 100GB, ranging from a minimum of 100GB to a maximum of 16TB. Extensions are configured with the required performance category, as per Table 4-19. Multiple persistent Storage Extensions can be attached to virtual compute instances.
 
-.conf |Read IO/s |Write IO/s Read |Throughput (MB/s) |Write Throughput (MB/s)
----|---|---|---|---
-.bronze |Up to 3K |Up to 15K |Up to 180 |Up to 120
-.silver |Up to 60K |Up to 30K |Up to 1200 |Up to 400
-.gold |Up to 680K |Up to 360K |Up to 2650 |Up to 1400
+Note, CNTT documentation uses GB and GiB to refer to a Gibibyte (2<sup>30</sup> bytes), except where explicitly stated otherwise.
+
+.conf |Read IO/s |Write IO/s Read |Throughput (MB/s) |Write Throughput (MB/s) |Max Ext Size
```
deebhatia: The table heading in col. 3 and 4 is mixed up. 


| .conf | Read IO/s | Write IO/s **~~Read~~** | **++Read++** Throughput (MB/s) | Write Throughput (MB/s) |Max Ext Size |
|--|--|--|--|--|--|

rabiabdel:
@markshostak would you mind please addressing @kedmison and @tomkivlin comments above. -thx

doc/ref_model/chapters/chapter04.md
markshostak: Good catch! Please mark this sub-thread as resolved. Thx

doc/ref_model/chapters/chapter04.md
markshostak: Good catch! Please mark this sub-thread as resolved. Thx

doc/ref_model/chapters/chapter04.md
markshostak: @tomkivlin Hi Tom,
>
> ...a NF vendor would choose an instance profile, and then select a storage extension (gold/silver/bronze) and then a size

I'm thinking they should be able to do that as currently written. Do you see something that would inhibit that process?

> ...what I'm not seeing here at the moment is what drives a maximum value. i.e. how would it affect the certification...

I concur, the drivers aren't listed. However, that's pretty consistent w/ how we've written most of the doc. For example, we don't provide a narrative on how the limits for RAM, number of cores, etc. were arrived at, either. Do you think this is a corner case?

tomkivlin: @markshostak let's park the narrative part - as you say this is consistent so let's take that outside this PR.

> I'm thinking they should be able to do that as currently written. Do you see something that would inhibit that process?

Only that, unless I'm mistaken, the performance of a storage extension will be driven by the size of it.  i.e. you're not going to get 2650Mbps of storage throughput from a 10G gold extension.  So, what I don't feel I'm able to do at the moment is, as someone who will be designing a workload, or deploying a workload, is to say "I need x throughput, therefore I need a gold storage extension yGB in size, even if the workload doesn't actually need that much GB capacity."

markshostak: @tomkivlin I'd say the required performance impacts the maximum extension size, but that's intentional. Specifically, the rationale is related to how much space NVMe servers have these days in practice. Also, worse case, if you need more, you can attach multiple extensions to achieve the space you need, assuming it's available and you don't get shot for hogging a precious resource.

PR #791, opened 2019-12-12
[RM Glossary] Platform definition

### LINKED ISSUE #741: [RM Glossary] Need definition of "Platform"
[RM Glossary] Need definition of "Platform" -- platform is used in many places and it is not always clear as to what is included in the platform.

pgoyal01:
[RM Glossary] Platform definition

Fixes #741

### DIFF
--- a/doc/ref_model/chapters/glossary.md
+++ b/doc/ref_model/chapters/glossary.md
@@ -98,5 +98,5 @@ To help guide the reader, this glossary provides an introduction to the terminol
 
 - **Virtualised Infrastructure Manager (VIM)**: responsible for controlling and managing the **NFVI** compute, storage and network resources.
 - **NFV Orchestrator (NFVO)**: manages the VNF lifecycle and **NFVI** resources (supported by the **VIM**) to ensure an optimised allocation of the necessary resources and connectivity.
-- **Platform**: all hardware and software that serves as a foundation or base to host VNFs and CNFs. 
+- **Platform**: A cloud capabilities type in which the cloud service user can deploy, manage and run customer-created or customer-acquired applications using one or more programming languages and one or more execution environments supported by the cloud service provider ([ITU](https://www.itu.int/rec/dologin_pub.asp?lang=e&id=T-REC-Y.3500-201408-I!!PDF-E&type=items)). 
 >_*Note:*_ For CNTT,  this includes the physical infrastructure, Operating Systems, virtualization/containerization software and other orchestration, security, monitoring/logging and life-cycle management software.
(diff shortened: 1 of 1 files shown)


doc/ref_model/chapters/glossary.md
```diff
@@ -98,3 +98,5 @@ To help guide the reader, this glossary provides an introduction to the terminol
 
 - **Virtualised Infrastructure Manager (VIM)**: responsible for controlling and managing the **NFVI** compute, storage and network resources.
 - **NFV Orchestrator (NFVO)**: manages the VNF lifecycle and **NFVI** resources (supported by the **VIM**) to ensure an optimised allocation of the necessary resources and connectivity.
+- **Platform**: all hardware and software that serves as a foundation or base to host VNFs and CNFs. 
+>_*Note:*_ For CNTT,  this includes the physical infrastructure, Operating Systems, virtualization/containerization software and other orchestration, security, monitoring/logging and life-cycle management software.
```
tomkivlin: I have no problems with this definition as written, but as with all the others I'd rather link to and/or work from an existing well-defined and industry wide definition.  e.g. the ITU defines "platform" as:

> A cloud capabilities type in which the cloud service customer can deploy, manage and run customer-created or customer-acquired applications using one or more programming languages and one or more execution environments supported by the cloud service provider.

We could use `cloud service user` instead of `cloud service customer` as we already have that definition in the Glossary but like I say, depends whether we want to create our own definitions for things or not.

pgoyal01: @tomkivlin Thanks. I am OK with the ITU definition with your suggested change. May I suggest that we include a Note so that it is clear that the plaltform includes hardware and software.

rabiabdel: @pgoyal01 can you please add the note ad agreed above in a new commit?

pgoyal01: Done

doc/ref_model/chapters/glossary.md
```diff
@@ -98,3 +98,5 @@ To help guide the reader, this glossary provides an introduction to the terminol
 
 - **Virtualised Infrastructure Manager (VIM)**: responsible for controlling and managing the **NFVI** compute, storage and network resources.
 - **NFV Orchestrator (NFVO)**: manages the VNF lifecycle and **NFVI** resources (supported by the **VIM**) to ensure an optimised allocation of the necessary resources and connectivity.
+- **Platform**: A cloud capabilities type in which the cloud service user can deploy, manage and run customer-created or customer-acquired applications using one or more programming languages and one or more execution environments supported by the cloud service provider ([ITU](https://www.itu.int/rec/dologin_pub.asp?lang=e&id=T-REC-Y.3500-201408-I!!PDF-E&type=items)).
```
CsatariGergely: This is not Platform, but Platform capabilies type (PaaS) and I'm not sure why we need the PaaS definition here. 
Also there is a differentiation between PaaS and FaaS based on the interface between the platform and the applicaiton sand the ITU-T definition is more a FaaS, than a PaaS (according to my understanding).

pgoyal01: @CsatariGergely ITU differentiates between "Platform as a Service (PaaS)" and "platform capabilities type" -- Yes, the "platform" definition is for the "platform capabilities type."

rabiabdel:
@CsatariGergely can you please confirm you are OK with the above?

PR #835, opened 2019-12-23
[RC1] Fill Heat testing

collivier:
It describes Heat testing in RC1 4.3.4.7.

Closes: #834


collivier:
@rabi-abdel @rabiabdel labeler is failing here, right?

collivier:
@mf4716 The only requirement is to cover Heat as described in RA1 Chapter 5. I don't see your technical point.

PR #850, opened 2019-12-24
RA2 ch5 Initial ToC

### LINKED ISSUE #643: [RA2 Ops] Chapter 5 (Interfaces & APIs) ToC
We need to agree a ToC for chapter 5 (Interfaces and APIs) for the Kubernetes RA.

tomkivlin:
Fixes #643

### DIFF
--- a/doc/ref_arch/kubernetes/chapters/chapter05.md
+++ b/doc/ref_arch/kubernetes/chapters/chapter05.md
@@ -5,14 +5,35 @@
 
 ## Table of Contents
 * [5.1 Introduction](#5.1)
-* [5.2 Heading](#5.2)
-* [5.3 Heading](#5.3)
+* [5.2 Management APIs](#5.2)
+* [5.3 Exposed interfaces](#5.3)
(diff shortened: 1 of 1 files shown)


tomkivlin:
> I suspect that each of the three main subsections will need to be divided into even smaller sub-subsections.
> In general, we need a diagram showing a proper layering, with the assumption that each layer can be deployed by a different open source project (or a distro vendor) and maintained independently, assuming that the inter-layer integration is done via, identified in this Chapter, open (source) APIs. I have attached my diagram, used in a bit different context, as an example only. If this is not done elsewhere, we need to add a layering similar to this in this chapter. I can adapt the attached diagram for the purpose of CNTT RA2, if this helps.
> Arch layering

@walterkozlowski we need to remain in scope of RA2 however, which is only a subset of the middle layer of your diagram.  Otherwise I think I understand what you mean, and yes a diagram would clear up how the standard interfaces are provided/consumed (e.g. CRI, CNI, OCI runtime, etc.).

PR #852, opened 2019-12-24
RA2 Ch7 ToC

### LINKED ISSUE #645: [RA2 Ops] Chapter 7 ToC
We need an agreed ToC for chapter 7 (ops and lcm) for Kubernetes RA.

tomkivlin:
Fixes #645

### DIFF
--- a/doc/ref_arch/kubernetes/chapters/chapter07.md
+++ b/doc/ref_arch/kubernetes/chapters/chapter07.md
@@ -5,14 +5,33 @@
 
 ## Table of Contents
 * [7.1 Introduction](#7.1)
-* [7.2 Heading](#7.2)
-* [7.3 Heading](#7.3)
+* [7.2 Requirements](#7.2)
+* [7.3 Kubernetes Cluster Lifecycle Management](#7.3)
+* [7.4 Monitoring Kubernetes Clusters](#7.4)
+* [7.5 Kubernetes Cluster Maintenance](#7.5)
 
 <a name="7.1"></a>
 ## 7.1 Introduction
 
+> Purpose of the chapter
+
 <a name="7.2"></a>
-## 7.2 Heading
+## 7.2 Requirements
+
+> Any requirements from chapter 2
 
 <a name="7.3"></a>
-## 7.3 Heading
+## 7.3 Kubernetes Cluster Lifecycle Management
+
+> Create, Read, Update, Delete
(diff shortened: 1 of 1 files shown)


pgoyal01:
@tomkivlin should there be a section on Operations that includes items such as configurations, secrets management, etc.

tomkivlin:
> From pure Kubernetes perspective this ToC seems adequate. However, for K8s itself these topics are very well known. We need to make sure that chapter covers E2E architecture operations and LCM, including cloud (if K8s uses VM not only bare metal), networking and storage. We may refer to RA1, where applicable, but I do not think we can avoid these topics.

@walterkozlowski I'm not sure what the value of including all of that in a CNTT document is, given the scope and desired outcomes of CNTT (certification of infrastructure and workloads).  How will including the e2e operations of not only what is in scope of RA2 but also the elements not in scope (i.e. any bare metal provider and any cloud provider) aid that goal?

tomkivlin:
> @tomkivlin should there be a section on Operations that includes items such as configurations, secrets management, etc.

I'll elaborate the LCM section, as it should be in there (LCM must be driven by desired state configuration, for example) and secrets are just another object to manage the configuration of.

tomkivlin:
> Per my previous comments, I do not see any value in discussing purely Kubernetes Ops/LCM in CNTT as this is very well documented by K8s projects. NFVI consists of: Compute, Storage and Networking (ETSI definition) and that's the E2E scope I mentioned before. As an example: tenant isolation has to be treated on K8s level (K8s namespace, well documented) together with networking aspects (e.g. slicing) and storage - how will these aspects work together?

Thanks @walterkozlowski - have been having a hard think about this.  I'm concerned that we put too much in this chapter, but how about we show and articulate where integration with other LCM aspects is needed, without necessarily describing the full e2e LCM of all potential scenarios (which is what I want to avoid - the document will become unwieldy).  Would that suffice for you?

walterkozlowski:
Hi Tom,
I am open to try this approach.
I share the same concern about overdoing this chapter, hence my concern about talking about K8s LCM, which can be found somewhere else.  I think we came from different angles, but we have the same concern actually
Regs
Walter

tomkivlin:
@walterkozlowski I've added a placeholder section 7.6.  Does this address your concern?

tomkivlin:
@walterkozlowski if you could approve this in your morning, we can look to merge, thanks.

PR #864, opened 2019-12-28
[RM Glossary] Add SMT definition

### LINKED ISSUE #863: [RM Glossary] Add Definition for SMT
This is a follow-up to PR #726 , to add the supporting definition for SMT.

deebhatia:
Fixes #863

### DIFF
--- a/doc/ref_model/chapters/glossary.md
+++ b/doc/ref_model/chapters/glossary.md
@@ -60,8 +60,9 @@ To help guide the reader, this glossary provides an introduction to the terminol
   - **Host Profile**: is another term for a **NFVI hardware profile**.
 - **NFVI Hardware Configuration**: a set of settings (Key:Value) that are applied/mapped to **NFVI** HW deployment.
 - **Simultaneous Multi-threading**: Simultaneous multi-threading (SMT), also known as Hyper-threading in Intel
-CPUs, is a technology that enables the Operating System to see a single core per CPU as two cores per CPU. These new
-cores are called *logical cores* or *threads*. This feature can usually be enabled or disabled in the Compute Node's BIOS.
+CPUs, is a hardware technology that enables the Operating System to see each physical core of a CPU as two virtual cores.
+These virtual cores are independent threads of execution. This feature can usually be enabled or disabled in the
+Compute Node's BIOS.
   >_*Note:*_ SMT **will not** make jobs run faster. It only allows two jobs to run simultaneously where only one job
 would have run before. In some cases, it can increase the amount of completed jobs within the same time span.
 
--- a/doc/ref_model/chapters/glossary.md
+++ b/doc/ref_model/chapters/glossary.md
@@ -53,11 +53,18 @@ To help guide the reader, this glossary provides an introduction to the terminol
 <a name="11.1.2"></a>
 ### 11.1.2 Hardware Layer Terminology
 
-- **Physical Network Function (PNF)**: Implementation of a network function via tightly coupled dedicated hardware and software system. >_*Note:*_ This is a physical NFVI resource with the NF software.
+- **Physical Network Function (PNF)**: Implementation of a network function via tightly coupled dedicated hardware and software system.
+  >_*Note:*_ This is a physical NFVI resource with the NF software.
 - **Hardware resources**: Compute/Storage/Network hardware resources on which the NFVI platform software, virtual machines and containers run on.
 - **NFVI Hardware Profile**: defines the behaviour, capabilities and metrics provided by an NFVI Hardware Layer.
   - **Host Profile**: is another term for a **NFVI hardware profile**.
 - **NFVI Hardware Configuration**: a set of settings (Key:Value) that are applied/mapped to **NFVI** HW deployment.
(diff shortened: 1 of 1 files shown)


doc/ref_model/chapters/glossary.md
```diff
@@ -53,11 +53,18 @@ To help guide the reader, this glossary provides an introduction to the terminol
...
 
-- **Physical Network Function (PNF)**: Implementation of a network function via tightly coupled dedicated hardware and software system. >_*Note:*_ This is a physical NFVI resource with the NF software.
+- **Physical Network Function (PNF)**: Implementation of a network function via tightly coupled dedicated hardware and software system.
+  >_*Note:*_ This is a physical NFVI resource with the NF software.
 - **Hardware resources**: Compute/Storage/Network hardware resources on which the NFVI platform software, virtual machines and containers run on.
 - **NFVI Hardware Profile**: defines the behaviour, capabilities and metrics provided by an NFVI Hardware Layer.
   - **Host Profile**: is another term for a **NFVI hardware profile**.
 - **NFVI Hardware Configuration**: a set of settings (Key:Value) that are applied/mapped to **NFVI** HW deployment.
+- **Simultaneous Multi-threading**: Simultaneous multi-threading (SMT), also known as Hyper-threading in Intel
+CPUs, is a technology that enables the Operating System to see a single core per CPU as two cores per CPU. These new
```
pgoyal01: @deebhatia The following is confusing: "... a single core per CPU as two cores per CPU ..". Did you mean "... each physical core of a CPU as two virtual cores ...". May I suggest adding "hardware" before "technology as in "... is a hardware technology that enables .." 

"These new cores are called *logical cores* or *threads* .." -- suggest change to "The virtual cores are independent threads of execution."

deebhatia: Done. Please check if it is ok.

doc/ref_model/chapters/glossary.md
```diff
@@ -53,11 +53,19 @@ To help guide the reader, this glossary provides an introduction to the terminol
...
 - **Hardware resources**: Compute/Storage/Network hardware resources on which the NFVI platform software, virtual machines and containers run on.
 - **NFVI Hardware Profile**: defines the behaviour, capabilities and metrics provided by an NFVI Hardware Layer.
   - **Host Profile**: is another term for a **NFVI hardware profile**.
 - **NFVI Hardware Configuration**: a set of settings (Key:Value) that are applied/mapped to **NFVI** HW deployment.
+- **Simultaneous Multi-threading**: Simultaneous multi-threading (SMT), also known as Hyper-threading in Intel
+CPUs, is a hardware technology that enables the Operating System to see each physical core of a CPU as two virtual cores.
+These virtual cores are independent threads of execution. This feature can usually be enabled or disabled in the
+Compute Node's BIOS.
+  >_*Note:*_ SMT **will not** make jobs run faster. It only allows two jobs to run simultaneously where only one job
+would have run before. In some cases, it can increase the amount of completed jobs within the same time span.
```
markshostak: @deebhatia A few things I'd remove:
- Ditch HT: We try not to include trademarked/proprietary terms such as hyper-threading. If you include HT, then next you'll need to include CMT (AMD's name for SMT), etc.
- Ditch BIOS: That SMT can be disabled isn't really relevant to CNTT. As far as I know, it would almost never happen, and not all machines have a BIOS. Even in machines that do have one, these days it's basically relegated to a setup program, and not even that, as setup is now under control of IPMI or RedFish.
- Ditch the Note: Those performance statements could easily be contested, and aren't required to explain what SMT is.

Regarding the core definition (bad pun intended), here are some thoughts for you, @karinesevilla and all to consider:
- I'd suggest we use the term logical core, instead of virtual core. Virtual is ambiguous w/ the virtualized resource terminology we use for workloads. For example, you can say "one physical core is acting as 10 virtual cores". If we define SMT w/ virtual cores, you could also say "one virtual core is acting as 10 virtual cores". IMHO it's more effective to say "one logical core is acting as 10 virtual cores".
- Overall, our definition is very mechanical. We might consider some form of this definition I found (see below), which is more focused on what SMT is, than how it works. I'm ok, if we just revise what we have, but people should consider the option.

"_**Simultaneous multithreading (SMT)** is a technique for improving the overall efficiency of superscalar CPUs with hardware multithreading. SMT permits multiple independent threads of execution to better utilize the resources provided by modern processor architectures._"

Cited from: https://en.wikipedia.org/wiki/Simultaneous_multithreading

Thoughts?

karinesevilla: @markshostak @deebhatia  regarding SMT definition
Agree to keep the definition from wikipedia as long as the defintion mentions multiple threads

markshostak: @deebhatia ? :-)

deebhatia: Agree. Updated.

markshostak: Thanks @deebhatia. @rabi-abdel Do we need to cite where it came from since it's taken verbatim?

markshostak:
@karinesevilla Can you approve? Merci

karinesevilla:
@deebhatia Thanks!

PR #871, opened 2020-01-01
[RC1] Fix broken links due to baseurl (/CNTT) set by GitHub

collivier:
It leverages on relative_url filter [1] as proposed by Jekyll

[1] https://jekyllrb.com/news/2016/10/06/jekyll-3-3-is-here/


collivier:
@rabi-abdel @rabiabdel @mf4716 @deebhatia It would be great to merge it quickly to fix our docs (sorry for that)

collivier:
@mf4716 No it doesn't work when executed locally with Jekyll. Then I revert to my proposal which is better regarding Jekyll baseurl guidelines (at least works)

PR #874, opened 2020-01-02
RA2 Ch4 - Host OS, Kubernetes, Container Runtime and Storage content

### LINKED ISSUE #642: [RA2 Core] Chapter 4 initial content
Now that the ToC for chapter 4 has been published, and the scope is agreed, we need to start populating this chapter.

tomkivlin:
Fixes #642


doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -51,6 +51,19 @@ These features and capabilities that are described in the software and hardware
 
 > This chapter should describe considerations about container compute services.
 
+The primary interface between the Physical / Virtual Infrastructure and any container-relevant components is the Host Operating System.  This is the OS within which the container runtime exists, and within which the containers run (and therefore, the OS whose kernel is shared by said containers).  This is shown in Figure 4-1 below.
+
+<p align="center"><img src="../figures/ch03_hostOS.png" alt="Kubernetes Host Operating System" Title="Kubernetes Host Operating System" width="65%"/></p>
+<p align="center"><b>Figure 4-1:</b> Kubernetes Host Operating System</p>
```
CsatariGergely: Should be Figure 3-1

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -17,10 +17,36 @@
...
+- A deb/rpm compatible distribution of Linux (this must be used for the master nodes, and can be used for worker nodes).
+- Windows Server 2019 (this can be used for worker nodes, but be aware of the [limitations](https://kubernetes.io/docs/setup/production-environment/windows/intro-windows-in-kubernetes/#limitations)).
+- In order to support `req.gen.cnt.03` and `req.lcm.gen.01` and `req.lcm.adp.04`, the Host OS must be disposable and therefore the configuration of the Host OS (and associated infrastructure such as VM or bare metal server) must be consistent - e.g. the system software and configuration of that software must be identical apart from those areas of configuration that must be different such as IP addresses and hostnames.
+
+Table 4-1 lists the Operating Systems that comply with this Reference Architecture specification.
+
+|OS|Version(s)|Notes|
+|---|---|---|
+|Ubuntu|16.04+||
+|Debain|9+||
```
CsatariGergely: typo: Debian

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -17,10 +17,36 @@
...
+Figure 4-1 below shows the architectural components that are described in the subsequent sections of this chapter.
+
+<p align="center"><img src="../figures/ch04_k8s_architecture.png" alt="Kubernetes Reference Architecture" Title="Kubernetes Reference Architecture" width="65%"/></p>
+<p align="center"><b>Figure 4-1:</b> Kubernetes Reference Architecture</p>
+
 <a name="4.2"></a>
 ## 4.2 Host OS
 
-> This chapter should describe the requirements for a host os to run the software stack of Reference Architecture 2.
+In order for a Host OS to be compliant with this Reference Architecture it must meet the following requirements:
```
CsatariGergely: Can we assume that all deb/rpm compatible Linux distros have modern enough Kernel to run containers?

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -30,10 +56,51 @@
 > * Which optional features are used and which optional API-s are available
 > * Which [alfa or beta features](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/) are used
 
+The distribution and version of Kubernetes used must be listed in the [Kubernetes Distributions and Platforms document](https://docs.google.com/spreadsheets/d/1LxSqBzjOxfGx3cmtZ4EbB_BGCxT_wlxW_xgHVVa23es/edit#gid=0) which lists all those that are Certified as being conformant with the CNCF Kubernetes specification.
```
CsatariGergely: - This document starts to have versions from 1.6. I believe we should have a more recent Kubernetes version required.
  - With this statement we require the CNCF Certification what excludes those who are not CNCF members and do not want to pay a lot for the certification.

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -30,10 +56,51 @@
...
+
+This Reference Architecture also specifies:
+
+- Master nodes must run the following Kubernetes control plane services:
+    - kube-apiserver
+    - kube-scheduler
+    - kube-controller-manager
+- Master nodes can also run the etcd service and host the etcd database, however etcd can also be hosted on separate nodes
+- In order to support `req.gen.rsl.01`, `req.gen.rsl.02` and `req.gen.avl.01` a Reference Implementation must:
+    - Consist of either three, five or seven nodes running the etcd service (can be colocated on the master nodes, or can run on separate nodes)
```
CsatariGergely: How about co-locating in the worker nodes?

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -30,10 +56,51 @@
...
+    - At least one master node per availability zone or fault domain to ensure the high availability and resilience of Kubernetes control plane services
+    - At least one worker node per availability zone or fault domain to ensure the high availability and resilience of workloads managed by Kubernetes
+- Master node services, including etcd, and worker node services (e.g. consumer workloads) must be kept separate - i.e. there must be at least one master node, and at least one worker node
```
CsatariGergely: So we do not support the case when the master nodes are also workers?

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: Thanks, fixed.

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: Thanks, fixed.

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: Perhaps, but I think the kernel version should be covered by Table 4-1?

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: > * This document starts to have versions from 1.6. I believe we should have a more recent Kubernetes version required.

Happy to specify a more recent version (and am expecting to) but it needs to be driven by the requirements - i.e. if there is a requirement that can only be met by a newer set of versions.  I plan to do that based on the feature gates etc. that I've already specified as they won't all be in those older versions.

tomkivlin: > * With this statement we require the CNCF Certification what excludes those who are not CNCF members and do not want to pay a lot for the certification.

Yes - they must either be a CNCF member or willing to pay a fee.  I think it's important to state this conformance even so.

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: Thanks, fixed.

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: Correct, we don't.  Keep the two separate to avoid issues where worker node workloads affect the control plane and vice versa.

doc/ref_arch/kubernetes/chapters/chapter04.md
CsatariGergely: Yes, but for me somehow Table 4-1 should be the consequence of the kernel and other requirements.

tomkivlin: Oh I see what you mean - yes fair enough.  I took the contents of Table 4-1 from the requirements of kubeadm - would calling that out make it clearer why I've put what I have in that table?

CsatariGergely: Yes

tomkivlin: Done - thanks.

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -17,10 +17,37 @@
...
+- Windows Server 2019 (this can be used for worker nodes, but be aware of the [limitations](https://kubernetes.io/docs/setup/production-environment/windows/intro-windows-in-kubernetes/#limitations)).
+- In order to support `req.gen.cnt.03` and `req.lcm.gen.01` and `req.lcm.adp.04`, the Host OS must be disposable and therefore the configuration of the Host OS (and associated infrastructure such as VM or bare metal server) must be consistent - e.g. the system software and configuration of that software must be identical apart from those areas of configuration that must be different such as IP addresses and hostnames.
+
+Table 4-1 lists the Operating Systems that comply with this Reference Architecture specification.
+
+|OS|Version(s)|Notes|
+|---|---|---|
+|Ubuntu|16.04+||
+|Debian|9+||
+|CentOS|7||
```
Levovar: CentOS / RHEL 8?

doc/ref_arch/kubernetes/chapters/chapter04.md
Levovar: the basic requirement is that nobody should ever use a non-supported version, as there will be no security fixes for it.
so in K8s case it is a must have for an infra to use at least N-2 major version, and the latest minor version

BTW features required by us are coming out continuously, for example usable NUMA allocation in 1.17, NUMA aware hugepages possibly in 1.18 / 1.19 etc.
New is always better

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -30,10 +57,51 @@
...
+  TopologyManager: true|false (ALPHA - default=false)
+```
+
 <a name="4.4"></a>
 ## 4.4 Container runtimes
 
-> This chapter should describe which container runtimes are part of the Reference Architecture.
+In order to support `req.inf.com.03`, the chosen runtime must be compliant with the [Kubernetes Container Runtime Interface (CRI)](https://kubernetes.io/blog/2016/12/container-runtime-interface-cri-in-kubernetes/) and the [Open Container Initiative (OCI) runtime spec](https://github.com/opencontainers/runtime-spec). Examples of container runtimes that are compliant with these specification are (note this is not a complete list and in no particular order):
+- container-d (with CRI plugin enabled, which it is by default)
+- rkt
```
Levovar: is rkt still a thing? also it does not support OCI format, therefore incompatible with the previous statement

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -30,10 +57,51 @@
...
+- Docker CE (via the dockershim, which is currently built in to the kubelet)
+- CRI-O
+- Frakti
+
+These specifications cover the [full lifecycle of a container](https://github.com/opencontainers/runtime-spec/blob/master/runtime.md#lifecycle) `creating > created > running > stopped` which includes the use of storage that is required during this lifecycle - this is management of the Host OS filesystem by the container runtime. This lifecycle management by the container runtime (when compliant with the above specifications) supports the requirement `req.inf.stg.06` for ephemeral storage for Pods.
+
+If privileged containers are required (not recommended, but required for some features such as the use of CSI drivers) then a container runtime that provides additional security isolation is required in order to support `req.sec.gen.01` and `req.sec.gen.04`.  This could be via the use of lightweight virtual machines (e.g. Kata) or alternative sandbox methods (e.g. gVisor).  For more information on isolation of workloads and underlying Host OS kernel, see the [Security chapter](./chapter06.md).
```
Levovar: IMO this also falls into the "hard-coding implementation choice into the architecture" category
controlling the availability of such run-time features on the K8s level is perfectly possible, and suitable via K8s PodSecurityPolicies
I would argue it is even the preferred way

doc/ref_arch/kubernetes/chapters/chapter04.md
CsatariGergely: rkt was [archived by CNCF](https://www.cncf.io/archived-projects/)

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: At present, RHEL8 isn't compatible with kubeadm.  As we progress we can add more based on other LCM tools, but I've started with just kubeadm as the "native" tool from the Kubernetes project.  i.e. I'd rather get this published as-is and add more post-Snezka.  Is that ok?

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: Will remove - thanks.  Can this be resolved?

doc/ref_arch/kubernetes/chapters/chapter04.md
Levovar: well, K8s also doesn't officially support e.g. multi-tenancy, Docker versions after 17. something etc., but still we state requirements :)
but I understand your reasoning, no strong feelings one way or the other
however if we state CentOS/RHEL 7 is okay, then we need discuss kernel versions too IMO

doc/ref_arch/kubernetes/chapters/chapter04.md
Levovar: sure, thanks

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: Good point - I have removed this line for now, as Security chapter will deal with the details.  OK to resolve?

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: Good point.  Have added the following to row60:

>In order to align with the [Kubernetes version support policy](https://kubernetes.io/docs/setup/release/version-skew-policy/#supported-versions), a Reference Implementation must use one of three latest minor versions (`n-2`) - e.g. if the latest version is 1.17 then the RI must use either 1.17, 1.16 or 1.15.

Does that resolve the change request @CsatariGergely?

tomkivlin:
> Good text overall. I have a bit of an issued with subchapter 4.2:
> 
> I understand Host OS-es specifiedare the ones compatible with kubeadm, however Peter and I will present at Prague a proposal on OS modularity. That proposal - if we can gain general consensus around it - would invalidate 4.2 as it is written today (or at least the list of Host OS-es supported would have to be modified - since the criteria of having a Linux kernel of a certain version would have to be introduced).
> 
> My suggestion is to keep this PR open and decide what to do with 4.2 after (or at) the meeting in Prague.

Mmm, good point - I am 100% supportive of that modularity concept too.  As per @Levovar 's comments as well, what I'll do is modify that and have a skeleton table, so we can merge the PR (otherwise, we miss the date for Snezka release) and create backlog item to populate - does that sound ok?

tomkivlin:
@TamasZsiros do the changes in [f4f7c71](https://github.com/cntt-n/CNTT/pull/874/commits/f4f7c715c2f0fcb5ce33161034d64efb34be372b) address your concern?

cc @Levovar as well, given previous comments.

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -51,6 +51,19 @@ These features and capabilities that are described in the software and hardware
 
 > This chapter should describe considerations about container compute services.
 
+The primary interface between the Physical / Virtual Infrastructure and any container-relevant components is the Host Operating System.  This is the OS within which the container runtime exists, and within which the containers run (and therefore, the OS whose kernel is shared by said containers).  This is shown in Figure 4-1 below.
```
pgoyal01: Typo: Figure 3-1 not 4-1

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -17,10 +17,32 @@
...
+- A version of the Linux kernel that is [compatible with kubeadm](https://kubernetes.io/docs/reference/setup-tools/kubeadm/implementation-details/#kubeadm-init-workflow-internal-design) - this has been chosen as the baseline because kubeadm is focussed on installing and managing the lifecycle of Kubernetes and nothing else, hence it is easily integrated into higher-level and more complete tooling for the full lifecycle management of the infrastructure, cluster add-ons, etc.
+- Windows Server 2019 (this can be used for worker nodes, but be aware of the [limitations](https://kubernetes.io/docs/setup/production-environment/windows/intro-windows-in-kubernetes/#limitations)).
+- In order to support `req.gen.cnt.03` and `req.lcm.gen.01` and `req.lcm.adp.04`, the Host OS must be disposable and therefore the configuration of the Host OS (and associated infrastructure such as VM or bare metal server) must be consistent - e.g. the system software and configuration of that software must be identical apart from those areas of configuration that must be different such as IP addresses and hostnames.
```
pgoyal01: How does 

req.lcm.adp.04 | Automated deployment | The Architecture must support declarative  specifications of hardware and software assets for automated deployment,  configuration, maintenance and management.
-- | -- | --

require "disposable OS"?

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -17,10 +17,32 @@
...
+
+Table 4-1 lists the Linux kernel versions that comply with this Reference Architecture specification.
+
+|OS Family|Version(s)|Notes|
+|---|---|---|
+|Linux|3.10+||
+|Linux|4+||
```
pgoyal01: OS Family | Version(s) | Notes
-- | -- | --
Linux | 3.10+ |  
Linux | 4+ |  

The second entry is already covered by the first and so should be deleted.

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: Thanks - have corrected.

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: That one is more about the consistent configuration, rather than disposability.

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -30,10 +52,48 @@
 > * Which optional features are used and which optional API-s are available
 > * Which [alfa or beta features](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/) are used
 
+The distribution and version of Kubernetes used must be listed in the [Kubernetes Distributions and Platforms document](https://docs.google.com/spreadsheets/d/1LxSqBzjOxfGx3cmtZ4EbB_BGCxT_wlxW_xgHVVa23es/edit#gid=0) which lists all those that are Certified as being conformant with the CNCF Kubernetes specification.  In order to align with the [Kubernetes version support policy](https://kubernetes.io/docs/setup/release/version-skew-policy/#supported-versions), a Reference Implementation must use one of three latest minor versions (`n-2`) - e.g. if the latest version is 1.17 then the RI must use either 1.17, 1.16 or 1.15.
```
pgoyal01: "The distribution and version of Kubernetes used must be listed in the [Kubernetes Distributions and Platforms document](https://docs.google.com/spreadsheets/d/1LxSqBzjOxfGx3cmtZ4EbB_BGCxT_wlxW_xgHVVa23es/edit#gid=0) which lists all those that are Certified as being conformant with the CNCF Kubernetes specification."

Suggest changes. Possibly:
CNTT will use one of the Kubernetes distributions listed in the [Kubernetes Distributions and Platforms document](https://docs.google.com/spreadsheets/d/1LxSqBzjOxfGx3cmtZ4EbB_BGCxT_wlxW_xgHVVa23es/edit#gid=0); the document lists distributions certified as being conformant with the CNCF Kubernetes specification.

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: Ha - good point, have removed, thanks.

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -30,10 +52,48 @@
...
+
+- Master nodes must run the following Kubernetes control plane services:
+    - kube-apiserver
+    - kube-scheduler
+    - kube-controller-manager
+- Master nodes can also run the etcd service and host the etcd database, however etcd can also be hosted on separate nodes
+- In order to support `req.gen.rsl.01`, `req.gen.rsl.02` and `req.gen.avl.01` a Reference Implementation must:
+    - Consist of either three, five or seven nodes running the etcd service (can be colocated on the master nodes, or can run on separate nodes, but not on worker nodes)
+    - At least one master node per availability zone or fault domain to ensure the high availability and resilience of Kubernetes control plane services
+    - At least one worker node per availability zone or fault domain to ensure the high availability and resilience of workloads managed by Kubernetes
```
pgoyal01: "At least one worker node per availability zone or fault domain to ensure the high availability and resilience of workloads managed by Kubernetes"

As stated it seems to imply that we only need at least one worker node in each AZ/fault domain for workload resiliency.  This needs rewording for workload resiliency: "Workload resiliency requires the need to run them on at least one worker node per availability zone or fault domain."

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: How about:

> In alignment with the Kubernetes version support policy, a Reference Implementation must use one of three latest minor versions (n-2) - e.g. if the latest version is 1.17 then the RI must use either 1.17, 1.16 or 1.15. The Kubernetes distribution or product that is used in the RI must be listed in the Kubernetes Distributions and Platforms document and marked (X) as conformant for the Kubernetes version that is being used.

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: > As stated it seems to imply that we only need at least one worker node in each AZ/fault domain for workload resiliency.

That is correct, and I'm not sure how your suggestion differs from the existing wording?  If I was to move "Consist of" from the first bullet to the statement before the list, would that make more sense?

doc/ref_arch/kubernetes/chapters/chapter04.md
pgoyal01: @tomkivlin That is great.

doc/ref_arch/kubernetes/chapters/chapter04.md
pgoyal01: @tomkivlin Maybe the sentence needs improving but I tried to state that the workload must be "run" on worker nodes in each AZ: "Workload ... need to run them on at least one worker node per availability zone" -- if not clear then does the following reword help: "Workload resiliency requires the need to run the workload on at least one worker node per availability zone or fault domain."

doc/ref_arch/kubernetes/chapters/chapter04.md
pgoyal01: @tomkivlin That was the other point how does "...the Host OS must be disposable and therefore the configuration of the Host OS (and associated infrastructure such as VM or bare metal server) must be consistent .." OS disposable require consistency as implied by "therefore."

Maybe needs 2 sentences.

TamasZsiros:
> 
> 
> > Good text overall. I have a bit of an issued with subchapter 4.2:
> > I understand Host OS-es specifiedare the ones compatible with kubeadm, however Peter and I will present at Prague a proposal on OS modularity. That proposal - if we can gain general consensus around it - would invalidate 4.2 as it is written today (or at least the list of Host OS-es supported would have to be modified - since the criteria of having a Linux kernel of a certain version would have to be introduced).
> > My suggestion is to keep this PR open and decide what to do with 4.2 after (or at) the meeting in Prague.
> 
> Mmm, good point - I am 100% supportive of that modularity concept too. As per @Levovar 's comments as well, what I'll do is modify that and have a skeleton table, so we can merge the PR (otherwise, we miss the date for Snezka release) and create backlog item to populate - does that sound ok?

-- Yes, ok!

TamasZsiros:
> 
> 
> @TamasZsiros do the changes in [f4f7c71](https://github.com/cntt-n/CNTT/pull/874/commits/f4f7c715c2f0fcb5ce33161034d64efb34be372b) address your concern?
> 
> cc @Levovar as well, given previous comments.

Yes it does, thank you.

Levovar:
> @TamasZsiros do the changes in [f4f7c71](https://github.com/cntt-n/CNTT/pull/874/commits/f4f7c715c2f0fcb5ce33161034d64efb34be372b) address your concern?
> 
> cc @Levovar as well, given previous comments.

hmm, unfortunately no :)
the reason I said "we must elaborate on kernel versions if CentOS/RHEL7 is supported" because their stock kernel is not suitable for a production CaaS.
so, if these OSs are supported, their kernel need to be changed, and IMO we should make it very explicit
the content behind the link does not state anything related to kernel, but kernel features we reference throughout the RA2 definitely do: IPVLAN, ovelayfs2 on XFS / ext4, CFS quotas etc. all require sthing sthing sthig 4.X
4.14, at the very least. 4.18 would be better a minimum. 4.19 would be the best

tomkivlin:
> the reason I said "we must elaborate on kernel versions if CentOS/RHEL7 is supported" because their stock kernel is not suitable for a production CaaS.

Can you reference some evidence for that - for example OpenShift uses RHEL7, are you suggesting it's not suitable for production use?  There will be many customers concerned I'm sure :)

> the content behind the link does not state anything related to kernel

Sorry, there is a later commit where I added the [correct link](https://kubernetes.io/docs/reference/setup-tools/kubeadm/implementation-details/#kubeadm-init-workflow-internal-design), which contains the following:

```
Kubernetes system requirements:
if running on linux:
[error] if not Kernel 3.10+ or 4+ with specific KernelSpec
```

I have no problem restricting further in future PRs, but we need to be complete about it - i.e. reference the system requirement from Kubernetes documentation that states it is required, and trace back to a requirement in chapter 2. e.g. which requirement is CFS needed for, and which Kubernetes doc states it is required?

I would suggest that we add as a backlog item(s) to modify in later PRs.

Levovar:
> Can you reference some evidence for that - for example OpenShift uses RHEL7, are you suggesting it's not suitable for production use? There will be many customers concerned I'm sure :)

hmm. kernel release notes? :)
if you can live with the restrictions imposed upon you by an older kernel, sure it is okay. I'm not here to bash or praise products or solutions, but facts are facts: some features _we_ value highly need newer kernels to even work, or just to work properly
for more info on RH I suggest asking RH representatives  how do they support e.g. IPVLAN in RHEL7 :)

Anyway some basic references related to the few examples I have used:
https://cateee.net/lkddb/web-lkddb/IPVLAN.html
https://github.com/kubernetes/kubernetes/issues/67577
https://docs.docker.com/storage/storagedriver/overlayfs-driver/ (this one was at least backported to RHEL stock kernel it seems, but nevertheless there were many bug corrections later AFAIK)
https://lwn.net/Articles/671627/


> I have no problem restricting further in future PRs, but we need to be complete about it - i.e. reference the system requirement from Kubernetes documentation that states it is required, and trace back to a requirement in chapter 2. e.g. which requirement is CFS needed for, and which Kubernetes doc states it is required?

K8s CPU management itself is based CFS quotas. E.g. when you set a "CPU limit", you actually set a CFS quota.
Inspect above link - you might want it to work properly, without additional throttling in your K8s environment :)

> I would suggest that we add as a backlog item(s) to modify in later PRs.

I'm fine however you want to handle it, but if it is stated in one document that "you shall use this & that feature and CNI and plugin", and then it is stated in another that "an operating system flat-out not supporting some, or only supporting them with bugs is OK" -> that's a direct contradiction to me

Levovar:
BTW my two cents, generally speaking: it is okay to use the upstream K8s project requirements as a starting point, however it is important to note Ra2 does not define a K8s deployment.
It defines a TelCo CaaS based on K8s, that is K8s _plus_ all the other stuff we have decided earlier to add to the scope of a "CaaS".

Therefore requirements directly coming from the K8s project are essential, but _not_ satisfactory in themselves. Our end requirements and restrictions must account for the non-K8s projects too, therefore probably end up much stricter than the base K8s criteria

tomkivlin:
> BTW my two cents, generally speaking: it is okay to use the upstream K8s project requirements as a starting point, however it is important to note Ra2 does not define a K8s deployment.
> It defines a TelCo CaaS based on K8s, that is K8s _plus_ all the other stuff we have decided earlier to add to the scope of a "CaaS".
> 
> Therefore requirements directly coming from the K8s project are essential, but _not_ satisfactory in themselves. Our end requirements and restrictions must account for the non-K8s projects too, therefore probably end up much stricter than the base K8s criteria

Totally agree - but those extra restrictions must be driven by and traceable back to requirements in chapter 2.

tomkivlin:
> I'm fine however you want to handle it, but if it is stated in one document that "you shall use this & that feature and CNI and plugin", and then it is stated in another that "an operating system flat-out not supporting some of the features, or only supporting them with known bugs is OK" -> that's a direct contradiction to me

It's a fair point.  I have created Issue #920 to capture this for a future PR.

tomkivlin:
@TamasZsiros can you approve please (still marked as you requested changes)?

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: Have modified to the following and removed reference to `req.lcm.adp.04` as that'll be covered in chapter 7 - is this ok?

> - In order to support `req.gen.cnt.03` (immutable infrastructure), the Host OS must be disposable, meaning the configuration of the Host OS (and associated infrastructure such as VM or bare metal server) must be consistent - e.g. the system software and configuration of that software must be identical apart from those areas of configuration that must be different such as IP addresses and hostnames.
> - This approach to configuration management supports `req.lcm.gen.01` (automated deployments)

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: Ah I see - I think that is outside the scope of this RA. The requirement is to support workload resilience (which having one worker node in each AZ does), not to specify how/where the workloads are run.  Bit nit-picky I know, but I am keen to avoid too much specification that isn't about the platform.  Is that ok?

tomkivlin:
In order to merge for Snezka, I can see in the comments that Tamas is ok with the changes, it's just the approval response that needs to change.  As a result it is ready to merge @rabiabdel .

PR #888, opened 2020-01-08
[RI Ch07] Update Ch07.md

lei-cloud:
I've added some contents into Ch07, covers from introduction to intaller & install steps.

### DIFF
--- a/doc/ref_impl/cntt-ri/chapters/chapter07.md
+++ b/doc/ref_impl/cntt-ri/chapters/chapter07.md
@@ -18,14 +18,7 @@
 <a name="7.1"></a>
 ## 7.1 Introduction
 
-The purpose of this chapter is to establish an operational run-book containing sequences and steps explaining, with enough detail, how to install a Reference Implementation (RI).
-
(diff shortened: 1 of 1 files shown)


lei-cloud:
Thanks for Mike @mf4716 and Yan @yangyanyj 's suggestions, I'll try to make the corresponding revisions and add these contents into PR # 817 in time.

markbeierl:
I agree, this looks like a great start for the RI 2 cookbook as far as the K8s components go.

For the ONAP part, I don't know where it should go, as right now the RA does not specify an orchestrator.  I need to see where that belongs.

rabi-abdel:
@lei-cloud what is the status of this?>

rabi-abdel:
@lei-cloud would it be possible to create a fresh new PR for the content? the chapter names has changed (this information is now in chapter 8), and this PR is based on pretty old content. please let me know if this is ok and we can close this?

rabi-abdel:
based on the comments above, seems like the conclusion was to close it.

PR #909, opened 2020-01-09
[RC1 Ch03] Clarifying requirement for offline execution of TCs

### LINKED ISSUE #890: [RC1 Ch03] - NFVI test case requirements
Need to fix some issues in https://github.com/cntt-n/CNTT/pull/812 as identified by @collivier

### LINKED ISSUE #891: [RC1] Change "[RC1 Ch03] - NFVI test case requirements" or possibly revert it
It's written that NFVI test cases *must* support either proxied Internet connectivity or an offline mode if additional container or virtual machine images need to be downloaded at runtime.

I disagree to forbid testcases when it could be considered out of NFV (enterprise proxies?) or at least up to Operator to setup mirrors. I would de facto remove VNF testing and possibly raise side effect in RI (it leverages on external DNS).

3.9.1 Functional (API) test cases falsy duplicates chapter04 https://github.com/cntt-n/CNTT/blob/master/doc/ref_cert/lfn/chapters/chapter04.md

If Chapter 3.9 is a better place, we do copy the right content from Chapter 4 to Chapter 3.

gkunz:
Clarifying the requirement that test cases muyst support some means of
running in an internal enterprise lab environment.

Moving empty sections for listing performance, resilience and bare-metal
validation test cases to chapter 4.

fixes #890 
fixes #891


gkunz:
Thanks @collivier for pointing out that the performance section is redundant. I should have checked the existing sections more thoroughly.

PR #923, opened 2020-01-10
RA2 Ch8 - update status

tomkivlin:
IFO -> SDC


rabiabdel:
This is a very minor change, it will be "Force Merged"

Issue #615, opened 2019-11-15
[RA2 Core] Requirements: The Architecture **must** provide APIs needed to discover and manage NFVI resources


tomkivlin:
What is meant with NFVI resources here? Is there the assumption that the underlying infrastructure is RA1?

_Originally posted by @hdamker in https://github.com/cntt-n/CNTT/pull/452/files_

tomkivlin:
Fixed in #948

Issue #614, opened 2019-11-15
[RA2 Core] Requirements: The Architecture **must** provide Control API endpoints to cloud platform core services.


tomkivlin:
Open-ended. What is the set of core services this refers to?

_Originally posted by @fzdarsky in https://github.com/cntt-n/CNTT/pull/452/files_

tomkivlin:
Tweak to make the requirement about the robustness and cleanliness of the APIs being offered.

tomkivlin:
ALso remove reference to 'cloud platform'.

tomkivlin:
Closed as I think this is covered by the following requirements:


```
req.gen.cnt.04 | Cloud nativeness | The Architecture must run conformant Kubernetes as defined by the CNCF.
req.int.api.02 | API | The Architecture must leverage the Kubernetes APIs to discover and declaratively manage compute (NFVI and bare metal resources), network, and storage.
```

The first ensures the Kubernetes used has a CNCF conformant API, and the second ensures all resources are managed through that API.

tomkivlin:
Sorry, re-opened as a commented-out line needs removing from chapter02.md.

petorre:
I don't know the full context of this but it might also mean Platform Services beyond Kubernetes, which are non-differentiating for vendors and bring fragmentation to CommSPs. There was another issue on that which I cannot find now.

Issue #610, opened 2019-11-15
[RA2 Core] Requirements: The Architecture **must** include kubernetes artefacts (e.g., images, Helm charts, etc.) repository capabilities.


tomkivlin:
Unspecific and open-ended. If a Kubernetes platform supported VM workloads, would this also include qcow2 images? What if the platform had integrated management of the host OS, would it then also include RPMs? Raw disk images? ostrees? Kubernetes Operators?

_Originally posted by @fzdarsky in https://github.com/cntt-n/CNTT/pull/452/files_

xmulligan:
I think this is the same as #1108 and will be solved with #1109 
@tomkivlin can we close this issue?

tomkivlin:
@xmulligan yes, I think it can.  Will close now.

Issue #602, opened 2019-11-15
[RA2 Core] Requirements: The Architecture **must** support dynamic request and configuration of resources (compute, network, storage) through Kubernetes APIs.


tomkivlin:
Ambiguous. It suggests compute/network/storage resources themselves would have to be represented as K8s objects and thus manageable via K8s APIs instead of using whatever APIs the underlying VIM (OpenStack, VMware) provided. I assume what's meant here is more resources on a pod/network_attachment/PV level of abstraction of these resources?

_Originally posted by @fzdarsky in https://github.com/cntt-n/CNTT/pull/452/commits/adcc94b60fc68f58c354867c623f8f5447951523_

tomkivlin:
Fixed in #948

PR #943, opened 2020-01-14
[RA2 Ch02] add artefact repository and update LCM

xmulligan:


xmulligan:
Adding in the line about conformant K8s address #601 
Adding in the line about an artefact repository address #610

xmulligan:
Changing Infrastructure LCM from should to must address #615 maybe also #614 ?
Adding API to LCM address #602

xmulligan:
Also added compute, network, storage to address #602

doc/ref_arch/kubernetes/chapters/chapter02.md
```diff
@@ -230,12 +231,12 @@ Please note that "shared" is a reference to multi-tenant support and pooled stor
 | Ref # | sub-category | Description |
 |---|---|---|
 | `req.kcm.07` | General | The Architecture **must** support policy driven horizontal auto-scaling of Kubernetes cluster. |
-
+| `req.kcm.04` | General | The Architecture **must** include Kubernetes artefacts (e.g., images, Helm charts, etc.) repository capabilities. |
```
tomkivlin: Not sure about this - need to understand which RM requirements this addresses.  As a more generic requirement for Kubernetes architecture, absolutely. But in the context of an architecture that is about standardising on the infrastructure capabilities consumed by workloads, I'm less certain.

xmulligan: I actually agree with you here that I'm not sure if it should be in the RA. I included it because of: 
#610 
https://cntt-n.github.io/CNTT/doc/ref_arch/kubernetes/chapters/chapter01.html Figure 1-1 here
https://cntt-n.github.io/CNTT/doc/ref_arch/kubernetes/chapters/chapter03.html here 3.2.4
https://cntt-n.github.io/CNTT/doc/ref_arch/kubernetes/chapters/chapter04.html Figure 4-1

Maybe it would be better to adopt the language from 4.8? "The reference architecture must support the usage of a Kubernetes Application package manager using the Kubernetes API-s, like Helm v3."

tomkivlin: Yes I like that. And probably a new issue to clarify and make the documentation consistent around artefact repos.

xmulligan: ok, let me create the new issue then split this PR out to make it a little more clear and separate the different issues

Issue #601, opened 2019-11-15
[RA2 Core] Requirements: The Architecture **must** use Kubernetes APIs


tomkivlin:
Too unspecific. Use Kubernetes APIs where and for what? Must use them somewhere? Everywhere?

_Originally posted by @fzdarsky in https://github.com/cntt-n/CNTT/pull/452/commits/adcc94b60fc68f58c354867c623f8f5447951523_

PR #947, opened 2020-01-16
[RA2 Ch02] adding conformant k8s as a requirement

xmulligan:
Adding in the line about conformant K8s address #601


hdamker:
Question: will this imply that an RI has to be certified according to the CNCF rules?

tomkivlin:
> Question: will this imply that an RI has to be certified according to the CNCF rules?

Yes. In effect the CNCF conformance is an API conformance check only.  The RC2 will end up being more restrictive, I would imagine as it will add further tests.

PR #951, opened 2020-01-22
[RM Appendix A] Additional HW dependent coding policies

markbeierl:
Two additional technologies to consider: raw disk access and RDMA.

### DIFF
--- a/doc/ref_model/chapters/appendix-a.md
+++ b/doc/ref_model/chapters/appendix-a.md
@@ -43,6 +43,8 @@ Several specific technology areas have been identified by CNTT as using an ABI i
 - GPU/NPU
 - FPGA/Other Acceleration
 - CPU Instruction Sets and Extensions
+- Raw Disk Device Mapping
+- Remote Direct Memory Access
 
 The preceding list is not exhaustive; technologies will be added as required.
 
@@ -66,6 +68,18 @@ Without arguing for or against SR-IOV, CNTT provides the following anecdotes whi
 
 **CPU Instruction Sets and Extensions:**
 <content needed>
+  
+**Raw Disk Device Mapping**
+<content needed>
+Overview of technology:
+
+It is possible to map SAN LUNs or local disks to a VM, giving direct hardware access to the specified device.  Given the state of today's storage technology, perhaps CNTT should not provide an exception for this.
+
+**Remote Direct Memory Access**
+<content needed>
+Overview of technology:
+
+RDMA allows direct memory access from the memory of one computer to the memory of another computer without involving the operating system or CPU . The transfer of memory is offloaded to the RDMA-capable Host Channel Adapters (HCA) . A PVRDMA network adapter provides remote direct memory access in a virtual environment.
 
 End of policies.
 
--- a/doc/ref_model/chapters/appendix-a.md
+++ b/doc/ref_model/chapters/appendix-a.md
@@ -45,6 +45,7 @@ Several specific technology areas have been identified by CNTT as using an ABI i
 - CPU Instruction Sets and Extensions
 - Raw Disk Device Mapping
 - Remote Direct Memory Access
+- Persistent Memory
 
 The preceding list is not exhaustive; technologies will be added as required.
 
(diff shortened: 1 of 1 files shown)


markbeierl:
What about technologies such as Large Receive Offload/Receive Side Coalescing or TCP Segmentation Offload?  These typically involve configuration in both the hypervisor and VM in order to work.

doc/ref_model/chapters/appendix-a.md
```diff
@@ -66,6 +68,18 @@ Without arguing for or against SR-IOV, CNTT provides the following anecdotes whi
 
 **CPU Instruction Sets and Extensions:**
 <content needed>
+  
+**Raw Disk Device Mapping**
+<content needed>
+Overview of technology:
+
+It is possible to map SAN LUNs or local disks to a VM, giving direct hardware access to the specified device.  Given the state of today's storage technology, perhaps CNTT should not provide an exception for this.
```
tomkivlin: I thought that ideally we shouldn't be talking about VMs at RM level - needs to be agnostic to bare metal, VMs, containers or whatever?

doc/ref_model/chapters/appendix-a.md
```diff
@@ -66,6 +68,18 @@ Without arguing for or against SR-IOV, CNTT provides the following anecdotes whi
...
+
+**Remote Direct Memory Access**
+<content needed>
+Overview of technology:
+
+RDMA allows direct memory access from the memory of one computer to the memory of another computer without involving the operating system or CPU . The transfer of memory is offloaded to the RDMA-capable Host Channel Adapters (HCA) . A PVRDMA network adapter provides remote direct memory access in a virtual environment.
```
tomkivlin: If we're going to give an example of how this works in a virtual environment, we should also include the example for bare metal / containerised (or state that this is not applicable).

doc/ref_model/chapters/appendix-a.md
markbeierl: Sorry, that was a quick write up to give an overview and meant to start the conversation of should we call this out as explicitly not supported.  Perhaps the use of C/VNF instead of VM?

tomkivlin: OK, understood.  Yes, I think something more agnostic like that would be good.

doc/ref_model/chapters/appendix-a.md
markbeierl: I am not familiar enough with the current state of container networking to know if RDMA is available to it.  RDMA by itself is the bare metal technology, and what I mean to indicate here is there there needs to be a translation layer for the virtualised/containerised environment, and the software running must have a 'driver' or something to talk over the RDMA channel.

doc/ref_model/chapters/appendix-a.md
markbeierl: Do we have a preferred term yet?  ETSI uses Virtual Deployment Unit (VDU) but I think it's the use of the term Virtual that we are concerned about?

doc/ref_model/chapters/appendix-a.md
```diff
@@ -67,6 +70,35 @@ Without arguing for or against SR-IOV, CNTT provides the following anecdotes whi
...
+**Raw Disk Device Mapping**
+<content needed>
+
+Overview of technology:
+
+It is possible to map SAN LUNs or local disks to a virtual unit (VM or container), giving direct hardware access to the
+specified device.  As this is a form of PCI passthrough, it should be considered as a violation of the abstration
+principle.
+
+See: vfio-pci for qemu.
```
markshostak: @markbeierl Hi Mark, Thanks for the input. I can easily see a use for this, and one thing to know about adding any features is even though the use may seem obvious, please be sure to include a _**rationale**_ for including the feature. There are lots of features we could add, but we try only to add the ones with a defendable rationale in the CNTT context. 

WRT PCI-PT, if you read the section carefully, there is no prohibition on PCI-PT, per se. The prohibition is on hardware-specific device drivers in the workload. However, you present an interesting case b/c it may or may not require special h/w drivers, depending on the ABI presented (raw block devices can be primitive or complex), and hence may or may not violate the rule. However, the local disk case may be transgressing a different concern, that we haven't captured. Specifically, if you anchor to a local piece of h/w (i.e. storage), even if no special drivers are required, you're still impeding portability. We don't currently have a prohibition on duct taping your workload to a physical server, even if there are no other concerns. The counter argument is, in some cases using local storage may be required for latency reasons, and we _have_ discussed that purely in the context of storage options, irrespective of the cloud-native discussion. If we choose to prohibit such associations, we can potentially mitigate by favoring other features you mention, such as RDMA or NVMeoF. @rabi-abdel Thoughts? @ASawwaf Comments?

doc/ref_model/chapters/appendix-a.md
```diff
@@ -67,6 +70,35 @@ Without arguing for or against SR-IOV, CNTT provides the following anecdotes whi
...
+Ref: https://github.com/qemu/qemu/blob/master/docs/rdma.txt
+
+**Persistent Memory**
+<content needed>
+
+Persistent Memory, also known as Non-Volatile Memory (NVM), is capable of maintaining data even after a power outage and is
+typically implemented as virtual NVDIMM.  As this is often backed by a host physical file, this technology must be examined to
+determine if it violates the abstraction principle.
+
+Ref: https://docs.pmem.io/persistent-memory/getting-started-guide/creating-development-environments/virtualization/qemu
```
markshostak: @markbeierl Please provide a rationale for (v)PMem. This one is not as clear to me, b/c if you're proposing to expose the "memory" as some kind of block device or especially as a SCSI device, you still effectively have a "disk", which is obviously something that is already available, and a filesystem is one of the easiest/most expedient ways to manage a big block of bytes. This is even more especially the case, if the virtual persistent memory is, in fact, implemented via a real disk drive at the physical layer (even if it's local NVMe) .  It seems really getting our money out of this paradigm (persistent memory statically mapped into a directly addressable CPU memory region), would require a highly specialized workload (at least at this time), where the kernel loader is aware the contents of this memory region is "already loaded" (i.e. it's not a file system or disk), so it doesn't try to load said contents after every kernel restart, which more akin to an embedded application, than a cloud. Also, our general direction is to persist as little as possible (except where we don't :-). Not saying it couldn't happen, and times do change, but I'm not sure we're there yet. Looking forward to seeing the rationale and use cases . Thanks, and interesting concept! -Mark

doc/ref_model/chapters/appendix-a.md
petorre: There are SW Def Storage use cases like CDN VOD with content on local drives (sizable content, can but doesn't have to  be latency sensitive) when VM cannot simply be instantiated on another node.

doc/ref_model/chapters/appendix-a.md
petorre: Current Intel implementation (DIMM capacities of 128, 256 and 512GB) has two modes:

1. non-persistent large memory beyond DRAM,
2. Persistent where app needs to access the memory and be enabled so node after reboot it can find the data (instead of reloading TBs from somewhere).

It can be used for in-memory DBs like store state for decomposed functions, or CDN Live Streaming high-res channels.

rabiabdel:
This PR will be closed if the comments above has not been addressed for more than 48 hours.

rabiabdel:
closing according to above

PR #1002, opened 2020-01-31
[RA2 Ch3]: Adding clarification about the usage of Tiller

### LINKED ISSUE #984: [RA2] Ch3 - clarification re. Tiller
From deep dive in Prague:
3.2.4 State that a pachage manager shall only interact wit hte k8s API (no Tiller)

We should put a generic statement that the RA2 does not recommend any package managers with a server side component

CsatariGergely:
This change adds a clarificaiton about the usage of Tiller.

Fixes #984


doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -150,4 +150,4 @@ There are no restrictions or constraints that Kubernetes places on the storage t
...
+To manage complex applications consisting from several Pods the reference architecture may provide support for a Kubernetes Application package manager. The package manager may be able to manage the lifecycle a set of Pods and provide a framework to customise a set of parameters for the deployment. The requirement on this Reference Architecture is to provide the Kubernetes API in conformance with the CNCF Conformance test, for the package managers to use in the lifecycle management of the applications they manage. The Referrence Architecture does not recomment the usage of a Kubernetes Applicaiton package manager with a server side component installed to the Kubernetes cluster (e.g.: Tiller).
```
nickolaev: Helm drops Tiller in its 3.0 [release](https://helm.sh/blog/helm-3-released/) and this is the recommended version to be used today. The Helm 2 [support plan](https://helm.sh/blog/2019-10-22-helm-2150-released/) suggests that it will be discontinued by the end of 2020.

tomkivlin: Suggest a twist on the wording so rather than "The Reference Architecture does not recommend" to something like "As such, any cluster-side components such as Tiller for Helm v2 are not supported by this Reference Architecture", to avoid recommending (or not) and focusing on what is or isn't tested by the compliance tests.
Also typos:
"Referrence" > "Reference"
"Applicaiton" > "Application"
"recomment" > "recommend"

rabi-abdel:
@nickolaev can you re-review Please?

One more approval will result in merging this PR since no further input is received by @nickolaev for 48 hours as per our process.

PR #1014, opened 2020-02-02
[RA2 Ch02] Persistent storage installation should be optional.

fkautz:
There are likely going to be edge clusters which will be deployed with no storage. Requiring persistent storage is counterintuitive in a cluster when your intention is to deploy a firewall (or similar) with only ephemeral requirements.

For reference, many enterprise clusters are built without persistent storage.

### DIFF
--- a/doc/ref_arch/kubernetes/chapters/chapter02.md
+++ b/doc/ref_arch/kubernetes/chapters/chapter02.md
@@ -195,7 +195,7 @@ Reference to "Architecture" in this chapter refers to the NFVI Hardware (e.g. ph
 | `req.inf.stg.04` | Storage | The Architecture **may** support Software Defined Storage (SDS) that seamlessly supports shared block storage, object storage and flat files. |
 | `req.inf.stg.05` | Storage | The Architecture **should** provide high-performance and horizontally scalable storage. |
 | `req.inf.stg.06` | Storage | The Architecture **must** support ephemeral storage (non-persistent) storage for Pods. |
-| `req.inf.stg.07` | Storage | The Architecture **must** support persistent storage for Pods. |
+| `req.inf.stg.07` | Storage | The Architecture **should** support persistent storage for Pods. |
 | `req.inf.ntw.01` | Network | The Architecture **must** support Container Network Interface (CNI). |
 | `req.inf.ntw.02` | Network | The Architecture **must** support intra-node communications, such as between agents on a node and all pods on that node |
 | `req.inf.ntw.03` | Network | The Architecture **must** support inter-node communications without NAT, such as communications between pods on a node with all other pods on all nodes |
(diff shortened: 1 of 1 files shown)


tomkivlin:
@fkautz - I see that as being an implementation choice. For me, the Architecture must include the capability as if it doesn't (which is allowed if we use "should"), those implementations that choose to use persistent storage will not be able to test compliance as there won't be a spec for persistent storage.  However, I agree that persistent storage is not required for all implementation scenarios.

fkautz:
@tomkivlin can we add something where persistent storage capabilities don't need to be installed? Basically, The architecture must support it, but not require it to be installed or used.

tomkivlin:
> @tomkivlin can we add something where persistent storage capabilities don't need to be installed? Basically, The architecture must support it, but not require it to be installed or used.

Yes, makes sense. This may be handled in the profiles mapping but could add a note to the requirement too.

peterwoerndle:
Isn't that a more generic issue with regards to what requirements does the solutions as such support and what do we mandate to be present in a deployment. I assume this is important input to RI2 and RC2 as particularly RC2 needs to handle cases where we have a requirement but it does not have to be implemented in all cases. Should we create a cross workstream issue on that?

fkautz:
@tomkivlin can you check the wording here? I think it expresses the intention but want to make sure the wording matches.

peterwoerndle:
It's good that we keep the mandatory requirement on the architecture to support persistent storage. To express the conditionality for deployment based on the "SW/HW profile" I suggest to explicitly refer to the implementation rather than to the architecture.
Somethinge like:

> An implementation following the architecture **may** provide persistent storage for Pods based on the applied HW or SW profile.

tomkivlin:
@fkautz @peterwoerndle I think trying to word a requirement is going to be tricky, for the reasons Peter mentions, this is an Implementation requirement based on the profiles of the workloads so doesn't fit, for me, in the list of Architecture requirements.  I wonder whether we just add a note to the end of `req.inf.stg.07` along the lines of what Peter has suggested?  e.g.

> The Architecture **must** support persistent storage for Pods. (Note, whilst the Architecture must support persistent storage, an Implementation may provide persistent storage for Pods based on the applied HW or SW profile.)

fkautz:
@tomkivlin my concern is that leaving out some wording doesn't reflect the concept that persistent storage is not required to be installed. 

If an architecture must support a feature, it implies that this feature exists in the implementation and deployment.

tomkivlin:
Meeting 13/02: Ramki to add suggested wording here.  @tomkivlin to check profiles in RM.

fkautz:
There are two primary possibilities that I can see:

1. Architecture requires storage
2. **Vendor implementation is required**
3. Operator may choose to deploy or not

vs

1. Architecture requires storage
2. **Vendor implementation is optional but suggested**
3. Operator may choose to deploy if available

I recommend the first option for RA2.

gkunz:
I also believe that the first option is the better one. Thinking about RC2, the second option implies that a test for persistent storage cannot be included in the mandatory scope of the compliance test suite, thereby allowing vendor platforms to pass the compliance test suite without supporting persistent storage support.

fkautz:
> I also believe that the first option is the better one. Thinking about RC2, the second option implies that a test for persistent storage cannot be included in the mandatory scope of the compliance test suite, thereby allowing vendor platforms to pass the compliance test suite without supporting persistent storage support.

My thoughts exactly. We want to make sure there is a minimum set of required capabilities that may be optionally enabled/disabled at runtime.

tomkivlin:
I agree completely with first option.  I was wondering if we could/should modify how the requirements are presented.  I am planning to bring a proposal to the call later today.

rabiabdel:
@tomkivlin is this PR now a duplicate?

tomkivlin:
> @tomkivlin is this PR now a duplicate?

Yes and now you've merged #1130 it can be closed.

PR #1016, opened 2020-02-02
[RA-2 Ch02] req.inf.ntw.16 states a capability rather than an implementation detail

fkautz:
The former requirement enforces an implementation detail.

The new wording outlines a capability instead of an implementation detail.

If we want this capability to be programmatically configured/controlled via an API, then that should be a separate requirement. Even then, such a requirement should not specifically call for a custom CNI plugin since this is an implementation detail.

### DIFF
--- a/doc/ref_arch/kubernetes/chapters/chapter02.md
+++ b/doc/ref_arch/kubernetes/chapters/chapter02.md
@@ -208,7 +208,7 @@ Reference to "Architecture" in this chapter refers to the NFVI Hardware (e.g. ph
 | `req.inf.ntw.11` | Network | The networking solution **should** be able to be centrally administrated and configured. |
 | `req.inf.ntw.14` | Network | The Architecture **must** support capabilities for integrating SDN controllers. |
 | `req.inf.ntw.15` | Network | The Architecture **should** support Service Mesh Interface (SMI). |
-| `req.inf.ntw.16` | Network | The Architecture **should**  support co-exisence of multiple network interfaces. |
+| `req.inf.ntw.16` | Network | The Architecture **should**  support coexistence of multiple connection points in a pod. |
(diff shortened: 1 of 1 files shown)


doc/ref_arch/kubernetes/chapters/chapter02.md
```diff
@@ -208,7 +208,7 @@ Reference to "Architecture" in this chapter refers to the NFVI Hardware (e.g. ph
 | `req.inf.ntw.11` | Network | The networking solution **should** be able to be centrally administrated and configured. |
 | `req.inf.ntw.14` | Network | The Architecture **must** support capabilities for integrating SDN controllers. |
 | `req.inf.ntw.15` | Network | The Architecture **should** support Service Mesh Interface (SMI). |
-| `req.inf.ntw.16` | Network | The Architecture **should**  support co-exisence of multiple Container Network Interface (CNI). |
+| `req.inf.ntw.16` | Network | The Architecture **should**  support co-exisence of multiple network interfaces. |
```
tomkivlin: Agree with the sentiment however suggested wording isn't clear if we're talking about for Pods or Nodes. 

Suggest changing "network interfaces" to "connection points" in line with [chapter 4 from the Reference Model](../../blob/master/doc/ref_model/chapters/chapter04.md) and specifying that it's for Pods.

deebhatia: Also, fix small typo.
`co-exisence` -> `coexistence`

fkautz: I've updated both. Can you please both review the changes?

deebhatia: Thanks. Good from my side.

tomkivlin: Yes me too, thanks

PR #1024, opened 2020-02-03
Build script of the documentation

### LINKED ISSUE #1023: There is no automatic checking on the changes
At the moment we have no tools to check if the changes in the documentation are okay from syntax point of view. In ideal case we would need to run an .md linter with agreed rules and a link checker on the changes before they are merged.

CsatariGergely:
This change adds a build script to build the documentation.
At the moment the script supports only Ubuntu and it is very
rudymentary, but at least a start.

This is the 0th step to fix #1023

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,0 +1,3 @@
+source 'https://rubygems.org'
+gem "github-pages", group: :jekyll_plugins
+gem "jekyll"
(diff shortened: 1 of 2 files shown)


ulikleber:
Very much appreciate this

ulikleber:
@CsatariGergely , did you run the script against the Snezka released docs? It would be great to see the result.

rabi-abdel:
@walterkozlowski @ulikleber @bfcohen @collivier please review and approve

PR #1047, opened 2020-02-06
[TSC] remove labler

### LINKED ISSUE #1040: [tooling]: Consider dropping actions/labeler for the fork and pr model
We are using the actions/labeler GitHub action to do automatic labelling, but this prevents us from using the fork and PR model. (See more about this in the issue of [actions/labeler](https://github.com/actions/labeler/issues/12#event-3010934809))
With this current ways of working every PR have to be done from a branch in cntt-n/CNTT. The drawback of this are that to create a pr one needs a write access to the repo to be able to create the branch. This leads to not needed administration and makes onboarding difficult.

rabi-abdel:
Fixes #1040


CsatariGergely:
I hope at some point of time we will have a labeller what can work with the fork and pr model, but the current one does not support the creation of pr-s from forked repos (see more in [here](https://github.com/actions/labeler/issues/12)). I think it is more important to get pr-s from forked repos (withou the need of giving write access to the upstream repo) than the automatic labelling.

PR #1079, opened 2020-02-17
[RA2 Ch02] Values for virtual CPU, memory, and disk capabilities

### LINKED ISSUE #1068: [RA2 Ch02] requirements for virtual CPU, memory, and disk capabilities
Currently, the descriptions of requirements e.nfvi.res.cap.001, e.nfvi.res.cap.002, and e.nfvi.res.cap.003 in chapter 2 [1] state _max_ values for virtual CPU, memory and disk resources allocated to a pod. Phrasing these as maximum values is misleading as they should not define a hard upper bound.

[1] https://github.com/cntt-n/CNTT/blob/master/doc/ref_arch/kubernetes/chapters/chapter02.md#224-instance-capabilities-mapping

gkunz:
The requirements e.nfvi.res.cap.001, e.nfvi.res.cap.002, and
e.nfvi.res.cap.003 in chapter 2 state maximum values for virtual
CPU, memory and disk sizes. This PR clarifies that those numbers
are not intended as a hard upper limit, but just state that a platform
has to support *at least* those maximum values. The actual maximum
values can be larger in a vendor implementation.

fixes #1068

### DIFF
--- a/doc/ref_arch/kubernetes/chapters/chapter02.md
+++ b/doc/ref_arch/kubernetes/chapters/chapter02.md
@@ -88,9 +88,9 @@ From Reference Model section [4.2.5 Instance Capabilities Mapping](../../../ref_
 
 | Attribute | Description | Value | Supported |
 |-----------|---------------------------|-------|-------|
-| e.nfvi.res.cap.001 | Max number of vCPU that can be assigned to a single pod by the NFVI | 16<sup>1)</sup> | Y |
-| e.nfvi.res.cap.002 | Max memory in MB that can be assigned to a single pod by the NFVI | 32 GB<sup>1)</sup> | Y |
-| e.nfvi.res.cap.003 | Max storage in GB that can be assigned to a single pod by the NFVI | 320 GB<sup>1)</sup> | Y |
+| e.nfvi.res.cap.001 | Max number of vCPU that can be assigned to a single pod by the NFVI | at least 16<sup>1)</sup> | Y |
+| e.nfvi.res.cap.002 | Max memory in MB that can be assigned to a single pod by the NFVI | at least 32 GB<sup>1)</sup> | Y |
+| e.nfvi.res.cap.003 | Max storage in GB that can be assigned to a single pod by the NFVI | at least 320 GB<sup>1)</sup> | Y |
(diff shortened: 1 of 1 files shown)


tomkivlin:
In theory, I don't have a problem with this, however I think more care is needed over the wording, to avoid software vendors bringing huge pods (and therefore minimising some of the benefits of decomposing into smaller independent software processes).  For example, should we state that the maximum size of a Pod can be no larger than the number of cores in a physical CPU, to ensure Pods can be pinned to a NUMA node?  Or is that making it too complicated?

CsatariGergely:
> 
> 
> In theory, I don't have a problem with this, however I think more care is needed over the wording, to avoid software vendors bringing huge pods (and therefore minimising some of the benefits of decomposing into smaller independent software processes). For example, should we state that the maximum size of a Pod can be no larger than the number of cores in a physical CPU, to ensure Pods can be pinned to a NUMA node? Or is that making it too complicated?

I think it makes things a bit more complicated, and not all applications need all full NUMA alignement. I would let the applications decide if they need NUMA alignement or other CPU pooling features or not.

tomkivlin:
> > In theory, I don't have a problem with this, however I think more care is needed over the wording, to avoid software vendors bringing huge pods (and therefore minimising some of the benefits of decomposing into smaller independent software processes). For example, should we state that the maximum size of a Pod can be no larger than the number of cores in a physical CPU, to ensure Pods can be pinned to a NUMA node? Or is that making it too complicated?
> 
> I think it makes things a bit more complicated, and not all applications need all full NUMA alignement. I would let the applications decide if they need NUMA alignement or other CPU pooling features or not.

Good point - yes I agree.  That then raises the question of, if the "maximum" size of something is "at least 16" - does that mean there is no maximum, and is there any value in stating "at least 16" - i.e. as opposed to "at least 1"?

gkunz:
Hi @tomkivlin. I have to admit that I didn't consider your point of avoiding too large pods. Still, I would expect this to be a requirement on the application (flavor equivalent) rather then the platform. Anyway, glad you agreed and approved.

rabiabdel:
Will give it the needed 48 hours before merging.

PR #1122, opened 2020-02-20
[RA-2 Ch02] Don't interfere with default K8s network

fkautz:
When we add or remove networking capabilities to a cluster, node or pod, they must not interfere with the network contract that the cluster is providing to its parts.


tomkivlin:
I think I understand what you're after, but I wonder if we need to be more explicit.
e.g. a requirement stating that the architecture must support the standard model of "pods on a node can communicate with all pods on all nodes without NAT", 'agents on a node (e.g. system daemons, kubelet) can communicate with all pods on that node", etc.
Basically all the standard "contracts" as you've described them, as listed here: https://kubernetes.io/docs/concepts/cluster-administration/networking/

Is that what you're after?

gkunz:
I'd like to agree with @tomkivlin, but waiting for @fkautz to clarify. :-)

tomkivlin:
@fkautz could you let us know if we're on the right lines please?

fkautz:
Thinking through this a bit more, I think we should word this a bit more tightly.

There are two things that people generally want for supporting multiple users/workloads:

* Safety between workloads that are mutually ignorant of each other
* Workloads must not be able to starve others of resources.

Perhaps we should word it as:

```
| `req.inf.ntw.17` | Network | The network solution **must** must not interfere with connection points of which it is ignorant of. |
```

What are your thoughts?

tomkivlin:
@fkautz I'm struggling a little.  Can you provide an example perhaps to illustrate?  I'm not sure how the multi-tenancy points you've made relate to the networking solution not interfering with connection points it is ignorant of.  In addition to that, how _could_ a networking solution interfere with something it is ignorant of, if it is ignorant of it?

tomkivlin:
From 27/02 meeting: @fkautz to update wording including consideration of `ntw.02` and `ntw.03` (thinking about the Kubernetes model and the standard networking contract.

tomkivlin:
@fkautz have you had a chance to consider this wording?

rabiabdel:
@tomkivlin we should have discussed this PR during TSM. shall we close it since no activity in this PR?

tomkivlin:
> @tomkivlin we should have discussed this PR during TSM. shall we close it since no activity in this PR?

No, thanks. @fkautz has been unwell but is back online now.  Will pick it up with him.

tomkivlin:
Following discussion:
@fkautz to call out specific networking constructs as detailed here: https://kubernetes.io/docs/concepts/cluster-administration/networking/#the-kubernetes-network-model
@fkautz to add a specific requirement along the lines of "config of one CNI must not be owned/updated by anything other than the controller of that CNI" to ensure networking solutions not interfering with default contracts / CNI plugins.

fkautz:
I've reworded this:

1. The architecture must provide a default network which is the kubernetes network.
2. Multiple networking solutions may be installed.
3. The network solution must not with interfere with other network solutions which may be installed.

doc/ref_arch/kubernetes/chapters/chapter02.md
```diff
@@ -214,6 +214,9 @@ Reference to "Architecture" in this chapter refers to the NFVI Hardware (e.g. ph
 | `req.inf.ntw.14` | Network | The Architecture **must** support capabilities for integrating SDN controllers. |
 | `req.inf.ntw.15` | Network | The Architecture **should** support Service Mesh Interface (SMI). |
 | `req.inf.ntw.16` | Network | The Architecture **should**  support coexistence of multiple connection points in a pod. |
+| `req.inf.ntw.17` | Network | The Architecture **may**  support more than one networking solution.
```
tomkivlin: Can I suggest splitting this into two requirements as per persistent storage:

`The Architecture must be able to support more than one networking solution.`

`The Architecture must support the ability for an operator to choose whether or not to deploy more than one networking solution.`

PR #1127, opened 2020-02-20
[RA2 Ch3] remove redundant note

### LINKED ISSUE #979: [RA2] Ch3 - remove redundant note
From deep dive in Prague:
We can delete the note starting with "As this RA is developed, it is quite p..." from the end of Ch3.1 as it is was done

xmulligan:
resolves #979


ulikleber:
Hi, where is the other note that makes this one redundant?

yangxu-hw:
> Hi, where is the other note that makes this one redundant?
I think he doesn't mean this note is redundant with another one. The issue is RM has done some modification adding container related descriptions, so that the existing note is no more needed.

PR #1128, opened 2020-02-20
[RA2 Ch4] add must meet all requirements

### LINKED ISSUE #989: [RA2] Ch4 - first sentence in Chapter 4.2 should state that all the bullet point must be met
From deep dive in Prague:
first sentence in Chapter 4.2 should state that all the bullet point must be met

xmulligan:
to resolve #989

### DIFF
--- a/doc/ref_arch/kubernetes/chapters/chapter04.md
+++ b/doc/ref_arch/kubernetes/chapters/chapter04.md
@@ -27,7 +27,7 @@ Figure 4-1 below shows the architectural components that are described in the su
 <a name="4.2"></a>
 ## 4.2 Host OS
 
-In order for a Host OS to be compliant with this Reference Architecture it must meet the following requirements:
+In order for a Host OS to be compliant with this Reference Architecture, it must meet all of the following requirements:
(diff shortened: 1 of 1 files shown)


xmulligan:
to resolve #989

gkunz:
Well, I understand the reasoning behind this change, but from a pure logical perspective the new statement is incorrect because a host cannot be a Linux distribution *and* Windows 2019 Server at the same time. So, if we really want to explicitly state that *all* requirements are met, we need to somehow restructure the first 3 bullets. I personally would be ok with the original version of the text.

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -27,7 +27,7 @@ Figure 4-1 below shows the architectural components that are described in the su
 <a name="4.2"></a>
 ## 4.2 Host OS
 
-In order for a Host OS to be compliant with this Reference Architecture it must meet the following requirements:
+In order for a Host OS to be compliant with this Reference Architecture, it must meet all of the following requirements:
```
CsatariGergely: I agree with  @gkunz the new text is impossible to satisfy, the old text is better from this perspective.

PR #1130, opened 2020-02-20
[RA2 Ch2] Wording to deal with requirements that are optional at deployment time.

### LINKED ISSUE #1048: [RA1 RA2] Express conditional requirements for implementations based on SW/HW profiles
In the discussions on #1014 we came across the issue that we'd like to have a mandatory requirement on the architecture but a conditional requirement on the implementation. In the RA2 call we concluded that we need some way to communicate that information to RI2 as part of the RA2 text. I'd therefore like to have a discussion on the following "pattern" to describe these kind of requirements.

> The architecture **must** support _capability xyz_.
> An implementation following the architecture **may** provide _capability xyz_ based on the applied HW or SW profile.

We can have a leading statement that if no conditionality is given, an implementation has to follow the architectural requirements. If possible we should also refer to the profile definition handling the conditionality.

tomkivlin:
Resolves #1048

### DIFF
--- a/doc/ref_arch/kubernetes/chapters/chapter02.md
+++ b/doc/ref_arch/kubernetes/chapters/chapter02.md
@@ -28,7 +28,7 @@
 
 **should**: Requirements that are marked as _should_ are expected to be fulfilled by the Reference Architecture but it is up to each service provider to accept a Vendor Implementation targeting conformance with this Reference Architecture that is not reflecting on any of those requirements. The same applies to _should not_.
 
-- ***NOTE:*** Some of these **must** and **should** requirements may be optional when it comes to an operator **deployment** (i.e. an operator may choose whether or not to deploy or not, depending on the NFVI Software Profiles required for each specific deployment). These will be marked with "(*optional for deployment*)" within the requirements tables in section 2.3.
+- ***NOTE:*** Some of these **must** and **should** requirements may be optional when it comes to an operator **deployment** (i.e. an operator may choose whether or not to deploy or not, depending on the NFVI Software Profiles required for each specific deployment). Specific requirements asking for the Reference Architecture to be flexible enough to allow that optionality will be added in the tables in section 2.3.
 
 **may**: Requirements that are marked as _may_ are considered optional. The same applies to _may not_.
 
@@ -197,7 +197,8 @@ Reference to "Architecture" in this chapter refers to the NFVI Hardware (e.g. ph
 | `req.inf.stg.04` | Storage | The Architecture **may** support Software Defined Storage (SDS) that seamlessly supports shared block storage, object storage and flat files. |
 | `req.inf.stg.05` | Storage | The Architecture **should** provide high-performance and horizontally scalable storage. |
 | `req.inf.stg.06` | Storage | The Architecture **must** support ephemeral storage (non-persistent) storage for Pods. |
-| `req.inf.stg.07` | Storage | The Architecture **must** support persistent storage for Pods (*optional for deployment*). |
+| `req.inf.stg.07` | Storage | The Architecture **must** support persistent storage for Pods. |
+| `req.inf.stg.08` | Storage | The Architecture **must** support the ability for an operator to choose whether or not to deploy persistent storage for Pods. |
 | `req.inf.ntw.01` | Network | The Architecture **must** support Container Network Interface (CNI). |
 | `req.inf.ntw.02` | Network | The Architecture **must** support intra-node communications, such as between agents on a node and all pods on that node |
 | `req.inf.ntw.03` | Network | The Architecture **must** support inter-node communications without NAT, such as communications between pods on a node with all other pods on all nodes |
--- a/doc/ref_arch/kubernetes/chapters/chapter02.md
+++ b/doc/ref_arch/kubernetes/chapters/chapter02.md
@@ -24,9 +24,11 @@
 <a name="2.1"></a>
 ## 2.1 Introduction
 
-**must**: Requirements that are marked as _must_ are considered mandatory and must exist in the reference architecture and reflected in any implementation targeting this reference architecture. The same applies to _must not_.
+**must**: Requirements that are marked as _must_ are considered mandatory and must exist in the Reference Architecture and reflected in any Reference Implementation or Vendor Implementation targeting this Reference Architecture. The same applies to _must not_.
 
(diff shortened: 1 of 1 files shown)


doc/ref_arch/kubernetes/chapters/chapter02.md
```diff
@@ -24,9 +24,11 @@
 <a name="2.1"></a>
 ## 2.1 Introduction
 
-**must**: Requirements that are marked as _must_ are considered mandatory and must exist in the reference architecture and reflected in any implementation targeting this reference architecture. The same applies to _must not_.
+**must**: Requirements that are marked as _must_ are considered mandatory and must exist in the Reference Architecture and reflected in any Reference Implementation or Vendor Implementation targeting this Reference Architecture. The same applies to _must not_.
```
gkunz: nit: targeting => compliant / conformant to

tomkivlin: good spot, have changed it now. Please resolve and approve if you're happy.

tomkivlin:
@fkautz are you happy with this wording?  If so could you approve please?  If so, I think it'll mean we can close #1014.

hdamker:
Just the question if we shouldn't  be more explicit about the optional deployment. Is it a *must* requirement that the operator can choose within a deployment to deploy or not deploy persistent storage. Or *may* the reference/vendor implementation allow this option? 

I'm in favor of a second, explicit requirement with either *may* or *must* for deployment choice. The implicit meaning of "optional for deployment" leaves too many interpretations.

tomkivlin:
> Just the question if we shouldn't be more explicit about the optional deployment. Is it a _must_ requirement that the operator can choose within a deployment to deploy or not deploy persistent storage. Or _may_ the reference/vendor implementation allow this option?

So definitely not "_may_ the reference/vendor implementation allow this option". RI/VI must include it so they can be tested.

The RI/VI are a kind of solution blueprint, not individual solution deployments.  It is these solution deployments where an operator might choose to use the capability or not, depending on their requirements for that specific deployment.

> I'm in favor of a second, explicit requirement with either _may_ or _must_ for deployment choice. The implicit meaning of "optional for deployment" leaves too many interpretations.

What I'm trying to avoid is having requirements in the RA, which aren't about the RA.  More than happy to consider re-wording what we've got here, but want to keep the requirements we list in this doc to be about what this doc is delivering (i.e. the architecture spec).

rabiabdel:
@hdamker please review to get this approved/merged. this will merged within 48 hours if no concern received.

doc/ref_arch/kubernetes/chapters/chapter02.md
```diff
@@ -195,7 +197,7 @@ Reference to "Architecture" in this chapter refers to the NFVI Hardware (e.g. ph
 | `req.inf.stg.04` | Storage | The Architecture **may** support Software Defined Storage (SDS) that seamlessly supports shared block storage, object storage and flat files. |
 | `req.inf.stg.05` | Storage | The Architecture **should** provide high-performance and horizontally scalable storage. |
 | `req.inf.stg.06` | Storage | The Architecture **must** support ephemeral storage (non-persistent) storage for Pods. |
-| `req.inf.stg.07` | Storage | The Architecture **must** support persistent storage for Pods. |
+| `req.inf.stg.07` | Storage | The Architecture **must** support persistent storage for Pods (*optional for deployment*). |
```
hdamker: I propose an **additional** requirement to make the intention of the previous one unambiguously:

req.infr.stg.08 The architecture **must** allow the operator the choice during deployment to not deploy the support of persistent storage.

tomkivlin: Ok, yes this kind of wording works for me as we're asking the architecture to support that optionality (as opposed to the target of the requirement being the implementation or deployment).

tomkivlin: @peterwoerndle , @fkautz what do you think?

pgoyal01: should: Requirements that are marked as should are expected to be fulfilled by the reference architecture but it is up to each service provider to accept an implementation targeting this reference architecture that is **not reflecting** on any of those requirements. 

The above definition could be better worded but doesn't "should" capture what is being proposed?

tomkivlin: I don't think so. We don't want the reference or vendor Implementations to not reflect the requirement, which the wording you've provided would allow. I.e. a vendor implementation could not deliver it and still conform to the specification. Our end goal is that the implementation must reflect the requirement but that the capability is optional when deploying.

pgoyal01: @tomkivlin then the simpler method would be to have a "must" requirement for implementation and another for "may" requirement for deployment

tomkivlin: Which is fine, in those documents. I don't think its appropriate to have a requirement in the RA that has the implementation or deployment as the target of the requirement as we wouldn't be able to test conformance of it. I like Herbert's version as it's about ensuring the RA provides the ability/flexibility for it to be optional during a deployment. That can be tested by checking for dependencies within the specification.

walterkozlowski: Herbert's version makes it really unambiguous.  Let's do it that way.  And follow this path with other requirements that are "optional for deployment".

rabiabdel:
Will approve this and merge it tomorrow if no further feedback received by EOT

hdamker:
> Will approve this and merge it tomorrow if no further feedback received by EOT

@rabiabdel my proposed changes are positively discussed but not yet done within. @tomkivlin: will you do them? It's your PR.

tomkivlin:
> > Will approve this and merge it tomorrow if no further feedback received by EOT
> 
> @rabiabdel my proposed changes are positively discussed but not yet done within. @tomkivlin: will you do them? It's your PR.

Yes, it's on my list to make the change tomorrow morning.

doc/ref_arch/kubernetes/chapters/chapter02.md
tomkivlin: I have updated in [c8e8857](https://github.com/cntt-n/CNTT/pull/1130/commits/c8e8857ab348d40520c2cfdf7b6d9284e12ab422) - if you're happy @hdamker could you resolve both conversations and approve the PR please?

hdamker: @tomkivlin Thanks, I'm happy ;-)

PR #1131, opened 2020-02-20
[RM] TFredberg's Network Layering Diagram v2

markshostak:
Upgraded to v2
Upgraded to v3
Please disregard PR# 1119  (it will be closed)


pgoyal01:
@TFredberg The "CaaS Virtualisation Layer 2" interface to the "IaaS Virtualisation Layer" is marked as "Vi-Ha" but for the "IaaS Virtualisation Layer" "Vi-Ha" is the south bound interface that the  "IaaS Virtualisation Layer" on the northbound side.

TFredberg:
@pgoyal01 First a small clarification that it is Vl-Ha (Virtualization Layer - Hardware resources) and not Vi-Ha. 
It is correct that the "IaaS Virtualisation Layer" in the picture have Vl-Ha both on south-bound and on north-bound interfaces. It is though not the same instances if Vl-Ha on both sides, which is indicated through their different colors.
In reality these Vl-Ha should on a model level be described in the same way, but in the instanciation that I depict to describe a possible deployment case- An exemple can be that "IaaS Virtualisation Layer" south-bound Vl-Ha connects to the HW resources such as an L3 interface and then acts on the data stream e.g. by encapsulating and switching that stream to another L3 interface on its north-bound Vl-Ha. In this example the L3 can follow the same model, but it is likely not the same identical L3 packet stream since it is in some way qualified and encapsulated by the "IaaS Virtualisation Layer".

pgoyal01:
@TFredberg sorry for the typo of using "Vi-Ha" instead of "VI-Ha". The VI-Ha interface is used for requests for virtual resources as a result of a request from the VIM (Nf-Vi) only. Would it not be much cleaner if the "CaaS Virtualisation Layer 2" used the Nf-Vi interface for interactions with the "IaaS Virtualisation Layer"?

TFredberg:
I see that I don´t have deep enough competence of what the NFV reference points include, so I might have misunderstood the ETSI NFI diagram "Execution reference point" to also include the actual packets flowing from the HW Layer into the Virtualization Layer. From this packet flow point of view, I feel the diagram would be correct, but partly agree with both @pgoyal01 and @kedmison about that it would be cleaner if "CaaS Virtualization Layer 2" communicating on Nf-Vi. Anyone that knows where the packet flow reference points or interfaces are described, if not here?

However, since VIM C2 is part of the Virtual Infrastructure Managers it is likely the responsible entity to use Nf-Vi and express its needs of resources and get response from VIM V1 and not the HW Layer or its HW Infrastructure Manager directly.

The intent with colors was to express all possible different administrative domains. This means that a VNF and CNF filling color is one possible administrative domain and its resources offered/allocated from the Virtualization Layer below making the framed color on the VNF/CNF be that same as the Virtualization Layer it "stands on".

Agree with @kedmison comments on the coloring of "CaaS Virtualization Layer 2" borders should then be colored red but the content of the "CaaS Virtualization Layer 2" would have a sperate color as now suggested in the updated picture below.

Regarding @kedmison comment on the depiction and meaning of multiple boxes of HW resources, it likely needs a bit of debate of what these HW resources should mean. 

- With the HW Layer being a composable abstraction layer controlled by the HW Infrastructure Manager, we should likely not show the physical HW components at all since they should not ever be exposed to the Virtualization Layers and certainly not to the VNF/CNF.

- From a CNTT perspective (and where most deployments are today without any HW composability) I guess we mostly care about simplification of physical HW procurement, NFVI support level selection and VNF/CNF component selection that fulfill the wanted Telco Operator network demands. 

- If we in some way would like show the physical HW components there is a need to show how a SmartNIC managed by the HW Layer on a server is provided as a Networking HW on a Computing Hardware and an unmanaged networking switch should be shown below the Managed HW Layer, if it should be shown at all.

I agree with @kedmison that a final diagram in the RM likely should only be a subset of this diagram that is more accurate as a Model and not as an exemplification of a plausible deployment case that we should be able to implement, that I now have drawn it to make some points in this debate.

Agree with @kedmison that file names should be more neutral in the end.

CNTT_NFV_Deployment_Example-1

rabiabdel:
@markshostak @TFredberg what is the status of this? has consensus reached with @kedmison

markshostak:
> 
> The updated image in @TFredberg's comment above looks good. Once the commit is updated with a new file name, and the updated image placed in it, this PR can be approved.

@kedmison Is it possible to change the filename through the GH webUI? Last time, I had to open a new PR, which is sub-optimal. Thx, -Mark

kedmison:
> Is it possible to change the filename through the GH webUI? Last time, I had to open a new PR, which is sub-optimal. 

@markshostak I don't use the GH UI too much but generally speaking you should be able to delete the current one and then add a new file with a different name, as a new commit to the PR.

rabiabdel:
let me know @markshostak if you want help with this.

markshostak:
> 
> let me know @markshostak if you want help with this.

@rabi-abdel Yes. Do you know how to change or add/delete a binary file through the webUI?

kedmison:
Thanks for renaming the graphic.  

The picture in the commit does not match @TFredberg's most recent picture in his [comment above](https://github.com/cntt-n/CNTT/pull/1131#issuecomment-589634302).

markshostak:
@kedmison Done, but be careful what you wish for...

For historical purposes I did not delete the previous version, and since they're under version control. LMK if you prefer I purge it. Thx

markshostak:
@ASawwaf Ahmed, Can you approve pls? Thanks, -Mark

ASawwaf:
@rabi-abdel @markshostak , sorry for the late reply, but I have a small comment that it will be great if we can accommodate it @TFredberg , for "HW infrastructure manager" the name may be misleading to replace VIM/CIM with "HW infrastructure manager"  , we can use "Cloud infrastructure manager" which cover HW+SW for both VNF/CNF especially the title is ETSI and the interface is ETSI interface ( Or-vi ) is between NFVO and VIM not between NFVO - HW infrastructure manager

The 2nd comment,  can we put a small box for SDN-C

PR #1146, opened 2020-02-21
[RA2 Ch02]: Adding a missing br to Ch2.4.4

CsatariGergely:


gkunz:
Is there actually a real need to mix markdown and pseudo HTML in this case? How about just using default markdown enumerations consistently? (yes, we probably wouldn't be able to print the number in bold font)

tomkivlin:
@CsatariGergely what's this fixing?  The ability to have a numbered list in bold?

CsatariGergely:
I'm not sure about this, because this is not a numbered list, but a list of "references" marked with 1), 2), ...
I'm okay to change it, if that is the agreement, but then we need to change it everywhere in a consistent way.

Issue #1106, opened 2020-02-18
[RA1][RC1]Check mandatory Neutron extensions


collivier:
A few Neutron extensions may be linked to the reference implementation:
- availability_zone
- l3_agent_scheduler
- l3-ha
- dvr
- l3-flavors
- fip-port-details
- floatingip-pools
- filter-validation
- extraroute-atomic

A straightforward proposal is to double check them via OVN.

collivier:
availability_zone is supported by OVN
https://docs.openstack.org/networking-ovn/latest/admin/features.html

neutron_tempest_plugin.api.admin.test_agent_availability_zone.AgentAvailabilityZoneTestCase.test_agents_availability_zone is linked to
- AGENT_TYPE_DHCP = 'DHCP agent'
- AGENT_TYPE_L3 = 'L3 agent'

ASawwaf:
@collivier , Can you elaborate more on what is scope of this issue?

collivier:
2 are out of Pike:
-floatingip-pools: https://github.com/openstack/neutron/commit/4e3fb3191927e07b37fe0937102dc17334e906d1#diff-6cdd456e9b1cad4a23c219ecef1ecdae
- extraroute-atomic:
https://github.com/openstack/neutron/commit/b3eda71638973836c68dca90da9e017b285a642c#diff-25237435a98c84c53fbb8ae0b1c82c00

For the time being, only the next extension may change:
l3_agent_scheduler
l3-ha
dvr
l3-flavors
fip-port-details
filter-validation

collivier:
@ASawwaf : I'm double checking a few mandatory extensions which may force the implementation of 
Neutron. The first list has to be double checked via a few reference impl.

collivier:
@ASawwaf It's not so simple because I'm also working from OpenStack master to OpenStack pike dealing with testing logic. I'm adding comments to ease writing test case list at the end.
+ https://bugs.launchpad.net/neutron/+bug/1863707

collivier:
@ASawwaf I will create the opposite thread later. dns integration could be added as mandatory.

collivier:
port-mac-address-regenerate, filter-validation and  fip-port-details were introduced in Rocky (oldest OPNFV release: Hunter)
https://github.com/openstack/neutron/commit/2b1d8ea4a202f24c1b485ccebdbf831c505a4e6a
https://github.com/openstack/neutron/commit/8f3a066b20b7ffafec95a618d60e40727504f37c
[...]

PR #1149, opened 2020-02-21
[RA1][RC1] Update Regex list in RC

collivier:
It update tempest_slow regex (Neutron) according to the last changes in
CNTT RA1 Chapter5 (l3_agent_scheduler is now optional).

Closes: #1106

### DIFF
--- a/doc/ref_cert/lfn/chapters/chapter03.md
+++ b/doc/ref_cert/lfn/chapters/chapter03.md
@@ -296,6 +296,7 @@ the following test names must not be executed:
 | .\*test_routers.RoutersIpV6Test.test_create_router_set_gateway_with_fixed_ip                       | https://launchpad.net/bugs/1676207    |
 | .\*test_routers.RoutersTest.test_create_router_set_gateway_with_fixed_ip                           | https://launchpad.net/bugs/1676207    |
 | .\*test_network_v6                                                                                 | https://gerrit.opnfv.org/gerrit/69105 |
+| .\*test_network_basic_ops.TestNetworkBasicOps.test_router_rescheduling                             | l3_agent_scheduler                    |
 
 Neutron API is also covered by [Rally](https://opendev.org/openstack/rally).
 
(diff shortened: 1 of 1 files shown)


collivier:
The generated links are right: https://cntt-n.github.io/CNTT/doc/ref_arch/openstack/chapters/chapter05.html

Reason is why the tests are excluded mostly because the extensions covered are optional.
That's the difference between IaaS verification as Functest has done for years and conformance where you must check that all mandatory capabilities and nothing else more.

collivier:
All regex are negative. It runs all tests expect the regex listed. The other way (white list) is wrong and can be hardly maintained.

collivier:
This patch is straightforward and should be merged quickly. It completes the previous one which was much more difficult.

mf4716:
The link works yes, I will add a few words to clarify 

> The generated links are right: https://cntt-n.github.io/CNTT/doc/ref_arch/openstack/chapters/chapter05.html
> 
> Reason is why the tests are excluded mostly because the extensions covered are optional.
> That's the difference between IaaS verification as Functest has done for years and conformance where you must check that all mandatory capabilities and nothing else more.

The link in the document works, yes.  What was not clear is the reasoning for the statement, "test names must not be executed".  The reason in the subsequent table just listed a capability as the reason, but with no explanation.  Therefore, before approving, I will edit the paragraphs and words to explicitly mention that the capability is not required, therefore, it's not a required test for RA-1 validations.   I'll update the Table Headers to reflect this too.  I'll add it as a commit to this PR, but please edit/modify if needed.  Once my change is in I'll approve the PR.

collivier:
It's not the first time and I would appreciate you stop changing by yourself other PRs.
It's unusual in opensource without the agreement of the author.

The scope is to complete (0.1%) the previous changes.
Please open your own PR which I can easily review. I don't want to vote -1 on my own PR because of your changes.
Then I'm reverting this PR to its original Functest-related content to encourage the normal collaborative process

Le me know if we disagree about RC lead scope.
Be free to vote -1 once I revert your commit.

collivier:
By the way the same change was merged on Functest by pair review.

PR #1159, opened 2020-02-21
[RA2 Ch02]: Changing req.kcm.09 from should to must

### LINKED ISSUE #1132: [RA 2] req.kcm.09 - The Architecture **should** prevent workloads from starving other workloads of guaranteed resources.
In PR #1121 , `req.kcm.09` The Architecture **should** prevent workloads from starving other workloads of guaranteed resources.

The requirement should be a **must**.

CsatariGergely:
Fixes #1132


ASawwaf:
@CsatariGergely , Just inquiry here, Do we need to have CPU pinning,.... to achieve it  ?

PR #1161, opened 2020-02-21
[RM Ch04]: Adding links ot the ETSI NFV specs

CsatariGergely:


karinesevilla:
@CsatariGergely 
It should be good to add also the reference to ETSI NFV-TST 008 clause 6 for Host CPU utilization. This metric is specified there.

CsatariGergely:
> 
> 
> @CsatariGergely
> It should be good to add also the reference to ETSI NFV-TST 008 clause 6 for Host CPU utilization. This metric is specified there.

Added

CsatariGergely:
> 
> 
> The ETSI links have versions in them.
> This is probably a good thing, to ensure consistency between Baldy and the ETSI documents at time of Baldy publication.
> 
> Would it be useful to put the version in the text as a reminder that periodically the link targets should be refreshed and re-considered?

Correct, versions added.

rabiabdel:
@kedmison, @karinesevilla please review/approve

PR #1167, opened 2020-02-22
[RA-1 Ch04] Minor Corrections

### LINKED ISSUE #1088: [RA-1 Ch04] Corrections
| Section | Issue | Resolution |
|-------|--------|---------|
|  4.2.2.5 | “The software components are as specified ..” | “The software and hardware configurations are as specified ..” |
|  4.2.2.5 | URL to RM 5.4 | Make it relative |
|  4.2.2.6 | “The specifications for this type of resource pooling is specified in 4.5.2.” |
|  “The specifications for this type of resource pooling is specified in 4.5.3.” | Non-existent section in RM, RA-1. |
|   |
|  Suggest delete sentences |
|  4.2.2.6 | “Multiple pools of hardware resources … profile (for example, network intensive). Leads to efficient use ...” | Typo: delete period before and lower case “Leads” |
|  4.2.3.4 |  “[FWaaS (Firewall as a Service)[(https://opendev.org/openstack/neutron-fwaas/)” | Correct linking for FWaaS: change italicized to closing ‘]’. |
|  Also, the link is to dev repo. Change to: |
|  https://docs.openstack.org/neutron/pike/admin/fwaas.html  |
|  4.3.1. | “A high level overview … their sizing rules (to be developed). | Delete “ (to be developed)” |
|  4.3.1. | Should we add “stateless/stateful” for the service components? |  |
|  4.3.10.  | Cyborg | 4.3.1. VIM Core Services; where to place “Cyborg” |
|  4.4.1. | Table is about flavors – CPU Allocation ratio is however about implementation.  | Replace the Nova.conf include to “In flavor create or flavor set specify |
|  --property cpu_allocation_ratio=4.0 (for Basic) and 1.0 for N and C |
|  4.4.1. | Add SMT as part of flavor definition | Replace the Nova.conf include to “In flavor create or flavor set specify |
|  --property smt_enabled=True |
|  4.4. | Do we need a section on SW and HW Host profiles? | Show how flavors are actually implemented in the hosts |
|  4.7  | Architectural Drivers – Requirements Traceability | Delete from TOC and content |

scottsteinbrueck:
Issue Fixes #1088
Fixed defects that could be found.
**Did not address the following, could not find a bug:**

4.3.1:  “A high level overview … their sizing rules (to be developed).
    >> Delete “ (to be developed)”
    >> @scottsteinbrueck says could not find bug
4.7:  Architectural Drivers – Requirements Traceability
    >> Delete from TOC and content
    >> @scottsteinbrueck says could not find bug

**Did not address the following, consider a new issue to address:**

4.3.1:  Should we add “stateless/stateful” for the service components?
    >> @scottsteinbrueck says @pgoyal01 to consider a new issue to address
4.3.10:  Cyborg
    >> 4.3.1. VIM Core Services; where to place “Cyborg”
    >> @scottsteinbrueck says @pgoyal01 to consider a new issue to address
4.4:  Do we need a section on SW and HW Host profiles?
    >> Show how flavors are actually implemented in the hosts
    >> @scottsteinbrueck says @pgoyal01 to consider a new issue to address

### DIFF
--- a/doc/ref_arch/openstack/chapters/chapter04.md
+++ b/doc/ref_arch/openstack/chapters/chapter04.md
@@ -422,8 +422,6 @@ To configure the T-shirt sizes (specified in [Table 4-17](../../../ref_model/cha
 
 In addition, to configure the storage IOPS the following two parameters need to be specified in the flavor create: --property quota:disk_write_iops_sec=<IOPS#> and --property quota:disk_read_iops_sec=<IOPS#>.  
 
-To configure SMT, the following parameter needs to be specified in the flavor create:  --property smt_enabled=True.
-
 The flavor create command and the mandatory and optional configuration parameters is documented in https://docs.openstack.org/nova/latest/user/flavors.html.
 
 <a name="4.4.2"></a>
--- a/doc/ref_arch/openstack/chapters/chapter04.md
+++ b/doc/ref_arch/openstack/chapters/chapter04.md
@@ -98,9 +98,9 @@ For OpenStack control nodes we use the BIOS parameters for the basic profile def
 -	Sizing rules
 
 #### 4.2.2.5. Compute Nodes
--	The software components are as specified in the [Reference Model chapter 5.4](https://github.com/cntt-n/CNTT/blob/master/doc/ref_model/chapters/chapter05.md#5.4)
+-	The software and hardware configurations are as specified in the [Reference Model chapter 5.4](../../../ref_model/chapters/chapter05.md#5.4)
 -	BIOS requirement
-    -	The general bios requirements are described in the [Reference Model chapter 5.4](https://github.com/cntt-n/CNTT/blob/master/doc/ref_model/chapters/chapter05.md#5.4)
+    -	The general bios requirements are described in the [Reference Model chapter 5.4](../../../ref_model/chapters/chapter05.md#5.4)
     -	Additionally, for OpenStack we need to set the following boot parameters:
 
 | BIOS/boot Parameter | Basic  | Network Intensive |
@@ -152,8 +152,8 @@ Caveats:
 -	Affinity and anti-affinity rules, among other factors, affect the sizing
 
 #### 4.2.2.6. Compute Resource Pooling Considerations
--	Multiple pools of hardware resources where each resource pool caters for workloads of a specific profile (for example, network intensive). Leads to efficient use of the hardware as the server resources are specific to the flavour. If not properly sized or when demand changes can lead to oversupply/starvation scenarios; reconfiguration may not be possible because of the underlying hardware or inability to vacate servers for reconfiguration to support another flavour type. The specifications for this type of resource pooling is specified in 4.5.2.
--	Single pool of hardware resources including for controllers have the same CPU type. This is operationally efficient as any server can be utilized to support a flavour or controller. The single pool is valuable with unpredictable workloads or when the demand of certain flavours is insufficient to justify individual hardware selection. The specifications for this type of resource pooling is specified in 4.5.3.
+-	Multiple pools of hardware resources where each resource pool caters for workloads of a specific profile (for example, network intensive) leads to efficient use of the hardware as the server resources are specific to the flavour. If not properly sized or when demand changes can lead to oversupply/starvation scenarios; reconfiguration may not be possible because of the underlying hardware or inability to vacate servers for reconfiguration to support another flavour type.
+-	Single pool of hardware resources including for controllers have the same CPU type. This is operationally efficient as any server can be utilized to support a flavour or controller. The single pool is valuable with unpredictable workloads or when the demand of certain flavours is insufficient to justify individual hardware selection.
 
 <a name="4.2.3"></a>
 ### 4.2.3. Network Fabric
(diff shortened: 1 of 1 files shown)


pgoyal01:
Fixes #1088

doc/ref_arch/openstack/chapters/chapter04.md
```diff
@@ -401,7 +401,7 @@ New Table w/o Compute Intensive column below
 
 | Flavor Capabilities | Reference<br>RM Chapter 4 and 5 | Basic | Network Intensive | 
 |----------|-------------|--------------|-------------|
-| CPU allocation ratio | nfvi.com.cfg.001| In Nova.conf include <br>cpu_allocation_ratio= 4.0 | In Nova.conf include <br>cpu_allocation_ratio= 1.0 |
+| CPU allocation ratio | nfvi.com.cfg.001| In flavor create or flavor set <br>--property cpu_allocation_ratio=4.0 | In flavor create or flavor set <br>--property cpu_allocation_ratio=1.0 |
```
deebhatia: I believe what you are after is CPU Policy and already covered by `nfvi.com.cfg.003`. This is a compute host configuration param to allow that.

doc/ref_arch/openstack/chapters/chapter04.md
```diff
@@ -420,7 +420,9 @@ To configure the T-shirt sizes (specified in [Table 4-17](../../../ref_model/cha
 | .2xlarge* | 8<br>-- vcpus 8 | 16 GB<br>-- ram 16384 | 160 GB<br>-- disk 160 |
 | .4xlarge* | 16<br>-- vcpus 16 | 32 GB<br>-- ram 32768 | 320 GB<br>-- disk 320 |
 
-In addition, to configure the storage IOPS the following two parameters need to be specified in the flavor create: --property quota:disk_write_iops_sec=<IOPS#> and --property quota:disk_read_iops_sec=<IOPS#>.
+In addition, to configure the storage IOPS the following two parameters need to be specified in the flavor create: --property quota:disk_write_iops_sec=<IOPS#> and --property quota:disk_read_iops_sec=<IOPS#>.  
+
+To configure SMT, the following parameter needs to be specified in the flavor create:  --property smt_enabled=True.
```
deebhatia: Can you add a source of this statement? I am not able to find `smt_enabled` property in openstack flavor.

rabiabdel:
@scottsteinbrueck can you please address @deebhatia input above?

doc/ref_arch/openstack/chapters/chapter04.md
pgoyal01: @deebhatia The table is for flavor definitions and not the implementation

doc/ref_arch/openstack/chapters/chapter04.md
pgoyal01: @deebhatia This was provided by me. My mistake -- needed to be clearer. As you know, in the flavor extra-specs it is possible to assign extra key-value properties. This statement needs to be accompanied by further details (reason why it was not added to the table). The main intent is to be able to have host aggregates with hosts with smt enabled and smt disabled (latter being necessary for secure workloads).

pgoyal01: @deebhatia on further thought, given that, both flavors require SMT to be enabled (no choice for SMT to be disabled), I will suggest to Scott to remove the statement.

deebhatia: It is not an issue in keeping it. it is just that for custom flags like host aggregate flags, we should either explicitly mention or prefix them `aggregate_instance_extra_specs` or do both. Currently, it looks like a standard flavor extra specs which might confuse the reader.

doc/ref_arch/openstack/chapters/chapter04.md
deebhatia: @pgoyal01 `cpu_allocation_ratio` is not a standard flavor specification. Can you explain what is the intent with this property here?

pgoyal01: @deebhatia the cpu_allocation_ratio option is specified in  nova.conf and in the host aggregates metadata. Since flavors themselves have the cpu allocation as part of their specification (originally CNTT Basic with cpu_allocation_ratio of 4.0 and Network Intensive with 1.0) it is a good practice to capture that meta data in the flavor also.

pgoyal01: @deebhatia BTW this section has existed in the chapter for quite a while. Scott only made typo corrections.

deebhatia: Understood. I will iterate the same here that for custom flags like host aggregate flags, we should either explicitly mention or prefix them `aggregate_instance_extra_specs` or do both.

doc/ref_arch/openstack/chapters/chapter04.md
scottsteinbrueck: @deebhatia deleted offending sentence.

doc/ref_arch/openstack/chapters/chapter04.md
deebhatia: @pgoyal01 Any suggestion for custom specs?

pgoyal01: @deebhatia If the AggregateInstanceExtraSpecsFilter is enabled then it will match specifications taht are scoped with "aggregate_instance_extra_specs" (as suggested by you).  The extra_specs as described are correct but maybe we can add a "Note" for the case where the "AggregateInstanceExtraSpecsFilter" is enabled. But this change should be made on its own -- new Issue and PR so that we capture why th change. The current PR is for minor corrections. Do you agree?

deebhatia: Adding details about `AggregateInstanceExtraSpecsFilter` can be taken as separate PR.
This PR can just mention that `it is a custom extra spec.`

pgoyal01:
@iangardner22 should that be a separate issue and not part of the minor cleanup

doc/ref_arch/openstack/chapters/chapter04.md
scottsteinbrueck: @deebhatia I added the phrase as you suggested.

deebhatia: thanks.

iangardner22:
> @iangardner22 should that be a separate issue and not part of the minor cleanup

Yes could be...  want me to open one?

PR #1185, opened 2020-02-26
[RA-2 Ch04] Removing a requirement that pegs the Host OS to debian/ubuntu/rhel/ce…

fkautz:
…ntos/fedora

There are a plethora of kubernetes deployments that do not run on deb/rpm enabled systems. These include all the major cloud providers such as GKE/EKS/AKS. Other deployments we may start seeing on the edge such as flatcar/rancher will likely not provide us with rpm/deb package management.

We also don't want to encourage people to use rpm/deb to install things into the infrastructure. These should be installed and hosted in K8s itself.


CsatariGergely:
I do not have a problem to remove the requirement, some thing I do not get from the last part of @fkautz  -s comment. What do you mean on "These should be installed and hosted in K8s itself"?

tomkivlin:
@fkautz the deb/rpm requirement is so that a container runtime and Kubernetes itself can be installed.  That line was from the Kubernetes documentation.  Happy to re-word it but it's to ensure that the "user space" components of a Kubernetes cluster (runtime, etc.) can be installed and lifecycle-managed.

rabiabdel:
@fkautz please make sure to name PR/Issue in the format [WSx Chx].

fkautz:
@rabiabdel I'm not sure which WSx I am supposed to name this with. Can you provide assistance?

fkautz:
@CsatariGergely often, many of these components can be installed on the host or in a container. CNI is a great example. It sometimes lives in `/opt/cni/bin` and sometimes it lives in a container, e.g. `kubectl apply -f https://docs.projectcalico.org/manifests/calico.yaml` installs Calico.

jnummelin:
> the deb/rpm requirement is so that a container runtime and Kubernetes itself can be installed.

There's plenty of other ways to install and run core k8s components, some k8s distros run everything in containers. And as @fkautz mentioned, many "core" components such as CNI providers do install (and manage the lifecycle) the needed bins via DaemonSet pods. Another such example is device accelerators such as GPU drivers, the direction on those seems to be to have an init container that sets up the host side setup. No dep/rpm etc needed.

tomkivlin:
> some k8s distros run everything in containers

> No dep/rpm etc needed.

No deb/rpm is needed to install Kubernetes, or CNI, correct - that's not what I've said though.

You're quite right that k8s control plane, CNI, Device Plugin, init-containers etc are all valid, but what are those containers being executed in or by?  A container runtime, right?  In user space of a Container Host Operating System?

My point is, how is that container runtime installed / package managed?  That's what the deb/rpm line is about - package management **of the container runtime** in the Host OS only - not about package management of Kubernetes or related add-ons like CNI etc.

More than happy to modify to ensure all scenarios such as public cloud, private cloud, edge cloud, IoT, etc. are covered, but want to be sure we're talking about the same thing.

jnummelin:
> My point is, how is that container runtime installed / package managed?

Why does it matter how we get container runtime in place? As long as its CRI compliant runtime it should be enough. And from CNTT perspective that is implied by the fact that we require certified/conformant k8s. Or am I missing something fundamental? (very much a possibility always)

If we want to imply that the container runtime installation has to be somehow "managed" (upgradeable etc.), we could word it something like:
```
Host OS _should_ support package/component management solution up to container runtime level or provide the container runtime as a pre-built component of the OS distribution.
```

IMO _should_ is correct term to use here as there are OSes built specifically for containers that do not really have package management. E.g. CoreOS in the past, and newer RedHat variants of it now, had docker/rkt "pre-built" into the image and no traditional package management at all.

CsatariGergely:
RM2 meeting on 2020.03.19 agreed on that it is okay to remove the packaging requirement.

PR #1187, opened 2020-02-26
[RM Ch04] CPU overbooking

### LINKED ISSUE #1126: [RM Ch04] CPU overbooking
Section 4.1.4.2 Table 4.6, the definition of "CPU overbooking" is missing and the capability is expressed with the ratio 1:N rather than N:1 as commonly used and defined in chapter 5 5.1.1
@kedmison  I propose to use the definition and wording of chapter 5 and modify the chapter 4 as follow:
- rename "CPU overbooking" by "CPU allocatio ratio"
- def: Number of virtual cores per physical core
- unit: N:1
Impact for tables 4.6 and 4.24

karinesevilla:
Fixes #1126 

Section 4.1.4.2 Table 4.6:
-  "CPU overbooking" renamed  "CPU allocatio ratio"
- definition provided " Number of virtual cores per physical core"
- expression of the ratio changed for consistency

Same modifications applied to table 4.24

### DIFF
--- a/doc/ref_model/chapters/chapter04.md
+++ b/doc/ref_model/chapters/chapter04.md
@@ -191,7 +191,7 @@ This section covers a list of implicit NFVI capabilities and measurements that d
 
 | Ref | NFVI capability | Unit | Definition/Notes |
 |--------------------|------------------------------------------|--------|---------------------------------------------------------------------------------------------------------------------|
-| i.nfvi.cap.016 | CPU overbooking | 1:N | <definition update scheduled> |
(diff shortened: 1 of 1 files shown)


markshostak:
@karinesevilla Hi Karine, When you say Number of virtual cores per physical core. are you referring to number of Threads (or Processes or VMs) scheduled per physical core? 

Consider including "overbooking" in the comments section; it's an easy way to help people understand the parameter, and still have the more technical definition from above which is more precise, but not as intuitive IMHO.

PR #1189, opened 2020-02-26
[RA2 CH03]-Broken link

prabhubalan:
Broken link due to "/ref_model" repeated twice. Can anyone guide me, is this pointing to the right section. I believe it is supposed to point to 5.1 for software profile description and 5.3 for hardware profile description right?


xmulligan:
I agree these should actually be pointing to 5.1 for software profile description and 5.3 for hardware profile description

xmulligan:
@prabhubalan can you fix the links to point to the right sections and resubmit?

prabhubalan:
> I agree these should actually be pointing to 5.1 for software profile description and 5.3 for hardware profile description

PR #1190, opened 2020-02-26
[RA2 Ch0x] More consistent use of conformance vs compliance in RA2

### LINKED ISSUE #988: [ALL] - We need to be more consistent on "compliant" vs "conformant"
From deep dive in Prague:
We need to be more consistent on "compliant" vs "conformant"

gkunz:
Attempt to  more consistently use the term "conformance" instead of
"compliance" in order to follow the agreed terminology.

Partially fixes #988


doc/ref_arch/kubernetes/chapters/chapter01.md
```diff
@@ -23,7 +23,7 @@ This Reference Architecture for Kubernetes will describe the high level system c
 
 To assist with this goal, we could create an "RA2 [Razor](https://en.wikipedia.org/wiki/Philosophical_razor)" along the lines of "if something is useful for non-Telco workloads, we should not include it only for Telco workloads". For example (thanks to @fzdarsky in #383) lets say we start with a vanilla Kubernetes (say, v1.16) feature set in the Reference Architecture, and we provide clear evidence that a functional requirement cannot be met by that system (say, multi-NIC support), we only then add the least invasive, Kubernetes-community aligned extension (say, Multus) to fill this gap. If there are still gaps that cannot be filled by Kubernetes community technologies or extensions then we need to concisely document the requirement and take it to the relevant project maintainers.
```
rabiabdel: Offcourse Thanks to @fzdarsky :) do we need to have it in the official specs though :)

gkunz: This is a bit unrelated to this particular PR, but I agree that we should remove this (no offense, Frank) ;-) Moreover, since we are on this topic (despite off-topic from this PR), this paragraph should be re-phrased such that it does not read like a proposal, but rather describes the agreed approach to define a RA2 from a current vanilla Kubernetes. I personally agree to this approach. However, I suggest to create a separate issue to track this.

rabi-abdel: OK fair enough :)

gkunz: ok, I'll create a new issue for this.

tomkivlin: Perhaps not, but can I suggest you do separate issue/PR given this text isn't related to this PR.

PR #1196, opened 2020-02-27
[RM] Delete ch06_ref_nfvi_hw_profiles.PNG

### LINKED ISSUE #472: [RM CH06] duplicate figures with case-insensitive name collisions
There are RM Chapter 6 figures that have names that differ only in case.  This creates warnings on case-insensitive filesystems, when sync-ing the repo.  One should be selected, and the other removed from the repo.

```
warning: the following paths have collided (e.g. case-sensitive paths
on a case-insensitive filesystem) and only one from the same
colliding group is in the working tree:

  'doc/ref_model/figures/ch06_ref_nfvi_hw_profiles.PNG'
  'doc/ref_model/figures/ch06_ref_nfvi_hw_profiles.png'
```

kedmison:
Delete duplicate diagram.  
Fixes #472.


rabiabdel:
> Okay from my side.
> @rabi-abdel What is the approach towards dealing with images having Compute intensive profile?

@deebhatia if the picture contains compute intensive profile, we need to remove the compute intensive profile from it. Images source code is in the artefacts folder.

kedmison:
Neither this graphic nor it's peer with the lowercase suffix are referred to anywhere in the repo.

PR #1234, opened 2020-03-04
RA2 Ch4 - update 4th bullet

### LINKED ISSUE #990: [RA2] Ch4 - 4th bullet of Chapter 4.2 needs rewording
From deep dive in Prague:
4th bullet of Chapter 4.2 needs rewording

tomkivlin:
Use wording from CNCF TUG Cloud Native Principles document.

Fixes #990


tomkivlin:
> Based on the titile, I think the intention is just to replace bullet 4? However, bullet 5 is also removed, is it on purpose?

My memory of the discussion in Prague was that the 5th bullet didn't make sense given the 4th bullet.  Happy to add it back in however if the consensus is to do so.

yangxu-hw:
> > Based on the titile, I think the intention is just to replace bullet 4? However, bullet 5 is also removed, is it on purpose?
> 
> My memory of the discussion in Prague was that the 5th bullet didn't make sense given the 4th bullet. Happy to add it back in however if the consensus is to do so.

Thanks for the clarification. I agree bullet 5 seems not needed considering the given bullet 4.

PR #1223, opened 2020-03-02
[RA-1 Ch02]  Requirmeents and Recommendations

pgoyal01:
In the RA-1 Technical call of March 2, 2020, it was decided to create Requirements and Recommendations.

Fixes #1222 
Fixes #1076

PR #1245, opened 2020-03-04
[RA1] Add API requirements

collivier:
It creates a few additional requirements related to the API exposed.
They will be added in a second change to the testcases covering this
requirements.

### DIFF
--- a/doc/ref_arch/openstack/chapters/chapter05.md
+++ b/doc/ref_arch/openstack/chapters/chapter05.md
@@ -49,6 +49,10 @@ For the purpose of compliance tests, this chapter also identifies the set of the
 | security_compliance     | X             |
 | trust                   | X             |
 
+| Ref #            | Category | Description                                                                                                                         |
+|------------------|----------|-------------------------------------------------------------------------------------------------------------------------------------|
+| `req.vim.api.01` | infra    | must expose equal Keystone API version, equal or newer Keystone API microversion and all mandatory Keystone features listed by CNTT |
+
 Identity API v3: https://docs.openstack.org/api-ref/identity/v3/index.html
 
 Identity API v3 extensions: https://docs.openstack.org/api-ref/identity/v3-ext/
@@ -60,6 +64,10 @@ Security compliance and PCI-DSS: https://docs.openstack.org/keystone/train/admin
 | **OpenStack Service** | **API Version** | **Minimal API Microversion** |
 |-----------------------|-----------------|------------------------------|
 | Image: Glance         | v2              | 2.5                          |
+
+| Ref #            | Category | Description                                                                                    |
+|------------------|----------|------------------------------------------------------------------------------------------------|
+| `req.vim.api.02` | infra    | must expose equal Glance API version and equal or newer Glance API microversion listed by CNTT |
 
 Image Service Versions: https://docs.openstack.org/api-ref/image/versions/index.html#version-history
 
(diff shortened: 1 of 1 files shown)


pgoyal01:
@collivier can you move these requirements to RA-1 Ch02? "equal to" what? shouldn't the API version be tied only to the Pike supported API version/micro-version? 

W.r.t. the API extensions, not all of these extensions can be mandatory as some may not have any use. For example, under what circumstances would an OAUTH extension be useful for an operators' private network cloud (not general purpose cloud)?

collivier:
@pgoyal01 The proposal here is to write the requirements related to Chapter RA1 Chapter5 as asked by @rabi-abdel and @mf4716 . I'm fine to move it before, in Chapter2, if it helps.
I note that we are looping here with the Airship results which is out of the scope of this PR (be free to open any new ticket).
BYW As mentioned in RI/RC, we should first look at Keystone history, design and the true fix in Airship (it's not the classical 404 errors). We shouldn't fail our main goal due to the early results by Airship without any technical dive. Else we could rather consider other implementations. Functest SUTs are passing very well and I can share several other test results.

pgoyal01:
@collivier In my previous suggestion and question, I was not considering Airship or functest. My sole consideration was from an architecture perspective. 

The API extensions support needs to be in the Recommendation section of RA-1 Ch02 PR#1223. Since this PR has not yet been merged, I allow you to add to RA-1 Ch02 Section 2.4.3. "VIM Recommendations." Please reword so it is clears as to the ask. Thanks

collivier:
@pgoyal01 thank you for the clarification. For my understanding we should avoid API case by case as much as possible. The feature selections are also driven by the service design.
I will dive into Airship deployment. As it's only about 3 conf data, it may be trivial to fix.

Regarding the rewording proposed, would you ask for the versions in requirements?
| `req.vim.api.06` | infra    | must expose Nova API version 2.1, Nova API microversion 2.53 or newer and all Nova features listed as mandatory by CNTT |

pgoyal01:
@collivier instead of  "infra | must expose Nova API version 2.1, Nova API microversion 2.53 or newer and all Nova features listed as mandatory by CNTT" how about "The Platform must expose the APIs for each of the OpenStack services' API version and micro-versions for the CNTT Reference Architecture's OpenStack release." or something similar

collivier:
@pgoyal01 That sounds good. All Functest testcases will refer to this req which may be the executive traceability expected.

karinesevilla:
@collivier @pgoyal01 
I agree that precise requirements for each OpenStack service API was missing and it is needed for RC. 
To be consistent with RA1 document, requirements are all gathered in chapter 2, for me it is the right place to put them. But the details  of version, micro versions and features can stay on Chapter 5. Can we add in chapter 2 a requirement per OpenStack service with a reference to to chapter 5 where APIs are fully specified? Doing this if a change has to be made in version, micro-versions or features it will only impact Chapter5.

pgoyal01:
@karinesevilla do we need to add per service or can we use a blanket and point to Chapter 5 for details?

rabiabdel:
@collivier @pgoyal01 @karinesevilla @iangardner22 i suggest moving the requirement to Ch02 as per @pgoyal01 proposal: 
"The Platform must expose the APIs for each of the OpenStack services' API version and micro-versions as listed in this document" The exact version and micro-versions are already listed in Ch05.

rabiabdel:
@collivier are you happy with the suggestions below?

karinesevilla:
Per RA1 meeting, 9th of March, decision to move the requirements to Chapter2

pgoyal01:
@rabi-abdel There is  a requirement in RA-1 Ch02 that can be modified as per your verbiage; the following currently exists:
| `req.int.api.01` | API | The Architecture must provide Control API endpoints to cloud platform core services. |
|----|----|-----|

To cover the point about the "latest version/microversion", may I suggest adding another requirement as follows:
| `req.int.api.04` | API | The Architecture must expose the latest version and microversion of the APIs for the given CNTT OpenStack release for each of the OpenStack core services |
|----|----|-----|

or some similar but simplified verbiage.

collivier:
Taken into account into https://github.com/cntt-n/CNTT/pull/1223
Closing this former one

Issue #1160, opened 2020-02-21
[RM] Introduce availability zones to Reference Model


kedmison:
Availability zones are a common construct provided to allow workloads to create a more highly available application on top of an infrastructure.  They communicate important information about shared failure domains for infrastructure.  

Availability zones, as a construct, are available in OpenStack and in Kubernetes. 

In the absence of multiple availability zones, workloads rely on anti-affinity.  For large distributed workloads, this can create issues in requiring a large number of infrastructure hosts, thus creating inefficiency in the host hardware allocation.  

Introducing availability zones would allow for 
- VNF vendors to create better availability strategies for their VNFs
- NFVI operators to push VNF vendors to minimize sprawl across the infrastructure

pgoyal01:
@kedmison are you suggesting adding a requirement around Availability Zones (AZs)? (I think you are referring to AZs in the context of hosts (nova AZs) and not network/storage). I hope you agree that AZs do not negate the need for anti-affinity and that the use of AZs as an anti-affinity mechanism is rare. Would the use as an anti-affinity mechanism require a large number of AZs? Please remember that with AZs, as operators, we are dividing our resources into separate resource pools with implications on capacity availability as the utilisation across these pools is unlikely to be the same. 

Typically, operators use AZs as failure domains-- enable workloads to achieve higher levels of availability as a mitigation against single points of failure. This latter aspect is highly operator environment dependent (racks with single or redundant ToR switches, set of racks connected to one power sources and the rest connected to another power source, etc.).  To guard against single points of failure, operators may implement 2 (based on needed redundancy) or 3 for quorum based workloads (databases, clusters, ...). AZs may also be used as a availability mechanism in certain maintenance activities (as in allowing rolling updates).

kedmison:
@pgoyal01 we are aligned on the context of AZ.  I am referring to using AZs as a way for the operator to indicate acceptable failure domain separation, yes.  The operator could choose to only separate the compute nodes into availability zones, and not the storage or the networking.  Or they could choose to separate all 3.  In the absence of AZ definitions, VNF vendors have to look for other options to leverage, and anti-affinity is one of the sole remaining tools, albeit one with drawbacks.

ASawwaf:
@kedmison @pgoyal01 , I think Host Agg  will be the right way to address what @kedmison mention

pgoyal01:
[...]

PR #1250, opened 2020-03-04
[RM] Add availability zones to RM

kedmison:
This change adds Availability Zones to the RM, and will need to trigger downstream changes in RA, RI, and RC.


kedmison:
Fixes #1160.

doc/ref_model/chapters/chapter03.md
```diff
@@ -128,7 +129,20 @@ _**Example**: a virtual compute descriptor as defined in TOSCA Simple Profile fo
 | `subnet` | network address of the subnet |
 | `acceleration` | key/value pairs for selection of the appropriate acceleration technology |
 
-<p align="center"><b>Table 3-4:</b> Attributes of network resources.</p>
+<p align="center"><b>Table 3-4:</b> Attributes of network resources</p>
+
+<a name="3.2.5"></a>
+### 3.2.5 Availability Zone
+An Availability zone is a logical grouping of resources (e.g. compute, block storage, network).  These logical groups allow cloud operators to segment their physical resources pools based on factors such as location (rack, datacenter), or indirect failure domain dependencies like power sources.  Workloads can leverage availability zones to avoid sharing failure domains for a workload, and thus increase its fault-tolerance.
```
pgoyal01: "An Availability zone is a logical grouping of resources" can we change to ""An Availability zone is a logical pool of  physical resources"? 

Do we have to state "(compute, block storage, network)'? We should not want to give the impression that each Availability Zone (AZ) will have their own storage and network.

"These logical groups allow cloud operators to segment their physical resources pools based on factors such as location (rack, datacenter), or indirect failure domain dependencies like power sources." can we change to ""These logical pools segment their physical resources of a cloud based on factors chosen by the cloud operator. The cloud operator may create availability zones based on location (rack, datacenter), or indirect failure domain dependencies like power sources."

tomkivlin: @pgoyal01 I agree with points 1 and 3, but don't understand the middle one.  Why wouldn't an AZ have it's own storage or network?  Surely that's what we want, to reduce the failure domain when an AZ fails?

pgoyal01: @tomkivlin I should have been clearer - we can't require the "AZ" failure domain in case of networks and storage as operators may support high availability of these for various reasons.  For tenant workloads on computes, AZ is an appropriate failure domain mechanism.

kedmison: @pgoyal01 I know that the block store is already going to be replicated or have some sort of redundancy scheme, but there are still issues of failure domains in terms of storage control plane, drivers, etc that can happen. 
There are several kinds of workloads (e.g. Kafka, clustered databases, etc) that replicate data amongst multiple nodes in a cluster to create a higher level of availability.  Having compute availability zones helps in that the DB instances can be divided so that they do not share fault domains at the compute level, but that does not address the issue of if there are shared fault scenarios in the storage tier.  Storage availability zones is one way to identify pools of storage that should not fail simultaneously, thus presenting a useful capability for these replicated/clustered data services.

pgoyal01: @kedmison I know at the RM we need to abstract away from implementation details but here is the scenario with OpenStack. 
1. In OpenStack unlike some public clouds, AZs do not isolate subnets.
1. An AZ is a logical partition of either compute, block storage or networking services where each of the services implements AZs differently.
1. A compute node can be in multiple host aggregates but in only one compute AZ.
1. Block Storage (Cinder) AZs may be configured for different backends. Many compute AZs can map to a Cinder AZ.
1. Neutron AZs (relatively new) allows grouping of network nodes that run networking services such as DHCP, L3, and others.

Compute AZs are more commonly used and, typically, are the subject when discussing AZs in the OpenStack context; unless referring to Cinder or Neutron AZs.

The failure domain for a typical AZ (a host aggregate that is visible to the tenants) depends upon implementations: it can be one or compute nodes or a rack or ...

kedmison: @pgoyal01 thanks for the discussion.  I'll take some of your points, one-by-one.
 
> 2. An AZ is a logical partition of either compute, block storage or networking services where each of the services implements AZs differently.

Openstack breaks it down this way, yes.  But is that the most useful way to present it to the users?  Perhaps we should be using Openstack's constructs to create a new presentation instead of passing them through directly.    Similar to how a 'container' isn't a first-class thing in the linux kernel, but instead an abstract entity composed of several cgroup implementations applied in a co-ordinated fashion.  An AZ in the datacenter space should be made up of compute separation (nova AZ) + storage separation (cinder AZ) + network isolation (independent physical network infrastructure as far up the leaf/spine/core-router tree as is possible).

> 3. A compute node can be in multiple host aggregates but in only one compute AZ.

Agree; no problem here.

> 4. Block Storage (Cinder) AZs may be configured for different backends. Many compute AZs can map to a Cinder AZ.

Here I have an observation: With a single Cinder AZ, there is likely a single Cinder back-end.  That can fail (not just a single-disk failure or single-storage-node failure but a total failure of the storage front-end.  This can cause all workloads using that storage to fail.  A properly designed workload could survive that event, but only if it was aware of and could leverage separate Cinder back-ends through Cinder AZ separation.  I think this is a useful addition to the availability model of an NFVI to have independent storage backends wherever there are independent compute AZs.

> The failure domain for a typical AZ (a host aggregate that is visible to the tenants) depends upon implementations: it can be one or compute nodes or a rack or ...

Agreed.  And, I understand that different operators may choose to implement failure domain separation strategies (e.g. different rooms w/ different HVAC services, and/or different utility power feeds, etc etc.)   But we should seriously look at ensuring that an AZ at least covers the compute domain and the storage domain.

pgoyal01: @kedmison we need separate requirements for the compute and storage domains including what are the characteristics (requirements) for a "failure domain" and then have the operators review and approve these requirements.

BTW these requirements are likely to impact both capital and operational expense.

ASawwaf:
@kedmison @pgoyal01 , thanks for the great discussion, What @kedmison try to highlight ( AWS like approach, which is one of our target ) which from my POV is into DR ( Disaster recovery )concept which I believe is an operator decision as in Telco most probably all the resiliency and redundancy is achieved form the VNF side itself due to the nature of VNF 

the bottom line here, we need to shed light on the AZ in RM with the above @pgoyal01 modified definition "with the operator implementation strategy

doc/ref_model/chapters/chapter03.md
kedmison: > the bottom line here, we need to shed light on the AZ in RM with the above @pgoyal01 modified definition "with the operator implementation strategy

@asawwaf @pgoyal01 this is a great discussion, and is one where we are really getting to some of the fine points and hard balances at the VNF<->NFVI interface.

@pgoyal01 Can I propose an option here?  That instead of having a single count of AZ in Table 4.1 in https://github.com/cntt-n/CNTT/pull/1250/files#diff-ea572522f5dd14982c918049c9c7dfa1R80 we update this to be a count of compute AZ and then add a count of storage AZ?  
the intent with this is to offer the flexibility to, for larger data centers, provide the full range of availability-zone capabilities while enabling smaller topologies, heading out towards the edge, to offer a reduced set.  This would also let operators make their own decisions while still providing clear information up-front to VNF vendors about what capabilities are present and can be leveraged.

rabiabdel:
@kedmison can this be added to next RM agenda item to get consensus on.

doc/ref_model/chapters/chapter03.md
pgoyal01: @kedmison Yes separating the different AZ counts would help. I still have a mental block.  If the operator is free to define an AZ on a non-failure domain (e.g., rack -- no independent power) then "Workloads can leverage availability zones" for redundancy (and for clusters) but not "avoid sharing failure domains for a workload, and thus increase its fault-tolerance." Can the sentences be changed to reflect that.

kedmison: I've been trying to leave it as an operator/deployment decision as to which specific failure domains are separated for compute AZs and storage AZs, as it's really a function of the size of deployment.
For a single rack deployment, one compute AZ and one storage AZ may be all that is provided.
For a multi-rack deployment, the AZs may not share TOR switches, but or PDUs, but still share building circuits, utility power, HVAC, buildings etc.
For major datacenters, the AZs may extend to having separate utility power feeds, separate buildings, etc.  

Simply providing the AZs can help with redundancy (which helps with fault tolerance), but by careful partitioning of failure domains along AZ boundaries, the NFVI deployment can help increase fault tolerance even further.

doc/ref_model/chapters/chapter04.md
```diff
@@ -76,6 +76,9 @@ This section describes a set of explicit NFVI capabilities and performance measu
 | e.nfvi.cap.003 | Total per-instance (ephemeral) storage | GB | Max storage in GB that can be assigned to a single VNFC by NFVI |
 | e.nfvi.cap.004 | # Connection points | number | Max number of connection points that can be assigned to a single VNFC by NFVI |
 | e.nfvi.cap.005 | Total external (persistent) storage | GB | Max storage in GB that can be attached / mounted to VNFC by NFVI |
+| e.nfvi.cap.017 | Total Availability Zones | number | The number of compute availability zones supported by NFVI |
```
pgoyal01: Total Availability Zones in all of an Operator's Data Centers? What would be the value of knowing that? The number of AZs may be different in each of the OpenStack Regions, for example.

Same comment for storage AZs.

walterkozlowski: I agree with @pgoyal01 comments: it is not clear what we want to achieve here.  In OpenStack AZs are exposed to end users who can select a particular AZ in a particular Region for their VNFC.  This might be an AZ for staging or production or other logical grouping of compute/storage nodes. Hence, this will be dependent  on operator's deployment and LCM architecture.  So the important factors are not the total number (as @pgoyal01 noted) not even number per OS Region, but the purpose  of each AZ, its capability, its capacity and relation to the Region's host aggregates, etc.  What would be the scenario that we could effectively use the knowledge about the number of AZs?

tomkivlin: > What would be the scenario that we could effectively use the knowledge about the number of AZs?

In particular, what is the desired outcome of any conformance tests that are defined to test this and why are they important in terms of what the CNTT is driving (abstraction, standardisation, etc.)?

kedmison: The desired outcome is a partitioning of compute nodes (and ideally storage resources) that workloads can leverage in a more optimal way.

Without availability zones, workloads are forced to use anti-affinity.  This is fine when there are 2 or even 4 VMs that must not share a failure domain, as anti-affinity would need only 2 or 4 different physical compute hosts. It is not ok when there are 50 or 60 VMs that must not share a failure domain, triggering workload placement on 50 or 60 different compute hosts and, depending on the size of the cluster, exceeding the actual number of compute hosts deployed.

This is a common sort of occurrence with horizontally-scalable, replicated databases.  One wants to ensure that the data is always available by making sure that the data and its replicated copy are not vulnerable to the same failure.  This allows the workload to partition the primary copy of data in one 'az' and the replica in another 'az', but then to not necessarily need to care about whether the DB VMs in the first AZ are sharing underlying compute, since there is no single failure that can take out both the primary copy and its replica.  
The net is that this allows a workload to go from needing 50-60 different compute hosts, down to perhaps 12-20, as the DB VMs in one AZ can now be bin-packed on the same compute hosts.

Having a documented measure, if not a guarantee, of a minimum number of availability zones allows workloads to design in this optimization.  From that standpoint, I see documenting the number, and publishing it, as an important, indirect part of infrastructure abstraction.

This count of AZs is trying to strike a balance between a strict requirement of >=2 AZs per region, which received some negative comments, vs. the need to at least publish and know about the AZ numbers.  

@karinesevilla commented in a different thread that the AZ names are exposed to VNFs.  That's true, and likely should be added as a requirement.

pgoyal01: @kedmison In OpenStack there is an AZ list command (openstack availability zone list -- compute  (for compute AZs) that lists the available AZs. But please remember AZs in OSTK are host aggregates that are visible to tenants (projects) and where a host can be a member of one AZ only; the latter is different from host aggregates. AZs may not necessarily be failure domains as an AZ in OSTK can be restricted for use of a particular tenant (project). Knowledge of a number of AZs where AZs are restricted for use by certain projects is unlikely to be helpful.

kedmison: @pgoyal01 A count seems to be the wrong way to express this, as there is a lot of differentiation in how you are saying AZs are used.  Can we express this instead as a list of AZs that do separate failure domains? (and thus are appropriate for use in the use case discussed further up the comment thread)?
This would not preclude the presence of other AZs defined for other purposes (like private zones of the data center, or isolating classes of workload)

pgoyal01: @kedmison Yes -- that would work. It should be changed to a VIM requirement instead of an NFVI capability. Something like, VIM must expose the list of availability zones and their details.

kedmison: @pgoyal01 that's great; thank you.

doc/ref_model/chapters/chapter04.md
```diff
@@ -125,7 +126,7 @@ Monitoring capabilities are used for the passive observation of workload-specifi
 
 | Ref | NFVI Capability | Unit | Definition/Notes |
 |--------------------|---------------------------|--------|----------------------------------------------------|
-| e.nfvi.cap.013 | Monitoring of L2-7 data | Yes/No | Ability to monitor L2-L7 data from workoad |
+| e.nfvi.cap.017 | Monitoring of L2-7 data | Yes/No | Ability to monitor L2-L7 data from workload |
```
CsatariGergely, line 129: Please merge with master and use nfvi-less attribute names, like `e.cap.017`.

kedmison: fixed.

rabi-abdel:
@pgoyal01 @karinesevilla @CsatariGergely can you guys re-review please

doc/ref_model/chapters/chapter03.md
```diff
@@ -128,7 +129,20 @@ _**Example**: a virtual compute descriptor as defined in TOSCA Simple Profile fo
 | `subnet`       | network address of the subnet                                            |
 | `acceleration` | key/value pairs for selection of the appropriate acceleration technology |
 
-<p align="center"><b>Table 3-4:</b> Attributes of network resources.</p>
+<p align="center"><b>Table 3-4:</b> Attributes of network resources</p>
+
+<a name="3.2.5"></a>
+### 3.2.5 Availability Zone
+An Availability Zone is a logical pool of physical resources (e.g. compute, block storage, network).  These logical pools segment the physical resources of a cloud based on factors chosen by the cloud operator. The cloud operator may create availability zones based on location (rack, datacenter), or indirect failure domain dependencies like power sources.  Workloads can leverage availability zones to avoid sharing failure domains for a workload, and thus increase its fault-tolerance.
```
pgoyal01, line 136: "Workloads can leverage availability zones to avoid sharing failure domains for a workload, and thus increase its fault-tolerance."
In the previous sentence AZs are defined as either "location" or "failure domains" and hence I suggest that we chnage to, say,:
"Workloads can leverage availability zones to utilise multiple locations or avoid sharing failure domains for a workload, and thus increase its fault-tolerance."

doc/ref_model/chapters/chapter04.md
```diff
@@ -235,13 +235,15 @@ Table 4-9 shows capabilities related to resources allocation
...
+| Ref       | Cloud Infrastructure management Capability | Unit            | Definition/Notes                                                 |
+|-----------|--------------------------------------------|-----------------|------------------------------------------------------------------|
+| e.man.001 | Virtual Compute allocation                 | Yes/No          | Capability to allocate virtual compute resources to a workload   |
+| e.man.002 | Virtual Storage allocation                 | Yes/No          | Capability to allocate virtual storage resources to a workload    |
+| e.man.003 | Virtual Networking resources allocation    | Yes/No          | Capability to allocate virtual networking resources to a workload |
+| e.man.004 | Multi-tenant isolation                     | Yes/No          | Capability to isolate resources between tenants                  |
+| e.man.005 | Images management                          | Yes/No          | Capability to manage workload software images                    |
+| e.man.010 | Compute Availability Zones                 | list of strings | The names of each Compute Availability Zone that was defined to separate failure domains |
```
pgoyal01, line 245: The definition/notes should reflect taht AZs are not only failure domains. Same comment for e.man.011

doc/ref_model/chapters/chapter03.md
ASawwaf: @pgoyal01 @kedmison, Utilize multiple location , introduce the concept of HA or DR which on of the target of AZ 
So I will go with @pgoyal01 modification

doc/ref_model/chapters/chapter04.md
kedmison: This is intended to be a list fo the AZs that are failure domain separations; not a list of all AZs.  So, it would be possible to create AZs for other reasons, but if they are not separated from a failure domain standpoint, they wouldn't be listed here.

pgoyal01: @kedmison 
1. How would one obtain a list of AZs that are defined by separate failure domains? OpenStack can provide you a list of AZs. 

1. My concern here is that we not set expectations that may not be supported by operators. For example, when we say separate power supply (failure domain) is that from separate UPSs or ....?

kedmison: I was expecting this would be a documented list rather than a list pulled from an API, so that when using this cloud infrastructure, VNF vendors could understand, enumerate, and depend on the failure-domain based zones.

The list is not actually prescriptive in the types of failure domain separation either, as it is expected that operators may not be able to resolve to a common definition, nor would that definition be consistent in all locations.  I think it's best left to an operator to decide what failure types need to be considered when defining failure domains, and perhaps later revisions of this document may be able to somewhat normalize that definition over time.

rabi-abdel:
@pgoyal01 can you re-review please?

pgoyal01:
@rabi-abdel There is a suggested chnaged from @ASawwaf and me that has not been addressed.

rabi-abdel:
@kedmison can you please review this PR

doc/ref_model/chapters/chapter03.md
kedmison: good point. applied.

PR #1253, opened 2020-03-05
[RA1][RI1][RC1]Add oauth1 as optional Keystone features

collivier:
It was not listed in Keystone which defacto sets it as optional.
It may be considered as mandatory in a second step.

It should be noted that it's not working in CNTT RI and there is no
true plan to update CNTT RI from the time writing.

RA1 and RC are updated in one change according to the traceability in
place.

### DIFF
--- a/doc/ref_arch/openstack/chapters/chapter05.md
+++ b/doc/ref_arch/openstack/chapters/chapter05.md
@@ -45,6 +45,7 @@ For the purpose of compliance tests, this chapter also identifies the set of the
 | application_credentials | X             |
 | external_idp            |               |
 | federation              |               |
+| oauth1                  |               |
 | project_tags            | X             |
 | security_compliance     | X             |
 | trust                   | X             |
--- a/doc/ref_cert/lfn/chapters/chapter03.md
+++ b/doc/ref_cert/lfn/chapters/chapter03.md
@@ -47,7 +47,7 @@ The following is a Requirements Traceability Matrix (RTM) mapping Test Case, and
 
 The RTM contains RM config (i.e. .conf) requirements listed “per profile”, followed by RA-1 requirements.  Requirements fall into 8 domains: general(gen), infrastructure(inf), VIM(vim), Interface & API(int), Tenants(tnt), LCM(lcm), Assurance(asr), Security(sec).
 
-For detailed information on RM & RA-1 NFVI and VNF requirements, please refer to [RI-1 Chapter 3](https://github.com/cntt-n/CNTT/blob/master/doc/ref_impl/cntt-ri/chapters/chapter03.md). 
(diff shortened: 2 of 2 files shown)


collivier:
@fuqiao123 @rabi-abdel @mf4716 About this Keystone features and all the JIRA tickets opened in RI, I would appreciate a clear plan of the updates/fixes in CNTT RI. Most of the bugs bugs are well known for a while and we have only fixed the catalog (Nova endpoints, useless tcp port, etc.). RC is working well and we can't block the Conformance due to RI. What's the plan here? Do we have to consider an alternative?  Governance topic?

collivier:
@fuqiao123 @mf4716 (OPNFV liaisons) From an OPNFV statepoint, it's also true as the test results are not well taken into account as input of the OPNFV installer which differs from the OPNFV model.

fuqiao123:
> @fuqiao123 @rabi-abdel @mf4716 About this Keystone feature and all the JIRA tickets opened in RI, I would appreciate a clear plan of the updates/fixes in CNTT RI. Most of the bugs are well known for a while and we have only fixed the catalog (Nova endpoints and 80/tcp port in endpoint). RC is working well and we can't continue blocking the Conformance due to RI. What's the plan here? Do we have to consider an alternative? Governance topic?

Agree. What I can see is the airship upstream seems either do not know what's happening here, or don't care? At least no further respond from them for the functest results. Bring this to the governance probably could be a solution. Since I think the officers from AT&T probably could help to improve the influence on the airship project.

rabiabdel:
> > @fuqiao123 @rabi-abdel @mf4716 About this Keystone feature and all the JIRA tickets opened in RI, I would appreciate a clear plan of the updates/fixes in CNTT RI. Most of the bugs are well known for a while and we have only fixed the catalog (Nova endpoints and 80/tcp port in endpoint). RC is working well and we can't continue blocking the Conformance due to RI. What's the plan here? Do we have to consider an alternative? Governance topic?
> 
> Agree. What I can see is the airship upstream seems either do not know what's happening here, or don't care? At least no further respond from them for the functest results. Bring this to the governance probably could be a solution. Since I think the officers from AT&T probably could help to improve the influence on the airship project.

Yes good point @collivier @fuqiao123 . let us discuss this in our sync call today. i hope @scottsteinbrueck can help facilitate some of this with Airship team.

mf4716:
Approved.  

Please refer to JIRA AIRSHIP-31 for details of the discussion about OAUTH1.  

For clarify purposes, from the JIRA discussion, "...It [have OAUTH1] is not an issue in the Airship installer in opnfv... It is a known missing feature in OpenStack Helm Keystone... OSH project welcomes anyone wants to contribute the work, but oauth1 is low on OSH community's priority list because low percentage of usage."  

Also, The RA-1 WS lead confirmed OAUTH/OAUTH1 is not a mandatory requirement for RA-1, or RM.

Therefore, OAUTH/OAUTH1 is not a bug with the Airship/OSH installer.  This is an optional extension, which if needed by the community at large, would be handled via OSH/Airship community forums as an enhancement.

PR #1265, opened 2020-03-06
[RM] Exceptions

rabi-abdel:
Fixes https://github.com/cntt-n/CNTT/issues/750

### DIFF
--- a/doc/ref_model/chapters/appendix-a.md
+++ b/doc/ref_model/chapters/appendix-a.md
@@ -5,7 +5,7 @@
 ## Table of Contents
 * [A.1 Goals](#A.1)
 * [A.2 Intro and Terminology](#A.2)
-* [A.3 VNF Evolution Phases](#A.3)
+* [A.3 Exception List](#A.3)
 * [A.4 Links](#A.3)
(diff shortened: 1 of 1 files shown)


pgoyal01:
@rabi-abdel I personally don't think that exceptions should be identified within RM or the RA's etc. I think there should be a separate consolidated document for both technology and version exceptions.

ASawwaf:
@pgoyal01 @rabi-abdel , I am in favor off to have separate Annex, amendment collecting all expectation that we need to adopt during our roadmap as by time the exception should be resolved to avoid keep changes Rx

rabiabdel:
@pgoyal01 @ASawwaf i would assume the opposite. if any exception is removed from any of the documents, that would be as a result of updating the document to support an alternative technology that gets rid off the exception so they will be updated together and stay on Sync. having those exception in the annex of RM (as to communicate this directly to VNF vendors) and in the annex of RA (as to communicate this directly to VI vendors) is the straight forward way of doing it.

ASawwaf:
@rabi-abdel in both cases we will go to edit / remove it  :) 
lets see

PR #1268, opened 2020-03-06
[RA2 appendix]: Following RM SR-IOV exception

### LINKED ISSUE #981: [RA2] Ch3 - track SR-IOV exception for RA2
From deep dive in Prague:
3.2.1 Raise an issue to track SR-IOV exception for RA-2

CsatariGergely:
Closes #981 

Should be merged only after #1265 was merged

### DIFF
--- a/doc/ref_arch/kubernetes/chapters/appendix-a.md
+++ b/doc/ref_arch/kubernetes/chapters/appendix-a.md
@@ -1,18 +1,26 @@
 [<< Back](../../kubernetes)
-# Appendix A - Guidance for VNF-to-CNF Transition
+# Appendix A - Guidelines For VNF Vendors
 <p align="right"><img src="../figures/bogo_ifo.png" alt="scope" title="Scope" width="35%"/></p>
 
 ## Table of Contents
-* [A.1 TBC](#A.1)
-* [A.2 TBC](#A.2)
+
+* [A.1 Exception List](#A.1)
+* [A.2 Guidance for VNF-to-CNF Transition](#A.2)
 * [A.3 TBC](#A.3)
 * [A.4 TBC](#A.3)
 
 <a name="A.1"></a>
-## A.1 TBC
+## A.1 Exception List
+
+| Ref         | Type       | Name               | RM Exception | Details |
+|-------------|------------|--------------------|--------------|---------|
+| ra2.exc.001 | Technology | SR-IOV             | rm.exc.001   | This exception allows workloads (VNF/CNF) to use SR-IOV Pass-Through. RA2 architectures allows plugins such as SR-IOV CNI to be used. |
+<!--
+| ra2.exc.002 |            |                    |              |         |
+-->
 
 <a name="A.2"></a>
-## A.2 TBC
+## A.2 Guidance for VNF-to-CNF Transition
 
 <a name="A.3"></a>
 ## A.3 TBC
--- a/doc/ref_arch/kubernetes/chapters/appendix-a.md
+++ b/doc/ref_arch/kubernetes/chapters/appendix-a.md
@@ -14,7 +14,7 @@
 
 | Ref         | Type       | Name               | RM Exception | Details |
 |-------------|------------|--------------------|--------------|---------|
-| ra2.exc.001 | Technology | SR-IOV             | rm.exc.001   | This exception allows workloads (VNF/CNF) to use SR-IOV Pass-Through. RA2 architectures allows plugins such as SR-IOV CNI to be used. |
+| ra2.exc.001 | Technology | SR-IOV             | rm.exc.001   | This exception allows workloads (VNF/CNF) to use SR-IOV. RA2 architectures allows plugins such as SR-IOV CNI and SR-IOV related Device Plugins, like [sriov-network-device-plugin](https://github.com/intel/sriov-network-device-plugin) to be used. |
 <!--
 | ra2.exc.002 |            |                    |              |         |
 -->
--- a/doc/ref_arch/kubernetes/chapters/appendix-a.md
+++ b/doc/ref_arch/kubernetes/chapters/appendix-a.md
@@ -4,15 +4,21 @@
 
 ## Table of Contents
 
-* [A.1 Guidance for VNF-to-CNF Transition](#A.1)
-* [A.2 TBC](#A.2)
+* [A.1 Exception List](#A.1)
+* [A.2 Guidance for VNF-to-CNF Transition](#A.2)
 * [A.3 TBC](#A.3)
+* [A.4 TBC](#A.4)
 
 <a name="A.1"></a>
-## A.1 Guidance for VNF-to-CNF Transition
+## A.1 Exception List
+
+The exception list is collected in [Appendix A - Guidelines for VNF Vendors of the Reference Model](../../../ref_model/chapters/appendix-a.md#A.3).
 
 <a name="A.2"></a>
-## A.2 TBC
+## A.2 Guidance for VNF-to-CNF Transition
 
 <a name="A.3"></a>
 ## A.3 TBC
+
+<a name="A.4"></a>
+## A.4 TBC
--- a/doc/ref_arch/kubernetes/chapters/appendix-a.md
+++ b/doc/ref_arch/kubernetes/chapters/appendix-a.md
@@ -1,29 +1,18 @@
 [<< Back](../../kubernetes)
-# Appendix A - Guidelines For VNF Vendors
+# Appendix A - Guidance For VNF Vendors
 <p align="right"><img src="../figures/bogo_ifo.png" alt="scope" title="Scope" width="35%"/></p>
 
 ## Table of Contents
 
-* [A.1 Exception List](#A.1)
-* [A.2 Guidance for VNF-to-CNF Transition](#A.2)
+* [A.1 Guidance for VNF-to-CNF Transition](#A.1)
+* [A.2 TBC](#A.2)
 * [A.3 TBC](#A.3)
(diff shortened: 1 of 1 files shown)


doc/ref_arch/kubernetes/chapters/appendix-a.md
```diff
@@ -1,18 +1,26 @@
 [<< Back](../../kubernetes)
-# Appendix A - Guidance for VNF-to-CNF Transition
+# Appendix A - Guidelines For VNF Vendors
```
tomkivlin: The heading in the doc is "Guidance" not "Guidelines", suggest changing to "Guidance"

doc/ref_arch/kubernetes/chapters/appendix-a.md
```diff
@@ -1,18 +1,26 @@
...
 * [A.3 TBC](#A.3)
 * [A.4 TBC](#A.3)
 
 <a name="A.1"></a>
-## A.1 TBC
+## A.1 Exception List
+
+| Ref         | Type       | Name               | RM Exception | Details |
+|-------------|------------|--------------------|--------------|---------|
+| ra2.exc.001 | Technology | SR-IOV             | rm.exc.001   | This exception allows workloads (VNF/CNF) to use SR-IOV Pass-Through. RA2 architectures allows plugins such as SR-IOV CNI to be used. |
```
tomkivlin: The exception in RM Appendix A is:
>This exception allows workloads (VNF/CNF) to rely on SR-IOV technology.

It doesn't include the Pass-Through part. My understanding was that the PCI-Passthrough part was the issue with SR-IOV as it ties a workload to a particular VF on a particular PCIe card. Therefore, should we remove that part from here?

rabiabdel:
@CsatariGergely is it possible to remove PCIe-Pass through details to avoid confusion and get this PR approved and merged?

doc/ref_arch/kubernetes/chapters/appendix-a.md
CsatariGergely: Humm, interesting. I believe I've copied the text from RM Appendix A. 
In case of containers SR-IOV means PCI-Passthrough as the VF-s are mounted to the containers as PCI devices by the device plugin. I can remove the PCI-Passthrough part and add, refer to the usage of sriov-network-device-plugin. Is that OK?

tomkivlin: Yes I would prefer that.  I agree that, for containers, the tying of a particular workload (container) to a particular PCIe card (hardware) isn't necessarily a problem as we don't move containers in the same way as we do VMs - they are re-scheduled and so the Kubernetes device plugin is the correct method.

CsatariGergely: Okey, done in 36a6b7f.

doc/ref_arch/kubernetes/chapters/appendix-a.md
tomkivlin: @CsatariGergely - still need this typo corrected, then I can approve 👍

rabiabdel:
@tomkivlin once this approved by you, this will get merged.

doc/ref_arch/kubernetes/chapters/appendix-a.md
```diff
@@ -1,18 +1,26 @@
...
 * [A.3 TBC](#A.3)
 * [A.4 TBC](#A.3)
 
 <a name="A.1"></a>
-## A.1 TBC
+## A.1 Exception List
+
+| Ref         | Type       | Name               | RM Exception | Details |
+|-------------|------------|--------------------|--------------|---------|
+| ra2.exc.001 | Technology | SR-IOV             | rm.exc.001   | This exception allows workloads (VNF/CNF) to use SR-IOV. RA2 architectures allows plugins such as SR-IOV CNI and SR-IOV related Device Plugins, like [sriov-network-device-plugin](https://github.com/intel/sriov-network-device-plugin) to be used. |
```
markshostak: Hi @CsatariGergely,
A) Can you clarify why this exception required? What type of kernel driver bindings do you envision? I'm asking because I didn't see any configurations of the sriov-network-device-plugin that violated any CNTT Principles, and you explicitly removed references to PCI-PT (which would require an exception), so you may simply be able to permanently codify your design in RA-2, without the need to draw out and maintain a formal exception. There's a related discussion in PR #1336 

B) If we're creating a new Exception, the intent is to record the rationale and need for the exception. Also, the exception should identify when the exception will be retired, and what the transition plan looks like. I'd suggest you use the draft template in the exception framework in the RM, as you can fill in the blanks and won't miss anything.

C) If you're replicating an existing exception, can it be incorporated by reference, so the 2 copies don't have to be kept in sync?

Thanks,
-Mark

CsatariGergely: A) sriov-network-device-plugin assigns PCI devices to the containers and the container needs to understand how to consume them. I've removed the PCI-PT references to address the comments from @rabi-abdel and @walterkozlowski and added the sriov-network-device-plugin to "soften" the PCI-PT a bit (bit in reality they are kinda the same). 
B) Okey, if we need a new exception I will do that. 
C) The original intent was to just copy the RM exception, but I agree that a) we should make the RM exceptions linkable and b) only refer to them in the RA exception list, or just do not have an RA exception list. 

For me the best solution would be to merge #1354 and just refer it here or remove the RA2 exception list completely.

@walterkozlowski , @rabi-abdel ?

markshostak: @CsatariGergely Hi Gergely, 
>
> A) sriov-network-device-plugin assigns PCI devices to the containers and the container needs to understand how to consume them. 

Agreed, containers do need to understand how to consume PCI-coupled resources, which describes many resources in a server. However, as you know better than I, host-based kernel drivers expose them as an abstraction, so they are normalized by the time the container deals with them. Based on the sriov-network-device-plugin docs, the actual h/w drivers are instantiated in the host, where they are bound to the actual PCI devices. Here's an excerpt of the workflow from their docs, [here](https://github.com/intel/sriov-network-device-plugin):

- Load device's (Physical function if it is SRIOV capable) kernel module and bind the driver to the PF
- Create required Virtual functions
- Bind all VF with right drivers

I interpret the context of those items as being in the host. So, if their docs are accurate, and I'm reading them correctly, sriov-network-device-plugin gives us more value than a simple PCI-PT transport.

>
>  "soften" the PCI-PT a bit (bit in reality they are kinda the same).

Again, based on what I see in their docs, the only thing I see the two have in common is they allocate (dedicate) a VF (PCI device) to a process. We do that all over the place, notably in virtual network interface assignment, inode assignment in the FS, CPU pinning and RAM allocation, and doing so does not violate our cloud-native principles. Whereas, PCI-PT doesn't just dedicate hardware resources, it exposes their underlying makeup, which then requires s/w custom written to communicate with the resources (and an exception in CNTT). In sriov-network-device-plugin, this custom s/w is instantiated in the host's kernel, which is consistent w/ CNTT architecture and principles. What is exposed to the container is an abstraction over a socket, so I'd suggest PCI-PT and sriov-network-device-plugin are intrinsically different (and if it's not, we should work with them to fix it).

If I'm missing something obvious, please help me with it.

rabiabdel: Let us remove it completely. i suggest. i.e. let us not have the exception in RA2, for now.

markshostak: @rabiabdel @CsatariGergely I'm fine w/ that, but this has implications for #1336

doc/ref_arch/kubernetes/chapters/appendix-a.md
CsatariGergely: Done

doc/ref_arch/kubernetes/chapters/appendix-a.md
CsatariGergely: I think whatever the conlusion in #1336 that should be concluded to the [Exception List of the RM](https://github.com/cntt-n/CNTT/blob/e13c1b9ce7e55df0fbd09a206bc5172c5e077bab/doc/ref_model/chapters/appendix-a.md#A.3) and this document should only refer to it.

rabiabdel:
@markshostak can you re-review/approve?

doc/ref_arch/kubernetes/chapters/appendix-a.md
```diff
@@ -1,18 +1,21 @@
 [<< Back](../../kubernetes)
-# Appendix A - Guidance for VNF-to-CNF Transition
+# Appendix A - Guidance For VNF Vendors
```
pgoyal01: Should this be "Guidance for CNF Vendors"; RA-1 has a  "Guidance for VNF Vendors"

deebhatia: @pgoyal01 It includes both because we also have a section for `Guidance for VNF-to-CNF Transition`

pgoyal01: @deebhatia I saw that but I was referring to the Appendix Title -- currently it refer only to VNF; we have spent a lot of energy changing from VNF and NFV to include explicit references to CNF.

rabi-abdel:
@tomkivlin once this approved by yourself, it can be merged.

doc/ref_arch/kubernetes/chapters/appendix-a.md
CsatariGergely: I've changed the title to "Appendix A - Guidance For application Vendors"

PR #1280, opened 2020-03-09
[RM Ch04]: Kubernetes related clarification to Table 4-17

CsatariGergely:
Table 4-17 describes the size of different comput flavors, but in Kubernetes the
CPU and memory size is defined with requests and limits. This change clarifies that
these numbers define resource requests in case of Kubernetes.

### DIFF
--- a/doc/ref_model/chapters/chapter04.md
+++ b/doc/ref_model/chapters/chapter04.md
@@ -387,19 +387,20 @@ The intent of the following Flavours list is to be comprehensive and yet effecti
 
 >_*Note:*_ Customised (Parameterized) Flavours can be used in concession by operators and, if needed, are created using TOSCA, HEAT templates, and/or VIM APIs.
 
-.conf |vCPU ("c") |RAM ("r") |Local Disk ("d") | Management Interface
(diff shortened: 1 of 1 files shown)


tomkivlin:
Genuine question: if the purpose of these t-shirt sizes is to aid the overall capacity management/dimensioning of a cloud infrastructure, would it not make more sense to use the limit values, rather than requests?

kedmison:
> Genuine question: if the purpose of these t-shirt sizes is to aid the overall capacity management/dimensioning of a cloud infrastructure, would it not make more sense to use the limit values, rather than requests?

@tomkivlin this is a good point.   I had thought they were indicating that, but the term 'resource requests limits' is perhaps not precise enough.  Would resource limits be sufficient?

tomkivlin:
@kedmison, to be honest I think both are useful as it helps with understanding normal run rates and peaks. In addition the combination of request and limits is used to determine the QoS class that the scheduler gives a container so perhaps we're needed both?

CsatariGergely:
'resource requests limits' just indicated, that I was not quite sure which one :). Now I've clarified and added some explanation about the different resource types.

PR #1308, opened 2020-03-14
[RA-1 Ch02]Fix '2.3.8 Security Requirements' table show

zhur0ng:
the extra |----| make the table can not show normally, this patch remove this.


doc/ref_arch/openstack/chapters/chapter02.md
```diff
@@ -188,7 +188,6 @@ Traceability to Reference Model.
 <!-- | `req.sec.ntw.01` | Networking | The Architecture **must** have the underlay network include strong access controls that comply with the applicable security standard (national, regional), for example ISO27001. | |
```
rabiabdel: may i suggest moving those to the bottom of the table (otherwise they wont be rendered correctly)

PR #1310, opened 2020-03-16
[RA-1 Ch04] update of 4.2.2.3 and 4.2.2.5

### LINKED ISSUE #1153: [RA1 Ch04] Section 4.2: content to be developed
Content needed for:
- introduction
- 4.2.2.3: Network nodes, more details
- 4.2.2.5: Compute nodes: develop the specificities for Basic and Network Intensive profiles

samuelhellec:
fix #1153

### DIFF
--- a/doc/ref_arch/openstack/chapters/chapter04.md
+++ b/doc/ref_arch/openstack/chapters/chapter04.md
@@ -102,7 +102,7 @@ For OpenStack control nodes we use the BIOS parameters for the basic profile def
          - 1 NIC for Other Networks (PXE, Mngt ...)
 -	Sizing rules
     - Scale out of network node is not easy 
-    - DVR can be an option for large deployment (see more detail
+    - DVR can be an option for large deployment (see more detail in chapter 4.3.1.5 - Neutron)
 
 #### 4.2.2.4. Storage nodes
 -	BIOS requirements
--- a/doc/ref_arch/openstack/chapters/chapter04.md
+++ b/doc/ref_arch/openstack/chapters/chapter04.md
@@ -94,8 +94,15 @@ For OpenStack control nodes we use the BIOS parameters for the basic profile def
 | …|  
  
 -	How many nodes to meet SLA
+    - Minimum 2 nodes for high availibility using VRRP.
 -	HW specifications
+    - 3 NICs card are needed if we want to isolate the different flows :
+         - 1 NIC for Tenant Network
+         - 1 NIC for External Network
+         - 1 NIC for Other Networks (PXE, Mngt ...)
 -	Sizing rules
+    - Scale out of network node is not easy 
+    - DVR can be an option for large deployment (see more detail
 
 #### 4.2.2.4. Storage nodes
 -	BIOS requirements
@@ -120,6 +127,7 @@ For OpenStack control nodes we use the BIOS parameters for the basic profile def
 |---------------|-----------|------------------|
 | Boot disks | RAID 1 | RAID 1 | 
 | CPU reservation for host (kernel) | 1 core per Numa | 1 core per Numa | 
+| CPU Pinning | No | Yes | 
 | <to be filled if needed> |  |  | 
 | … |  |  | <!--- | --->
 
@@ -133,6 +141,7 @@ CPU reservation for host: 1 core per NUMA
     - minimum: two nodes per profile
 -	HW specifications
     -	Boot disks are dedicated with Flash technology disks
+    - Additional 4 cores overhead per numa in case of DPDK usage for Network Intensive nodes
 -	Sizing rules
 
 | Number of CPU sockets| s | 
--- a/doc/ref_arch/openstack/chapters/chapter04.md
+++ b/doc/ref_arch/openstack/chapters/chapter04.md
@@ -85,6 +85,9 @@ For OpenStack control nodes we use the BIOS parameters for the basic profile def
 
 
 #### 4.2.2.3. Network nodes
+
+Networks nodes are mainly used for L3 traffic management for overlay tenant network (see more detail in section 4.3.1.5 Neutron)
+
 -	BIOS requirements 
 
 | BIOS/boot Parameter | Value |
--- a/doc/ref_arch/openstack/chapters/chapter04.md
+++ b/doc/ref_arch/openstack/chapters/chapter04.md
@@ -141,7 +141,7 @@ CPU reservation for host: 1 core per NUMA
     - minimum: two nodes per profile
 -	HW specifications
     -	Boot disks are dedicated with Flash technology disks
-    - Additional 4 cores overhead per numa in case of DPDK usage for Network Intensive nodes
+    - In case of DPDK usage, extra cores must be reserved. The number of core to book for this overhead depend on the CPU frequency (Mpps available per core) and the number of Mpps to be delivered by node.
 -	Sizing rules
 
 | Number of CPU sockets| s | 
--- a/doc/ref_arch/openstack/chapters/chapter04.md
+++ b/doc/ref_arch/openstack/chapters/chapter04.md
@@ -141,7 +141,8 @@ CPU reservation for host: 1 core per NUMA
     - minimum: two nodes per profile
 -	HW specifications
     -	Boot disks are dedicated with Flash technology disks
-    - In case of DPDK usage, extra cores must be reserved. The number of core to book for this overhead depend on the CPU frequency (Mpps available per core) and the number of Mpps to be delivered by node.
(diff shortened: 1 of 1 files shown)


doc/ref_arch/openstack/chapters/chapter04.md
```diff
@@ -133,6 +141,7 @@ CPU reservation for host: 1 core per NUMA
     - minimum: two nodes per profile
 -	HW specifications
     -	Boot disks are dedicated with Flash technology disks
+    - Additional 4 cores overhead per numa in case of DPDK usage for Network Intensive nodes
```
pgoyal01: @samuelhellec do we need 4 physical cores per NUMA or 4 virtual cores per NUMA? I think it is 2 physical cores (including their threads) per NUMA.

samuelhellec: @pgoyal01 I mean 4 physical cores. I tried to find  a recommandation on openstack docs but the best I came up with is https://docs.openstack.org/fuel-docs/newton/userdocs/fuel-user-guide/configure-environment/nfv/nfv-configure.html which is only in neutron release

rabiabdel:
@pgoyal01 can you please re-review/Approve

doc/ref_arch/openstack/chapters/chapter04.md
```diff
@@ -94,8 +94,15 @@ For OpenStack control nodes we use the BIOS parameters for the basic profile def
...
 -	How many nodes to meet SLA
+    - Minimum 2 nodes for high availibility using VRRP.
 -	HW specifications
+    - 3 NICs card are needed if we want to isolate the different flows :
+         - 1 NIC for Tenant Network
+         - 1 NIC for External Network
+         - 1 NIC for Other Networks (PXE, Mngt ...)
 -	Sizing rules
+    - Scale out of network node is not easy 
+    - DVR can be an option for large deployment (see more detail
```
deebhatia: see more detail where?

samuelhellec: I complete it with (see more detail on chapter 4.3.1.5) which will be completed with DVR description with the next PR of @pgoyal01

deebhatia: okay. thanks.

doc/ref_arch/openstack/chapters/chapter04.md
pgoyal01: @samuelhellec we only configure 2 physical cores (and their sibling threads) per NUMA. Do we need to specify the exact number or can we state 2 or more.

samuelhellec: @pgoyal01 , I investigate a bit more and we can have difference depending on the frequency of the CPU we have and the performance we want to deliver per compute. 
So I guess we will never reach an agreement on an exact value.

So I propose the following wording:
In case of DPDK usage extra cores must be reserved . the number of core to book for this overhead depend on the CPU frequency (Mpps available per core) and the number of Mpps we want to deliver by node.

iangardner22: PMD = Poll Mode Driver
PHY = To manage physical NICs
VHU = To manage virtual NICs

This assumes two active and two passive NICs in a bond

pgoyal01: @samuelhellec The throughput Mpps depends both on the CPU frequency but more significantly on the packet size.  Performance results demonstrate that while throughput initially increases with CPU frequency it begins to fall after a certain CPU frequency. This is because DPDK does not use  interrupts,  but  utilizes  a  busy  wait  loop  that  polls  the  NIC until at least one packet is returned. Thus, at some point the fast CPU causes DPDK to poll the NIC too often affecting the throughput.

Even with a single core @3GHz is able to achieve throughput of 30+Mpps and as you know for DPDK the performance is linear with number of cores. 

My I suggest, 
In case of DPDK usage, cores (together with their sibling threads) must be reserved for DPDK Poll Mode Drivers (PMD), hugepages set to at least 1 GB, and  all power settings disabled in the BIOS. The number of cores reserved for this overhead depends on the desired throughput (the number of Mpps we want to deliver).

samuelhellec: @pgoyal01 I am okay with your suggestion

@iangardner22 @deebhatia are you ok with this?

walterkozlowski: The latest suggestion by @pgoyal01 is very good. We can't specify exact value.  Btw: allowed packet loss will be another factor.

deebhatia: Okay with the language. One thing here is that no of cores for desired throughput is also limited by the no of queues present at the NIC.

doc/ref_arch/openstack/chapters/chapter04.md
```diff
@@ -94,8 +94,15 @@ For OpenStack control nodes we use the BIOS parameters for the basic profile def
 | …|  
  
 -	How many nodes to meet SLA
+    - Minimum 2 nodes for high availibility using VRRP.
 -	HW specifications
+    - 3 NICs card are needed if we want to isolate the different flows :
+         - 1 NIC for Tenant Network
+         - 1 NIC for External Network
+         - 1 NIC for Other Networks (PXE, Mngt ...)
```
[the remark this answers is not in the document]
pgoyal01: @ASawwaf wriet-up on network node is under development and should be posted soon

pgoyal01:
@ASawwaf re: physical cores (plus sibling thread) required for OVS-DPDK PMD: all of the documentation that I have seen seems to suggest at least 1 physical core (2 vCPUs) per 20 Gbps throughput. 

@ASawwaf @walterkozlowski @samuelhellec would it be ok to make such a statement in the document.

ASawwaf:
@pgoyal01 , it is an excellent number, but as you it is based on packet size  ( Mpps ) which is pne of the main factors
 On practical, I tested almost all the vendor ( RH , E/// , Cisco, Huawei ) OpenStack version and no one provide this 20G per one physical core allocated to PMD

@petorre , did you have numbers 


I suggest to put it as a disclaimer or Just recommendation as follow 

Number of Core allocated for OVS+DPDK ( PMD ) based on targeted throughput, packet size and packet per second, so we can start with one physical core for PMD

doc/ref_arch/openstack/chapters/chapter04.md
petorre: Last 4 lines of @pgoyal01 are very good for current VNFs in OpenStack.

iangardner22: Yes happy with that - thanks @pgoyal01

petorre:
> @pgoyal01 , it is an excellent number, but as you it is based on packet size ( Mpps ) which is pne of the main factors
> On practical, I tested almost all the vendor ( RH , E/// , Cisco, Huawei ) OpenStack version and no one provide this 20G per one physical core allocated to PMD
> 
> @petorre , did you have numbers
> 
> I suggest to put it as a disclaimer or Just recommendation as follow
> 
> Number of Core allocated for OVS+DPDK ( PMD ) based on targeted throughput, packet size and packet per second, so we can start with one physical core for PMD

Likely we should follow the old split of RA have requirements, RI implementation details and RC methodology+tools for testing (where numbers will come).
Measuring this with representative network traffic is non-trivial. Most performance marketing numbers don't correlate well with deployments.

doc/ref_arch/openstack/chapters/chapter04.md
samuelhellec: I corrected with the latest statement of pankaj where we all agree on

doc/ref_arch/openstack/chapters/chapter04.md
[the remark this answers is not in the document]
samuelhellec: I add a brief sentence to provide quick description on the network node with a reference to 4.3.1.5 Neutron for more detail

4.3.1.5 will be updated soon with full desciption of network node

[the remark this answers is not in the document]
samuelhellec: @ASawwaf  : I add a line to explain network node (detail will be in another section as explained by @pgoyal01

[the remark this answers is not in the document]
ASawwaf: @samuelhellec @pgoyal01 , Dears , I personally believe, no need to have sub-section for networking servers ( Neutron server )  requirements, it is one of implementation of L3 , we can have other option as OVN , DVR, VR 

So from MHO , Let covers it in details in neutron section

rabiabdel:
@petorre can you re-review

Issue #1316, opened 2020-03-17
[RM-Appendix A]What does "vGPUs" mean?


YuPengZTE:
What does "vGPUs" mean, In ```A textbook example of a capability that transgresses this principle, is SR-IOV. Other, less notable, yet very important examples include vGPUs and other acceleration hardware, such as FPGA``` of [A.5 Hardware-Dependent Coding Policies](https://github.com/cntt-n/CNTT/blob/master/doc/ref_model/chapters/appendix-a.md#a5-hardware-dependent-coding-policies) and [7.1.2 Hardware-Dependent Coding](https://github.com/cntt-n/CNTT/blob/master/doc/tech/policies.md#712-hardware-dependent-coding)?

vGPUs are not hardware, it is better to delete "v".

P.S.
In Kubernetes, virtual GPU can be provided by sharing, such as [gpushare-device-plugin](https://github.com/AliyunContainerService/gpushare-device-plugin) and [gpushare-scheduler-extender](https://github.com/AliyunContainerService/gpushare-scheduler-extender); In OpenStack, virtual GPU can be provided by passthrough.

ASawwaf:
@YuPengZTE ,We Just refer to GPU as one of special requirements beside HW acceleration to be considered in our roadmap

YuPengZTE:
@ASawwaf, I see, and I fix it with https://github.com/cntt-n/CNTT/pull/1324 and https://github.com/cntt-n/CNTT/pull/1323.
Please check them, thanks.

YuPengZTE:
In order to record more discussions, I reopened this issue.

For some previous discussions, please see https://github.com/cntt-n/CNTT/pull/1324

PR #1324, opened 2020-03-18
[TECH-POLICIES]Delete "v" from "vGPUs"

YuPengZTE:
Fix https://github.com/cntt-n/CNTT/issues/1316


doc/tech/policies.md
```diff
@@ -21,7 +21,7 @@ One such case is where CNTT strategically deems a given capability as mandatory,
...
+CNTT realizes there are implications to these types of prohibitions and practical limitations on the ability to live by the Abstraction Principle. A textbook example of a capability that transgresses this principle, is SR-IOV. Other, less notable, yet very important examples include GPU and other acceleration hardware, such as FPGA. Less obvious, yet critically important examples are the programs comprising the actual VNFs. As workloads execute native microarchitecture opcodes, their instruction sets effectively constitute an ABI (Application Binary Interface), which is by definition dependent on the host's hardware. Additionally, these programs may or may not, attempt to execute vendor-specific extensions to standard instruction sets, such as x86 or ARM.
```
markshostak: Hi @YuPengZTEl, Referring back to your original question in Issue 1316, "What does "vGPUs" mean?", the intent of the language is to illustrate examples of _virtualization_ implementations that aren't quite cloud-native. Hence, it's the **v**GPU that's the concern. There's no issue with the physical GPU, per se. If you'd like to clarify the language, I do agree that it could be made more explicit, but it's the hardware reference (i.e. second half of the sentence) that would need to be fixed, in that it's the virtual manifestation of the hardware, not the hardware itself, that creates the transgression. Can you clarify that part of the sentence, instead?

P.S. The issue (1316) looks like it may have been accidentally closed.

YuPengZTE: Thank @markshostak, I want to clarify the language.
Is the modification you said like this:
```Other, less notable, yet very important examples include acceleration hardware, such as GPU and FPGA.```?

P.S.
vGPU is a very complicated topic. If combined with cloud-native, I don't know what cloud-native vGPU is.
The example I just gave is to illustrate the complexity of vGPU, which is not suitable for introduction in this chapter, and even less suitable for discussion in this chapter.
From the view of VNF, it cares about the computing power of the GPU: Floating-point operations per second, but not how the GPU is virtualized.
From the view of infrastructure, GPUs are just an acceleration of computing and should be managed as ordinary hardware. They are not suitable for discussing how to virtualize GPUs.

Comment (markshostak) [reply doc/tech/policies.md]: Hi @YuPengZTE,
>
> Other, less notable, yet very important examples include acceleration hardware, such as GPU and FPGA.?

More like, "Other, less notable, yet very important examples include vGPUs as well as virtualised manifestations of other acceleration hardware, such as FPGAs.".

>
> ...VNF cares about the computing power of the GPU... ...but not how the GPU is virtualized.

I'll push back on that a bit. While I concur, by definition, a VNF's code will access whichever API it's designed to, and do so with no other affinity, in the larger CNTT picture, standardizing _how_ a GPU or any other resource is virtualized is one of the primary purposes of CNTT. Without this normalization of virtualization, VNFs require hardware-specific s/w (aka drivers), which breaks the intended VNF portability/compatibility. 

Also, I just realized this change is being proposed for 2 places, but they are split between 2 PRs that don't cross-reference each other. **The same change needs to be applied in both places.** The other PR was #1323 (now merged).

Comment (YuPengZTE) [reply doc/tech/policies.md]:
```
Without this normalization of virtualization, VNFs require hardware-specific s/w (aka drivers), which breaks the intended VNF portability/compatibility
```
Yes, you are right, we need this normalization of virtualization.
Unfortunately, at present, the way of virtualization depends on different hardware providers' solutions based on hardware, especially in the field of acceleration hardware, GPU is a typical example.
It would be better if CNTT can push this forward.

```
From the view of infrastructure, GPUs are just an acceleration of computing and should be managed as ordinary hardware.
```
It may be more appropriate to replace VNF with APP here.
When an app uses GPUor vGPU, there is no difference to the app. For example, after the NVIDIA GPU is virtualized, the app can call the GPU without modification.

```
From the view of VNF, it cares about the computing power of the GPU: Floating-point operations per second, but not how the GPU is virtualized.
```
GPU virtualization should be handled by infrastructure management。

**Comment by YuPengZTE** [comment]
@markshostak, I reopen [[RM-Appendix A]What does "vGPUs" mean?](https://github.com/cntt-n/CNTT/issues/1316), and we would add more discussions.

It is better to modify the document after we have discussed it clearly.

File: doc/tech/policies.md (hunk 1 above)
Comment (markshostak) [reply doc/tech/policies.md]: >
> ...we need this normalization of virtualization. ...It would be better if CNTT can push this forward.

I totally agree.. If you want to drive it in the industry, I'll support you any way I can. Until then, it's on the non-conforming tech list.

>
> It may be more appropriate to replace VNF with APP here.

I'll borrow your line "What does App mean?". :-) I'm not sure in your use of the term "App" in this context, what the specific difference between VNF and App is, that you are trying to remove. Can you clarify? If you really want to change it, how about "Workload"? IMHO App is too vague and abstract.

Bottom line - What are you wanting to do with the proposed edit in this PR?

Comment (YuPengZTE) [reply doc/tech/policies.md]: About this PR, I just want to delete “v“ for clarifying the language.
 :smile:

The other content we discuss is an expectation: this normalization of virtualization.
There is no more thinking.
Thanks for your support @markshostak.

The difference between VNF and App, I think it is like the difference of NFVI and Cloud Infrastructure which is mentioned in [ 2020-03-11 TSC Meeting ](https://wiki.lfnetworking.org/display/LN/2020-03-11+-+%5BCNTT+-+Technical+Steering%5D+-+Meeting+Agenda+and+Minutes).
The VNFs are special Apps.

Comment (markshostak) [reply doc/tech/policies.md]: We can't delete the "v", as there's no issue with physical GPUs and hence the vGPUs reference is not in error. The object of the sentence intentionally includes _virtualized_ GPUs (**v**GPUs).

**Comment by YuPengZTE** [comment]
@markshostak, I have modified it like this: 
```
Other, less notable, yet very important examples include vGPUs as well as virtualized manifestations of other acceleration hardware, such as FPGAs.
```

What you said makes sense, since accelerated hardware virtualization is not discussed elsewhere, then we start here and hope that more discussions on hardware virtualization will take place in other chapters.

PR #1337, opened 2020-03-19
[RA2 Ch04]: Runtime requirements

### LINKED ISSUE #1204: [RA2 Ch02] OS kernel and resource isolation requirements
I think we need to be more specific on the requirements

`req.sec.gen.04 `| General | The Architecture should provide Operating System kernel isolation.
`req.sec.gen.05 `| General | The Architecture must provide compute resources isolation.

Does Operating System kernel isolation here refer to using kata containers, containers in VM, etc?

Does Compute Resource isolation here refer to isolation provided by the standard container runtime?

CsatariGergely:
Fixes #1204

### DIFF
--- a/doc/ref_arch/kubernetes/chapters/chapter04.md
+++ b/doc/ref_arch/kubernetes/chapters/chapter04.md
@@ -90,7 +90,12 @@ In order to support `req.inf.com.03`, the chosen runtime must be conformant with
 - CRI-O
 - Frakti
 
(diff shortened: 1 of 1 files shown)


CsatariGergely:
> 
> 
> I know this would need to be propogated throughout this chapter (not as part of this PR), but can we change terms like "The Architecture must support" to "The Architecture specifies" in this chapter. I think any "must support" things need to live in chapter02.md only.

Done

CsatariGergely:
> 
> 
> Can you remove the m after req.sec.gen.04

Done

Issue #1345, opened 2020-03-20
[Overall] Generated documentation contains mailformed tables


CsatariGergely:
The [generated documentation](https://cntt-n.github.io/CNTT/doc/ref_model/chapters/chapter05.html) contains mailformed tables in several places.

rabiabdel:
Well spotted @CsatariGergely . i am happy to look at this.

PR #1347, opened 2020-03-20
[RM] Fix tables

rabi-abdel:
Fixes https://github.com/cntt-n/CNTT/issues/1345

There is a need to include space after each table for it to be rendered correctly.


rabiabdel:
> Even if the tables are ok from md pov, we usually align the columns (see a few tabs in RC).
> It seems more than doable here.

OK, make sense, will do that.

rabi-abdel:
@collivier done as you requested.

PR #1353, opened 2020-03-20
[RM Ch01 ] Introduction -1.3 Principles

ASawwaf:
"if one member wanted to use Ceph, and another member wanted to use Swift " replaced by if one member wanted to use Software-Defined storage (SDS) as CEPH, and another member wanted to use Storage attached network(SAN)


walterkozlowski:
@ASawwaf you mean Network Attached Storage (NAS) I presume? Btw: can you clarify your motivation for the change to this example?  What do you think was wrong in the original example? Implementing object storage as CEPH-based versus Swift-based is a real-life architectural choice depending e.g. on nature of workloads.

ASawwaf:
@walterkozlowski , lets put it in this way, we have the Openstack storage as Cinder(block)/Swift ( object ) and Manilla ( File ) all this type of storage consider a frontend, they need some backend for so we always use SDS(CEPH) or SAN

In the above original exmaple, we are comparing apple to orange ( CEPH with Swift )

regarding SAN/NAS, SAN into a block and NAS more into a file and SAN used as backend storage for OpenStack 



So our target to make generic 

I hope it is clear

doc/ref_model/chapters/chapter01.md
```diff
@@ -31,7 +31,7 @@ This section introduces the high-level principles of infrastructure abstraction
...
+        >*Depending on the relationships and substitutability of the component(s) in question, it may be possible to mitigate component incompatibility by creating annexes to a single Architecture, rather than creating an additional Architecture. With this approach, the infrastructure architecture designers might implement the Architecture as described in the reference document, however when there is a potential for incompatibility for  particular component, they would select their preferred option from one of the relevant annexes. For example, if one member wanted to use Software-Defined storage (SDS) as CEPH , and another member wanted to use Storage attached network(SAN) , assuming the components are equally compatible with the rest of the Architecture, there could be one annex for the Ceph implementation and one annex for the SAN implementation.
```
kedmison: minor nits: spacing, and consistency in capitalization.
```suggestion
        >*Depending on the relationships and substitutability of the component(s) in question, it may be possible to mitigate component incompatibility by creating annexes to a single Architecture, rather than creating an additional Architecture. With this approach, the infrastructure architecture designers might implement the Architecture as described in the reference document, however when there is a potential for incompatibility for particular component, they would select their preferred option from one of the relevant annexes. For example, if one member wanted to use Software-Defined storage (SDS) as CEPH, and another member wanted to use Storage attached network(SAN), assuming the components are equally compatible with the rest of the Architecture, there could be one annex for the CEPH implementation and one annex for the SAN implementation.
```

rabi-abdel:
@kedmison can you review please

PR #1355, opened 2020-03-22
[RA-1 Ch02] Make Object Storage mandatory

### LINKED ISSUE #1313: [RA-1 Ch02] Make Object Storage mandatory
req.inf.stg.02 -- make this a "must" | Make Object Storage mandatory
Add a note that the some operators may choose not to implement.

Add conditional notes to some other requirements.

pgoyal01:
Object Storage made mandatory with an opt-out for operators.

Fixes #1313


deebhatia:
@pgoyal01 Can we do it similar to what RA2 did here?
https://github.com/cntt-n/CNTT/pull/1130/files

PR #1356, opened 2020-03-23
[RM] Replicates Changes from TECH: Transition Plan 7.1.2

markshostak:
This is maintenance work to _try_ to keep these cloned sections in sync.

NOTE: **There is no new content in this PR.** This replicates agreed content from merged PR #1324


markshostak:
@YuPengZTE I tried to add you as a reviewer, but for some reason I don't see you on the list. @rabi-abdel  Does something need to be setup or corrected in GH?

kedmison:
right now we're not virtualizing GPUs, and the practice is to bind a whole GPU into a VM in Openstack (and the default is to do the same in Kubernetes).

@YuPengZTE proposed dropping the v from vGPU in https://github.com/cntt-n/CNTT/issues/1316 so it would be good to resolve these two conflicting directions.

ASawwaf:
also , can we put in as GPU not vGPU

markshostak:
> 
> 
> @markshostak , is it ABI or API , Did you mean here

ABI for this one.

markshostak:
> 
> @YuPengZTE proposed dropping the v from vGPU in #1316 so it would be good to resolve these two conflicting directions.

This PR merely replicates the outcome of the approved PR #1324 associated w/ the Issue you referenced. There is no new content here. Also, the outcome of the proposal you referenced was to not drop the "v". 

> 
> right now we're not virtualizing GPUs, and the practice is to bind a whole GPU into a VM in Openstack (and the default is to do the same in Kubernetes).

Exactly, and this language describes the _non-conformance_, not the current practice. vGPU is the non-conformance.

rabiabdel:
overruled by the other PR: https://github.com/cntt-n/CNTT/pull/1358

markshostak:
> 
> overruled by the other PR: #1358

@rabi-abdel Hi Rabi, By "overruled", do you mean combined with? :-)

PR #789, opened 2019-12-12
[RM Appendix A] text on decoupling app from infra

petorre:
This addresses comments in pull request #653 and helps with issue #678.

PR #1376, opened 2020-03-25
[RM Appendix] Levels of Decoupling

petorre:
Text on Levels of Decoupling from previous #789 that I will close. Here with all comments applied and without mentioning "CNTT".


rabi-abdel:
@kedmison once reviewed/approved by you, it will be merged.

PR #1370, opened 2020-03-25
[RA2 Ch4] updates to include NSX Container plugin(NCP)

xgao3:
The NSX Container Plugin (NCP) is a container plugin that integrates with NSX-T Data Center to build Kubernetes networking and security objects. The NCP can also serve as a metaplugin/CNI multiplexer that is capable of attaching multiple interfaces to Kubernetes Pods to provide multiple NIC container support

PR #1379, opened 2020-03-26
[RA2 Ch4]: Removal of VMware NCP

### LINKED ISSUE #1378: [RA2 Ch4]: Remove not open source CNI plugin
#1370 was prematurely merged and added the not open source NSX Container Plugin to RA2. This should be removed.

CsatariGergely:
Fixes #1378

### DIFF
--- a/doc/ref_arch/kubernetes/chapters/chapter04.md
+++ b/doc/ref_arch/kubernetes/chapters/chapter04.md
@@ -100,27 +100,25 @@ These specifications cover the [full lifecycle of a container](https://github.co
 
 The used CNI multiplexer/metapulgin may be [DANM](https://github.com/nokia/danm) as it provides the possibility to use several other CNI plugins (`req.inf.ntw.16`) and provides an API based solution to administer the networks (`req.inf.ntw.10`) from a central point (`req.inf.ntw.11`).<br>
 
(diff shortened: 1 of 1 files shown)


ramkri123:
Agree with Tom's observation. 

There is no mandate for the CNI plugin to be open source as long as the plugin adheres to the CNI specification -- https://github.com/containernetworking/cni. Other examples of non-open-source plugins are ACI, Nuage etc. -- they are described in https://kubernetes.io/docs/concepts/cluster-administration/addons/.

tomkivlin:
> Agree with Tom's observation.
> 
> There is no mandate for the CNI plugin to be open source as long as the plugin adheres to the CNI specification -- https://github.com/containernetworking/cni. Other examples of non-open-source plugins are ACI, Nuage etc. -- they are described in https://kubernetes.io/docs/concepts/cluster-administration/addons/.

I think the point is that there is a CNTT principle that we only used open source technologies.  However, it is a principle, so I think worth adding but highlighting.

ramkri123:
Hi Tom,

Would the following changes help clarify.

1) In the NCP text
Note that NCP is not a open source CNI plugin, but adheres to the CNI specification -- https://github.com/containernetworking/cni. Other examples of non-open-source CNI plugins are ACI, Nuage etc. -- they are described in https://kubernetes.io/docs/concepts/cluster-administration/addons/.

2) In the NCP column in the table 
Not open source

Thanks,
Ramki

CsatariGergely:
> 
> 
> I don't think the whole merge needs reversing. I am ok with NCP being listed, but we do need to highlight that it is not open source, perhaps a reference to the CNTT principles in the comparison table?

1) #1370 was merged before reaching a consensus. 
2) I think having a principle to use only open source technologies would mean that we use only open source technologies.
3) If I understand correctly this CNI assumes that the Kubernetes cluster runs on top of VM-s in a VMware vCloud environment.
4) I'm still not convinced that this is a metaplugin

tomkivlin:
Meeting 2nd April:
- Tom agreed with points 1/2 above
- Principle compliance should be made, unless there isn't an option that complies with all principles
- Do we need a discussion about "open source" - what it means - is an open source software owned by one company with contributions only from that company any better than proprietary?  Yes - but not in this PR :) @tomkivlin to raise an issue.

CsatariGergely:
> 
> 
> another is one word, otherwise lgtm

Done

Issue #950, opened 2020-01-21
[RA2] OS Modularity requirements and features


tomkivlin:
Suggested "minimum requirements":

- kernel version
- C library and version (glibc / musl c)
- Core utilities? (GNU Core Utils?)

tomkivlin:
@peterwoerndle - I was talking to another vendor about this today (BSS space) and thought I'd start jotting ideas down in an Issue.

peterwoerndle:
When you're referring to "minimum requirements", we should differentiate between Host OS and base OS, or? I assume some requirements would have to be matched, others would only apply to one or the other.

I remember that for the host OS a minimal set of kernel modules to be available have been mentioned but not further defined.

peterwoerndle:
@tomkivlin should I just start a PR so we have some concrete text proposal to discuss. In the Prague meeting we've been in agreement to document the OS modularity. The details on "minimum requirements" we can then add iteratively.

tomkivlin:
@peterwoerndle yes, I think some text in Chapter 04 would be useful to then start the discussion, unless you were thinking somewhere else?

peterwoerndle:
Yes, Ch04 sounds the right place to document this. I'll create a PR asap.

tomkivlin:
@peterwoerndle did you create a PR for this? Nothing is coming up as being linked to this.  Let me know if you need me to take it on instead.

jnummelin:
As a newbie in CNTT space, maybe a stupid question. :) Why would we want/need to pin down ANY requirements on kernel/OS/C lib level? I know k8s itself imposes some minimum requirements, but those should be implied as-is on CNTT too as we're basing RA2 anyway on k8s.

tomkivlin:
> As a newbie in CNTT space, maybe a stupid question. :) Why would we want/need to pin down ANY requirements on kernel/OS/C lib level? I know k8s itself imposes some minimum requirements, but those should be implied as-is on CNTT too as we're basing RA2 anyway on k8s.

We may have worded badly so far, but the intent of this is to increase the modularity between the Container Host OS and the Container Base Image.  Today, many OS vendors will specify a need for tight coupling between the two, essentially removing much of the stated benefits of Kubernetes from a portability and abstraction point of view.  i.e. if you're using Linux distro A on the Container Host OS, you _must_ use Linux distro base image within the containers too.

[...]

PR #1380, opened 2020-03-26
[RA2 Ch04] Added statement to not restrict container base OS

peterwoerndle:
This addresses the OS modularity requirement discussed in Prague and in #950 .

### DIFF
--- a/doc/ref_arch/kubernetes/chapters/chapter04.md
+++ b/doc/ref_arch/kubernetes/chapters/chapter04.md
@@ -31,6 +31,7 @@ In order for a Host OS to be conformant with this Reference Architecture it must
 - A version of the Linux kernel that is [compatible with kubeadm](https://kubernetes.io/docs/reference/setup-tools/kubeadm/implementation-details/#kubeadm-init-workflow-internal-design) - this has been chosen as the baseline because kubeadm is focussed on installing and managing the lifecycle of Kubernetes and nothing else, hence it is easily integrated into higher-level and more complete tooling for the full lifecycle management of the infrastructure, cluster add-ons, etc.
 - Windows Server 2019 (this can be used for worker nodes, but be aware of the [limitations](https://kubernetes.io/docs/setup/production-environment/windows/intro-windows-in-kubernetes/#limitations)).
 - In order to support `req.gen.cnt.03` (immutable infrastructure), the Host OS must be easily reproduced, consistent, disposable, will have a repeatable deployment process, and will not have configuration or artifacts that are modifiable in place (i.e. once it is running).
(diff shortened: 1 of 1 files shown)


fkautz:
We should add that a version of a Linux kernel capable of running modern Linux containers at the very least.

Vendors may also have their own requirements. E.g. a vendor utilizing kernel based wireguard will need a much more modern kernel.

YuPengZTE:
@peterwoerndle, could you add some examples for easy understanding?

peterwoerndle:
@fkautz We do have a requirement on the HostOS kernel version to be 3.10+ in table 4-1. I suggest to document further requirements on the kernel in that table.

@YuPengZTE an example with regards to what we want to avoid is that if I use distro X for my HostOS i.e. in my RA2 compliant NFVI, then I'm obliged to use distro X also to build my container images for my workloads.

tomkivlin:
> @peterwoerndle, would you add some examples for easy understanding?

Another example is that some of the specifications we document here will flow into the OVP badge for CNFs, as well as the badge for a Kubernetes platform.  And this is an example of where we would be specifying that those CNFs are not relying on anything other than the bare minimum from a Host OS.

peterwoerndle:
@tomkivlin can we put requirements on the CNFs in an NFVI RA? Strictly speaking the CNF is out of scope of RA2 as per the overview chapter. I think this relates a bit to @fkautz comment with regards to what are the minimum kernel features we expect from the HostOS kernel. We can then add a statement like: _A CNF running on a RA2 compliant architecture is expected to not require additional capabilities from the HostOS kernel._

tomkivlin:
@peterwoerndle I thought part of the output of CNTT was a set of standard interfaces and specifications for what the CNFs would be consuming.  So whilst the CNF isn't in scope, the interfaces from the NFVI/CaaS are in scope and so we will be provided requirements to CNF vendors as to what those interfaces are (and aren't).

tomkivlin:
@peterwoerndle @fkautz can I suggest we merge this one (which is about the Host OS not restricting the container base image) and we can discuss wording to deal with the container base image not putting restrictions/requirements (or if so, what they are) on the Host OS.

YuPengZTE:
Thanks, @peterwoerndle @tomkivlin.

> @tomkivlin can we put requirements on the CNFs in an NFVI RA? Strictly speaking the CNF is out of scope of RA2 as per the overview chapter. I think this relates a bit to @fkautz comment with regards to what are the minimum kernel features we expect from the HostOS kernel. We can then add a statement like: _A CNF running on a RA2 compliant architecture is expected to not require additional capabilities from the HostOS kernel._

This statement is very clear. :+1:

Issue #1382, opened 2020-03-27
[TSC] Define Technical Contribution Metrics


rabi-abdel:
For Fairer, More transparent method, we would like to collect the metrics of which the community feel reflective of the great work made by everybody.

PR #1383, opened 2020-03-27
[TSM] Define Technical Contribution Metrics

rabi-abdel:
For Fairer, More transparent method, we would like to collect the metrics of which the community feel reflective of the great work made by everybody.

Fixes https://github.com/cntt-n/CNTT/issues/1382

### DIFF
--- a/doc/gov/chapters/chapter08.md
+++ b/doc/gov/chapters/chapter08.md
@@ -21,5 +21,23 @@
 <a name="8.3"></a>
 ## 8.3.1 Contribution Metrics
 
+| Metric Category | Metric Ref | Definition                                      | Per Individual or Company |
+|-----------------|------------|-------------------------------------------------|---------------------------|
+| A (Functest)    | A.1        | Number of Lines added & Number of Lines Removed | Individual                |
+|                 | A.2        | Number of Lines Reviews Submitted.              | Individual                |
+|                 | A.3        | .. (Others)                                     | Individual                |
+| B (Airship )    | B.1        | Number of Lines added & Number of Lines Removed | Individual                |
+|                 | B.2        | Number of Lines Reviews Submitted.              | Individual                |
+|                 | B.3        | .. (Others)                                     | Individual                |
+| C (Functest)    | C.1        | Number of Lines added & Number of Lines Removed | Individual                |
+|                 | C.2        | Number of Lines Reviews Submitted.              | Individual                |
+|                 | C.3        | .. (Others)                                     | Individual                |
+| D (CIRV)        | D.1        | Number of Lines added & Number of Lines Removed | Individual                |
+|                 | D.2        | Number of Lines Reviews Submitted.              | Individual                |
+|                 | D.3        | .. (Others)                                     | Individual                |
+| E (CIRV)        | E.1        | Number of Lines added & Number of Lines Removed | Both                      |
+|                 | E.2        | Number of PRs Created.                          | Both                      |
+|                 | E.3        | Number of Reviews Created                       | Both                      |
+|                 | E.4        | Number of Comments Made                         | Both                      |
 
 
--- a/doc/gov/chapters/chapter08.md
+++ b/doc/gov/chapters/chapter08.md
@@ -5,14 +5,21 @@
 
 ## Table of Contents
 * [8.1 Introduction](#8.1)
-* [8.2 Heading](#8.2)
-* [8.3 Heading](#8.3)
+* [8.2 Recruiting](#8.2)
+* [8.3 Engagement](#8.3)
+  * [8.3.1 Contribution Metrics](#8.3)
 
 <a name="8.1"></a>
 ## 8.1 Introduction
 
 <a name="8.2"></a>
-## 8.2 Heading
+## 8.2 Recruiting
 
 <a name="8.3"></a>
-## 8.3 Heading
+## 8.3 Engagement
+
+<a name="8.3"></a>
+## 8.3.1 Contribution Metrics
+
+
+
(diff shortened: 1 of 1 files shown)


rabiabdel:
Please provide feedback and suggestions.

pgoyal01:
May I suggest that the TSC specify the Goal and some justification, and then task a small WG charged with defining in a methodical manner to define the measures etc. that help achieve the Goal.

Personally, this topic is not of interest to me.

tomkivlin:
> May I suggest that the TSC specify the Goal and some justification, and then task a small WG charged with defining in a methodical manner to define the measures etc. that help achieve the Goal.
> 
> Personally, this topic is not of interest to me.

I agree. If this is a means to an end, it's not clear what that end is other than individuals being able to justify the time spent on CNTT (but even that is tenuous).  More useful, I think, would be business metrics relating to the goals of CNTT such as TCO reduction, reduction in end-to-end certification time, etc.

If we do have to go to "implementation" metrics then I would suggest things that we think will contribute to those business metrics such as number of tests contributed (in total, not per company or individual), number of conformant vendor implementations, etc.

And finally if we do have to measure levels of contribution then let's do the number of contributions as a community to show how the community is growing, rather than measuring the "effort" of individuals.

tomkivlin:
Also - why are we tracking the number of lines etc. for Airship? Isn't that governed elsewhere? And if Airship, why not e.g. kubeadm or whichever installer we choose in RI2?  I don't understand the logic I'm afraid.

iangardner22:
I have no strong opinion on this - except agree it needs to represent true effort spent and sometimes that effort isn't reflected in github commits or reviews.   I will accept the majority view on this one.

ulikleber:
I am just a bit confused why we have metrics about OPNFV projects. I would understand if we have metrics for our own workstreams.

karinesevilla:
I also think that the goal of these metrics must be clearly described and that focusing too much on these metrics can be counterproductive in term of quality.

Now, if metrics must be provided, regarding documentation, I think that the metrics we use don't reflect enough the collaborative part of work. I consider it is important to associate co-authors to commits when text is written by several authors or in the case of valuable reviews and contributions improving the PRs.
This can be done in Github, but we must be sure then that it's taken into acount when processed by analytics.

collivier:
I will be very cautious regarding the current content of the website as it classifies users and companies. By its nature, the rankings defacto require trustable and agreed criteria.

Here we would also separate the metrics and the rankings (or filters).

We commonly select two metrics in the open source projects (LFN, OpenStack, etc.): commits and reviews. Commits mostly highlight the most active developers and reviews the core (aka key) developers (they could be the same). They are commit rankings everywhere. Reviews are mostly used for the promotions to Core Developers.

As mentioned during the governance meeting, the processes in CNTT are incorrect to fairly count the commits:
- all single commits in PRs were merged as they were and now they are squashed in a single commit
- most contributor doesn’t add the Co-Author-By label to list all people working on the change (Co-writers or very good reviews)

My understanding is that the project management (PTL in opensource) doesn’t need a specific metric in open source. Following the meritocratic model, PTL is usually elected from the core developer list.
And if we apply the right PR management in CNTT, the PRs will simply duplicate the commits.

Community message seems unclear and possibly unlinked to the contribution effort (meeting invitations on mailing lists?).

Yes OPNFV already counts the commits
Of course Functest includes changes for all and not only for CNTT but we can hardly evaluate the right numbers without too much tags and processes (and then human errors).

walterkozlowski:
My understanding is that the overall goal is to encourage the community to participate.  The question is whether publication of these metrics will do the trick.  As with any such metric, everybody will have a different opinion on such metrics. Most likely there is no set of metrics which will be completely informative and completely fair at the same time.  For example, I do not think that the currently proposed set of metrics would show a fair measure of great efforts by @rabiabdel!  My suggestion is that this topic needs to be thoroughly thought through. Wrong way of doing this may lead to results contrary to our intentions.

tomkivlin:
Couldn't agree more @walterkozlowski - the law of unintended consequences!

@rabi-abdel - feels to me from this thread that before we define what the metrics are we want to track, we need to define the desired outcomes, the overall goal.  Pointing out who has made the higher number of commits and which company they work for isn't going to help prove CNTT is valuable (I can see how the individuals and companies would use it as justification, but that's not the goal here, is it?)

rabi-abdel:
> Couldn't agree more @walterkozlowski - the law of unintended consequences!
> 
> @rabi-abdel - feels to me from this thread that before we define what the metrics are we want to track, we need to define the desired outcomes, the overall goal. Pointing out who has made the higher number of commits and which company they work for isn't going to help prove CNTT is valuable (I can see how the individuals and companies would use it as justification, but that's not the goal here, is it?)

@tomkivlin @walterkozlowski @collivier @pgoyal01 @ulikleber @karinesevilla agree with all comments. I suspected this PR will be controversial. I suggest closing this PR and have the topic discussed during TSM meeting. 

For me, it is very simple, we have been given a question by Governance: 

> **What are the metrics we need to show in the Dev analysis tool LFN is using?** 
they are currently showing some metrics that the general feeling was that they are not representing the contributions made.  do we NOT need to influence what metrics are shown?

walterkozlowski:
@rabi-abdel I agree with closing and having a TSM discussion on what we want to achieve

pgoyal01:
I will caution against the use of lines of code in any metrics. Some folks write a single short to the point sentence while someone else writes multiple lines that really do not say anything.

rabi-abdel:
ok I will remove this Pull Request as I don't see a straight forward way to resolve it. we will continue giving LFN feedback about what we would like to be displayed. thanks all for feedback.

PR #1389, opened 2020-03-27
[RA1 Ch4] More precise DPDK usage

petorre:
Fixes #1388.

PR #1402, opened 2020-03-29
[RA1 Ch4] More on NUMA-balanced IO, memory, storage

petorre:
... placement of NICs to get NUMA-balanced system (balancing the I/O, memory, and storage across both sockets)...


petorre:
Related to #1389. (Sorry that I likely do something wrong, but there it did not allow me to Edit.)

ASawwaf:
@peterwoerndle  ,We need to add "to avoid a cross NUMA traffic to boost the performance"

but I don't know where is the best place to add . is  cloud infrastructure or workload

petorre:
> @peterwoerndle ,We need to add "to avoid a cross NUMA traffic to boost the performance"
> 
> but I don't know where is the best place to add . is cloud infrastructure or workload

Added "to avoid crossing interconnect between CPUs".

ASawwaf:
@peterwoerndle , I reword it "NUMA-CPU alignment" instead of NUMA setting

petorre:
> @peterwoerndle , I reword it "NUMA-CPU alignment" instead of NUMA setting

I would keep it simply as "NUMA settings" because:
- NUMA is also for IO and memory.
- Used CPUs will be on their NUMA node.

petorre:
(Moving changes to #1389 that I can edit again, please comment/change further there, I'm closing this PR.)

Issue #1420, opened 2020-04-01
[TSC CONTRIBUTING] Remove requirement to always require an Issue when creating a PR


tomkivlin:
Specifically referring to [L53-L55](https://github.com/cntt-n/CNTT/blame/master/CONTRIBUTING.md#L53-L55):

> - PRs must only be created when there is an issue present and a decision to create a PR is made. A PR must reference the issue it is resolving in the description field.
>  - using GitHub predefined keyword "**Fixes Issue#**" will automatically close the issue which the PR is resolving when it is merged.
>- A PR for an issues must only be created by the contributor assigned to the issue by the respective WSL.

I would propose that for minor editorial changes such as typos, correcting table/figure numbers, correcting broken links etc. that a PR can and should be raised without an Issue being needed.  Proposed change being something like:

> - **PRs dealing with minor editorial changes can be raised without a corresponding Issue, but must be labelled as "**Minor**".  Otherwise,** _PRs must only be created when there is an issue present and a decision to create a PR is made. A PR must reference the issue it is resolving in the description field._
>  - _using GitHub predefined keyword "**Fixes Issue#**" will automatically close the issue which the PR is resolving when it is merged._
> - _A PR for an issues must only be created by the contributor assigned to the issue by the respective WSL._

CsatariGergely:
I like the idea.

PR #1430, opened 2020-04-03
[TSC] Update CONTRIBUTING.md

rabi-abdel:
Fixes https://github.com/cntt-n/CNTT/issues/1420

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,7 +53,7 @@ Every Issue and PR should be tagged with one or more of the following labels to 
 - PRs dealing with minor editorial changes can be raised without a corresponding Issue, but must be labelled as "Minor". Otherwise, PRs must only be created when there is an issue present and a decision to create a PR is made. 
 - A PR must reference the issue it is resolving in the description field.
    - using GitHub predefined keyword "Fixes Issue#" will automatically close the issue which the PR is resolving when it is merged.
-- A PR for an issues must only be created by the contributor assigned to the issue (either by self assignment or by the respective WSL).
+- A PR for an issue must only be created by the contributor assigned to the issue (either by self assignment or by the respective WSL).
 - Only the person assigned ("assignee") to a PR may edit it. The normal process is for others to make suggests in "Comments" but not directly edit a PR unless the assignee has granter another contributor permission to contribute directly to the PR.
 - A PR is recommended to only create/modify content within the scope of a single workstream.
 - A PR should only add/change content related to the issue associated with the PR.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -50,9 +50,10 @@ Every Issue and PR should be tagged with one or more of the following labels to 
 
 <a name="prs"></a>
 ### Pull Requests
-- PRs must only be created when there is an issue present and a decision to create a PR is made. A PR must reference the issue it is resolving in the description field.
-  - using GitHub predefined keyword "**Fixes Issue#**" will automatically close the issue which the PR is resolving when it is merged.
-- A PR for an issues must only be created by the contributor assigned to the issue by the respective WSL.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -50,9 +50,10 @@ Every Issue and PR should be tagged with one or more of the following labels to
 
 <a name="prs"></a>
 ### Pull Requests
-- PRs must only be created when there is an issue present and a decision to create a PR is made. A PR must reference the issue it is resolving in the description field.
-  - using GitHub predefined keyword "**Fixes Issue#**" will automatically close the issue which the PR is resolving when it is merged.
-- A PR for an issues must only be created by the contributor assigned to the issue by the respective WSL.
+- PRs dealing with minor editorial changes can be raised without a corresponding Issue, but must be labelled as "Minor". Otherwise, PRs must only be created when there is an issue present and a decision to create a PR is made.
```
pgoyal01, line 53: Can we clarify "minor editorial"? Is it spelling or grammatical changes or does it also include more substantive policy/meaning changes? For example, this PR proposes a change to the process but is classified as "minor".

CONTRIBUTING.md
```diff
@@ -50,9 +50,10 @@ Every Issue and PR should be tagged with one or more of the following labels to
 
 <a name="prs"></a>
 ### Pull Requests
-- PRs must only be created when there is an issue present and a decision to create a PR is made. A PR must reference the issue it is resolving in the description field.
-  - using GitHub predefined keyword "**Fixes Issue#**" will automatically close the issue which the PR is resolving when it is merged.
-- A PR for an issues must only be created by the contributor assigned to the issue by the respective WSL.
+- PRs dealing with minor editorial changes can be raised without a corresponding Issue, but must be labelled as "Minor". Otherwise, PRs must only be created when there is an issue present and a decision to create a PR is made. 
+- A PR must reference the issue it is resolving in the description field.
+   - using GitHub predefined keyword "Fixes Issue#" will automatically close the issue which the PR is resolving when it is merged.
+- A PR for an issues must only be created by the contributor assigned to the issue (either by self assignment or by the respective WSL).
```
pgoyal01, line 56: Typo: "an issues" ---> "an issue"

CONTRIBUTING.md
ASawwaf: definition of "minor" is indeed

walterkozlowski: Perhaps we should have another PR label like "Agreed" which would qualify for being raised without an issue?  This is the case with the example above: I would agree it is not "Minor" (it changes the process, as @pgoyal01 wrote) but it was agreed in TSC so it can fast processed.

tomkivlin: The minor label already had a definition,:
> Pull request suggesting minor/editorial changes

I don't set the problem with this PR being labelled as minor as it is a small change,  but as it is not editorial it had a corresponding issue. In other words I take the `/` in the definition to be an `or`. It doesn't mean, to me,  that all PRs labelled as minor are only editorial and without an issue.

CONTRIBUTING.md
rabi-abdel: fixed in: https://github.com/cntt-n/CNTT/pull/1430/commits/ac19eb494314516098c02e77958e3461f66700d4

CONTRIBUTING.md
rabi-abdel: @ASawwaf @walterkozlowski @pgoyal01 WSLs are expected to approve minor PRs as any other PR (the only difference is that minor PRs can be created without an issue), as @tomkivlin said, they can still have issue but not required.

In any case, if WSL feel the PR was not eligible to be considered minor, they can push back and force an issue to be created for it. There will be no accurate definition of minor and I am not very keen to try and define it as this will be a long discussion I feel. let us follow the common sense here and consider PRs on a case by case basis.

PR #1440, opened 2020-04-04
[RA1][RC1] Set Swift container_sync as optional

collivier:
It's not supported by radosgw which includes a different
synchronization model, multisite operations, for users who require
that feature [3].

From an interopability point of view it seems fine to set it as
optional. It should be noted that radosgw is used in both CNTT
reference implementations (Functest [2] and Airship [3]).

[1] https://access.redhat.com/documentation/en-us/red_hat_ceph_storage/3.2/html-single/release_notes/index
[2] http://artifacts.opnfv.org/functest/E46RCZLBRYCC/functest-opnfv-functest-smoke-hunter-tempest_slow-run-127/tempest_slow/tempest-report.html
[3] http://artifacts.opnfv.org/cntt/WKFRPQODPW34/cntt-opnfv-functest-smoke-latest-tempest_slow-run-37/tempest_slow/tempest-report.html

### DIFF
--- a/doc/ref_arch/openstack/chapters/chapter05.md
+++ b/doc/ref_arch/openstack/chapters/chapter05.md
@@ -94,7 +94,7 @@ REST API Version History: https://docs.openstack.org/cinder/latest/contributor/a
 | bulk_delete        | X             |
 | bulk_upload        | X             |
 | container_quotas   | X             |
-| container_sync     | X             |
+| container_sync     |               |
 | crossdomain        | X             |
 | discoverability    | X             |
 | form_post          | X             |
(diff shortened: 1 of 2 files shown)


iangardner22:
@pgoyal01  - I saw the agenda for todays RA-1 meeting and I am approving as this is not asking if Swift should be optional again but instead the sub-service -  **container_sync** which allows the content of one container to be kept in line with another.  This is of course very useful but assumes multi-site deployments if good resilience is expected.

rabi-abdel:
@fuqiao123 please review/approve

rabi-abdel:
@fuqiao123 once you approve this PR, it will be merged.

Issue #1475, opened 2020-04-15
[TSC] What do we need from other communities and when


rabi-abdel:
https://wiki.opnfv.org/display/DEV/Working+Progress+Page
Basic topics includes project allocation, release management, issue tracking. I think currently CNTT may consider these as OPNFV’s things, however without figuring out what CNTT requires, OPNFV can hardly make any movement. Taking release management as an example. CNTT currently release documents 4 times a year, however RI and RC release can be confusing since it is not bounded to a certain OPNFV release (OP

ASawwaf:
@rabi-abdel , we need to cover all other communitieSS , with different coverage for CNTT

for sure CNTT-OPNFV mapping/dependency is huge but on the other side, we need to have a wide view on others as we start with TIP


So this PR is good to have a clear interlocking between all SDO ( Input -Output )

rabi-abdel:
@ASawwaf Yes, I think we can tackling them one by one.

PR #1478, opened 2020-04-15
[TSC] Roadmap and what do we need from other communities and when

rabi-abdel:
Fixes https://github.com/cntt-n/CNTT/issues/1475


doc/tech/roadmap.md
```diff
@@ -35,21 +37,31 @@ A Release is:
   * An agreed common framework (template, criteria, best practice) for all projects
   * An unique release planning calendar with pre-defined milestones for each project
   * A vehicle to coordinate multiple projects and multiple type of projects (reference model and architecture, documentation, integration, packaging, deployment)
+
+<a name="bundle"></a>
+A Bundle is: A set of related specifications that are built to complement each other, specifically (RM -> RA -> RC -> RI).
```
pgoyal01, line 42: Do we need to clarify that bundle is a number such as "v1"

doc/tech/roadmap.md
```diff
@@ -35,21 +37,31 @@ A Release is:
...
+
+<a name="version"></a>
+A Version:
+* Each document has a version number attached to it that consist of **Bundle.Sequence**:
```
pgoyal01, line 46: typo: "consist" ---> "consists"

doc/tech/roadmap.md
```diff
@@ -35,21 +37,31 @@ A Release is:
...
+<a name="6.2.1"></a>
+### 6.2.1 High Level Roadmap
 <p align="center"><img src="./figures/cntt_roadmap.png" alt="scope" title="Roadmap" width="100%"/></p>
 <p align="center"><b>Figure 2:</b> CNTT Technical Specification Roadmap</p>
 
-<a name="2.2"></a>
-### Detailed Roadmap
+<a name="6.2.2"></a>
+### 6.2.2 Detailed Roadmap
 Please refer to individual [release plan](./release_notes/) and features for detailed roadmap.
```
pgoyal01, line 61: typo: "release plan"  ---> "release plans"

doc/tech/roadmap.md
```diff
@@ -78,24 +90,21 @@ The various workstreams in CNTT are
...
 - Containerized workloads, deployed on Kubernetes
 
 Each track follows the industry driven standards in the Reference Model as depicted in the below diagram.
 
-
-<p align="center"><img src="./figures/CNTT_Stable_3.0.png" alt="CNTT Stable 3.0 Portrait" title="CNTT Stable 3.0 Portrait
+<p align="center"><img src="./figures/tech_ws_dependencies.png" alt="CNTT WS Dependencies" title="CNTT WS Dependencies
 " width="100%"/>
- <p align="center"><b>Figure 3:</b> CNTT Stable 3.0 Portrait</p>
+ <p align="center"><b>Figure 3:</b> CNTT WS Dependencies</p>
```
pgoyal01, line 103: Figure 3, version numbers of RA, RI, RC etc are different from those in Figure 2. Figure 2 has the correct version numbers.

doc/tech/roadmap.md
```diff
@@ -78,24 +90,21 @@ The various workstreams in CNTT are
...
 
 
-<a name="4"></a>
-## Dependencies with Industry Communities
+<a name="6.4"></a>
+## 6.4 Dependencies with Industry Communities
 
 The CNTT is collaboratively working with other standard bodies within the Open Source industry. Some of the projects are
```
pgoyal01, line 109: These are not all standard bodies, also should it be "open source community"?

doc/tech/roadmap.md
```diff
@@ -104,11 +113,20 @@ The CNTT is collaboratively working with other standard bodies within the Open S
 - CNCF
 - ETSI NFV ISG
 
+<a name="6.4.1"></a>
+### 6.4.1 Dependencies with OPNFV and OVP
+
 The OpenStack based work stream in CNTT community works closely with OPNFV by leveraging and evolving OPNFV continuous integration pipeline with automation installer and testing capabilities. The Reference Implementation (RI1) has dependencies with other industry communities like OPNFV as shown [here](../ref_impl/cntt-ri/chapters/chapter01.md#1.3) and OVP as shown [here](../ref_impl/cntt-ri/chapters/chapter01.md#1.3). For information on the relationship to other communities, please read [Reference Implementation Chapter 01](../ref_impl/cntt-ri/chapters/chapter01.md#1.3). 
 
 <p align="center"><img src="./figures/Relation_to_other_communities.png" alt="Relation to other communities" title="Relation to other communities" width="60%"/></p>
```
pgoyal01, line 121: Should Figure 4 be generalised and just use RI instead of RI1?

doc/tech/roadmap.md
```diff
@@ -104,11 +113,20 @@ The CNTT is collaboratively working with other standard bodies within the Open S
...
 <p align="center"><b>Figure 4:</b> Relation to other communities.</p>
 
+<p align="center"><img src="./figures/tech_roadmap_cad.png" alt="Relation to other communities" title="Relation to other communities" width="100%"/></p>
+<p align="center"><b>Figure 5:</b> CNTT Roadmap relation to OPNFV and OVP.</p>
```
pgoyal01, line 125: Figure 5 -- same as Figure 3 -- defines Baldy and Baraque differently then Figure 2.

doc/tech/roadmap.md
rabi-abdel: fixed

doc/tech/roadmap.md
rabi-abdel: fixed

doc/tech/roadmap.md
rabi-abdel: fixed

doc/tech/roadmap.md
rabi-abdel: fixed

doc/tech/roadmap.md
rabi-abdel: well, OPNFV is only concerned with RI-1 for now. RI-2 is still to be determined.

doc/tech/roadmap.md
rabi-abdel: fixed

rabi-abdel:
@pgoyal01 @tomkivlin, @kedmison, @ASawwaf, @fuqiao123, @walterkozlowski, @xmulligan and @jayacha please review/approve

doc/tech/roadmap.md
```diff
@@ -78,37 +90,44 @@ The various workstreams in CNTT are
...
 
-The CNTT is collaboratively working with other standard bodies within the Open Source industry. Some of the projects are 
+The CNTT is collaboratively working with other standard bodies and open source communities such as: 
 - OpenStack
 - OPNFV
 - ONAP
 - CNCF
 - ETSI NFV ISG
+- TIP
+
```
ASawwaf, line 116: Telecom Infra Project (TIP)

rabi-abdel: done

pgoyal01:
@rabi-abdel the problem with the new numbering is that For Baldy, for example, with RM, RA-1 and RI-1 having the same version numbers seems to indicate that RA-1 reflects RM within Baldy but as you know that is not the case as RM v3.0 is still under development. If we want this level of alignment then we would need to stagger freezes of R< RA-1 etc. to allow times for the dependent specifications to reflect changes made in the predecessor specs.

rabi-abdel:
> @rabi-abdel the problem with the new numbering is that For Baldy, for example, with RM, RA-1 and RI-1 having the same version numbers seems to indicate that RA-1 reflects RM within Baldy but as you know that is not the case as RM v3.0 is still under development. If we want this level of alignment then we would need to stagger freezes of R< RA-1 etc. to allow times for the dependent specifications to reflect changes made in the predecessor specs.

@pgoyal01 the new versioning scheme is proposed to solve your exact concern. we will leave it up to each spec to decide wither it can align with the progression of RM on the same release or use the stable branch instead. as RM progresses, it gets easier but I understand your concern and this is to improve it.

collivier:
This change make lots of side effects even outside of CNTT.
We can hardly modify versions closed to the Baldy release even more without a global #agree.

It's worth mentioning that CNTT RC 1 as a good example is designed to be reused by Operators, OpenStack integrators, VNF Vendors, etc. and then be mentioned in RFPs.

Was it widely considered by all CNTT members? I would have voted -1 due its side effects closed to the release. I would agree reverting this PR.

rabi-abdel:
@collivier I don't understand your concern. it has been discussed in TSM, it has been discussed in vConf and it has been put in place to solve the challenge you just highlighted. I think it will simplify a lot our deliveries. it is possible you might miss-understand the intent. this allows RC-1 to decide what bundle it is related to.

rabi-abdel:
> @rabi-abdel @walterkozlowski @ASawwaf
> This change makes lots of side effects even outside of CNTT.
> We can hardly modify versions closed to the Baldy release even more without a global #agree.
> 
> It's worth mentioning that CNTT RC 1 as a good example is designed to be reused by Operators, OpenStack integrators, VNF Vendors, etc. and then be mentioned in RFPs (it already solves a current API problem).
> 
> Was it widely considered by all CNTT members? I would have voted -1 due its side effects closed to the release. I would agree reverting this PR.

@collivier is your concern around RC-1 being referenced in different locations as (ver 1.0) and this change will break that? This PR was merged two weeks ago and was also discussed in Governance call twice. I don't know what more consensus we can do around it! there was no objection to do it and was explained at least 5 times.

collivier:
@rabi-abdel : I did miss the vConf slot and possibly the previous discussions on this particular in TS. My point was not here and it could be my fault here even if I can participate to all meeting asked by CNTT and OPNFV. Switching to 1.0 to 3.0 as it was proposed before ask tor an update internally and externally. It should be manageable but we may care about this for the next releases.

rabi-abdel:
Thanks @collivier there is a discussion on email of which @karinesevilla and @pgoyal01 suggesting to make RA-1 based on RM2.0 so it will follow it's version which means RC-1 will have to follow RA-1 and be v2.0 now. let me know if that is a problem. (or reply to the email

collivier:
Let's try this model and we will see if they are external side effects. RC1 don't cover everything and it will continuously grow but RC1 is fully designed/tuned for the current RA1 **asis** (or any mismatch may be considered as a bug in RC1).
Thank you for all clarifications!

PR #1479, opened 2020-04-15
[GOV]: Handling of images

CsatariGergely:


artefacts/README.md
```diff
@@ -1 +1,16 @@
...
+
+* [1 Usage of figures](#1)
+
+<a name="1"></a>
+## 1 Usage of figures
+
+- Figures used in the document should be in Portable Network Graphic (PNG) format
+- Figures must be stored in the `figures` folder of the document where the figures are used (e.g.: `doc/ref_arch/figures/`)
+- File name of the figures shold contain the chapter number where the figure is used as a prefix
+- Every figure used in the documents should have an editable version commited to the repository in `ppt(x)` or in `svg` format
```
kedmison, line 14: If we're adding svg support, can Github support the display of svg natively (and thus not need to have a .png format generated separately from the artifacts?)

artefacts/README.md
```diff
@@ -1 +1,16 @@
...
+- Editable versions should be stored in the `figures/artefacts` folder of the document where the figures are used (e.g.: `doc/ref_arch/figures/artefacts`)
```
kedmison, line 15: .svg starts to cross the boundaries between editable artifact and displayable artifact.  Should we place the viewable artifact and the editable artifact in the same directory, in preparation for when the SVG format converges editable/viewable into a single artifact?

artefacts/README.md
```diff
@@ -1 +1,16 @@
...
 
+## Table of Contents
+
+* [1 Usage of figures](#1)
+
+<a name="1"></a>
+## 1 Usage of figures
+
+- Figures used in the document should be in Portable Network Graphic (PNG) format
+- Figures must be stored in the `figures` folder of the document where the figures are used (e.g.: `doc/ref_arch/figures/`)
```
pgoyal01, line 12: Can you use a better example such as '/doc/ref_arch/openstack/figures' or 'doc/ref_model/figures/'?

artefacts/README.md
```diff
@@ -1 +1,16 @@
...
+- File name of the figures shold contain the chapter number where the figure is used as a prefix
```
pgoyal01, line 13: Just the chapter number as in "4" or as in "Ch04"?
In case we decide to reorganise the Figures would it be better if we asked to include the WS and chapter as a prefix. What about the Figure legend? should it also contain the Figure Number?

<WS acronym (do not include "-")>><"_Ch"><2-digit chapter number><"_"><figure legend (spaces replaced by "_")><"_Figure_"><figure number with "." or "-" replaced by "_">

Example, "RM_Ch04_Cloud_Infra_Figure_4_2"; "RA2_Ch04_k8s_architecture_Figure_4_1"

artefacts/README.md
pgoyal01, line 15: Can we keep the artefacts directory at the same level as the figures? Artefacts may include items other than the figures.

CsatariGergely: My worry is that the figures folder becomes overloaded by the editable files. I would add the svg files in the figures folder if they are directly used while keep the editables in the artifacts folder.

artefacts/README.md
CsatariGergely: I would not use the figure number, as they change too often ;)

CsatariGergely:
> For editable figures I would stick to ppt format at this stage, until we clarify how github will handle this. Also, ppt will stay for a while as the most common. Why do we need svg at this time?
> I agree with @pgoyal01 for the need of a more precise naming convention for figures. His proposal seems to be good.
> The hierarchy and the names of the folders seems to be strange. Looks like "artefacts" are meant for editable documents (including editable versions of figures). This will be confusing as many people will look for editable figures under "figures". I agree with @pgoyal01 that "figures" should be on the same level as "artefacts". Even more, under "figures" we should have two underfolders "displayable figures" and "editable figures" (or equivalent names). This will be much clearer.

- on ppt(x) vs svg I have strong opinion. What do other think?
- file naming conventions updated
- @walterkozlowski your proposal is different from @pgoyal01 . One alternative to store the editable files is `ws_name/figures/editable-figures` while the other is `ws_name/artefacts`. I would not move the displayed figures from `ws_name/figures/`.

artefacts/README.md
CsatariGergely: Hopefully corrected, but please check @walterkozlowski -s comment and my question about removing the svg support.

artefacts/README.md
CsatariGergely: Done, thanks for the examples.

artefacts/README.md
CsatariGergely: Okay for me, but please check @walterkozlowski -s comment and my question on the topic.

kedmison: Ok, so I would interpret your proposal as:
- for SVG files, the only copy is stored in the figures folder, per-chapter, as that is both the editable and the viewable version of the figure
- for PNG files, the viewable form is stored in the figures folder, per-chapter, and the editable copy is stored in the single shared artefacts folder.

Am I understanding properly?

artefacts/README.md
kedmison: Giving the pictures names based on content rather than on where they are located would be good.  It's easy enough to search the repo for a filename to find out which chapters an image is referenced in.

artefacts/README.md
walterkozlowski: My suggestion as a compromise (reg the location):
Figures must be stored in the `figures` folder of the document where the figures are used ( `doc/ws_name/figures/`)
Editable versions should be stored in the `figures/editable` folder of the document where the figures are used  `doc/ws_name/figures/editable`)
This will provide a clear direction where the editable versions are stored.
@CsatariGergely, it is clear how the above would work for editable ppt and display png files.  It seems some confusion what is editable and display version for svg format, see @kedmison 's comments.  If this can be clarified, then i do not have anything against svg format.  Can you clarify please?

artefacts/README.md
```diff
@@ -1 +1,16 @@
...
+## Table of Contents
+
+* [1 Usage of figures](#1)
+
+<a name="1"></a>
+## 1 Usage of figures
+
+- Figures used in the document should be in Portable Network Graphic (`png`) or in Scalable Vector Graphics (`svg`) format
+- Figures must be stored in the `figures` folder of the document where the figures are used (e.g.: `doc/ref_model/figures/`or `'doc/ref_arch/openstack/figures`)
+- File name of the figures shold contain the Work Stream and the chapter number where the figure is used as a prefix (e.g.: `ra2-ch02-topic-of-the-figure`)
```
pgoyal01, line 13: Are hyphens allowed in file names? I think we have to use underscores, but I maybe wrong.

rabi-abdel:
@CsatariGergely is it possible to address comments above?

artefacts/README.md
rabi-abdel: @pgoyal01 we didn't specify wither to use underscore or dash. I think either will do.

artefacts/README.md
rabi-abdel: @kedmison this is what I understood. if .svg file is used directly in the md then it is placed in figures folder, if it is the editable version of a png file that is used, then it will be placed in the artefacts.

rabi-abdel:
@walterkozlowski @kedmison can you re-review/approve please

PR #1449, opened 2020-04-05
[Edge WS ] Initial Content

ASawwaf:
@Karimrabie @petorre @hdamker @sukhdevkapur 
Closes https://github.com/cntt-n/CNTT/issues/1415

PR #1490, opened 2020-04-19
[Edge WS] Edge_use_cases_1_2_3_Update usecases.md

Karimrabie:
This is to review and agree on the structure and the content for use cases 1, 2, & 3.

Once agreed. I can put the content for use cases 4 & 5. (vRAN & Remote Surveillance with Video Analytics)

Note - Can I assign to "Team" instead of individual accounts?

### DIFF
--- a/doc/tech/usecases.md
+++ b/doc/tech/usecases.md
@@ -55,23 +55,63 @@
   
      <img width="1079" alt="Screenshot 2020-04-14 at 8 37 57 PM" src="https://user-images.githubusercontent.com/25845305/79390856-c536f780-7f70-11ea-88d5-cfd89cf5121a.png">
 
-- **Use Case #2**
+- **Use Case #2 - Private 5G with Network Slicing**
 
   - **Business Objectives**
+  
+      Private 5G is considered one of the most anticipated Business use cases in the coming few years enabling Mobile Operators to provide a standalone private Mobile Network to enterprises that may include all the ingredients of PLMN such as Radio, Core, Infrastructure & Services covering the business requirements in terms of security, performance, reliability, & availability.
 
   - **Targeted Segments**
+  
+      - Governmental Sectors & Public Safety (Mission critical applications)
+      - Factories and Industry sector.
+      - Enterprises with Business-critical applications. 
+      - Enterprises with strict security requirements with respect to assets reachability.
+      - Enterprises with strict KPIs requirements that mandate the on-premise deployment.
 
   - **Architecture**
 
+      <img width="1089" alt="Screenshot 2020-04-16 at 6 22 58 PM" src="https://user-images.githubusercontent.com/25845305/79684936-635ae400-8235-11ea-9c97-4774aba24d17.png">
+      
+        Note (1) - There are multiple flavours for Private 5G deployments or NPN, Non-Public Network as defined by 3GPP.
+        
+        Note (2) - The use case addresses the technical realization of NPN as a Network Slice of a PLMN as per Annex D – 3GPP TS 23.501 R16 and not covering the other scenarios of deployment.
+        
+        Note (3) - The use case assumes a network slice that is constructed from a single UPF deployed on Customer premises while sharing the 5G Control Plane (AMF, SMF, & other CP Network Functions) with the PLMN. 
+        
+        Note (4) - The use case doesn’t cover the requirements of the private Application Servers (ASs) as they may vary with each customer setup.
+        
+        Note (5) - Hosting the CU/DU on-Customer Infrastructure depends on the enterprise offering by the Mobile Operator and the selected Private 5G setup.
+        
+        Note (6) – The Cloud Infrastructure can be governed by the client or handled by the Service Provider (Mobile Operator) as part of Managed-services model.
 
-- **Use Case #3**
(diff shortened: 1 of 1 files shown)


TFredberg:
I think the structure in general is OK.
I miss a sub chapter of main Edge Specific Requirements and think a 5.1.6 would be a good place (shifting current 5.1.6+ down). 
I don´t think it should be a complete Requirement Specification but rather a few vital requirements that makes this use case requirement stand out to be different from the "normal"/"simpler" centralized deployed applications/workloads. 
I think the requirements should be making up a list e.g. with each Use Case as a column, each requirement as a row and the cells in a row either having a check box for each Use Case or in some other requirements a characteristics value (i.e. the row have a "hard typed" response) .
As an example, it could be something like:
- Infrastructure and VNF/CNF resiliency (redundancy) required on Edge site as opposed to functions being redundant on network level (from another Edge site): Yes/No checkbox
- Max latency (or Round-Trip Delay): x mS

Personably I have never heard about SDO before I read it here, and would think the chapter heading might need some little explanation rather than the current cryptic title (at least for me).

doc/tech/usecases.md
```diff
@@ -55,23 +55,63 @@
...
   - **Targeted Segments**
 
+      - Automotive Industry.
+      - Governmental Departments (Smart Cities, Transport, Police, Emergency Services, etc). 
+      - Private residencies (Compounds, Hotels and Resorts).
+      - Enterprise and Industrial Campuses.
+
   - **Architecture**
 
+    <img width="1006" alt="Screenshot 2020-04-19 at 12 17 43 PM" src="https://user-images.githubusercontent.com/25845305/79685269-c8afd480-8237-11ea-90d5-0e7b0edc684e.png">
```
CsatariGergely, line 103: Please add also the editable source of the figure.

Karimrabie: > I think the structure in general is OK.
> I miss a sub chapter of main Edge Specific Requirements and think a 5.1.6 would be a good place (shifting current 5.1.6+ down).
> I don´t think it should be a complete Requirement Specification but rather a few vital requirements that makes this use case requirement stand out to be different from the "normal"/"simpler" centralized deployed applications/workloads.
> I think the requirements should be making up a list e.g. with each Use Case as a column, each requirement as a row and the cells in a row either having a check box for each Use Case or in some other requirements a characteristics value (i.e. the row have a "hard typed" response) .
> As an example, it could be something like:
> 
> * Infrastructure and VNF/CNF resiliency (redundancy) required on Edge site as opposed to functions being redundant on network level (from another Edge site): Yes/No checkbox
> * Max latency (or Round-Trip Delay): x mS
> 
> Personably I have never heard about SDO before I read it here, and would think the chapter heading might need some little explanation rather than the current cryptic title (at least for me).

I believe it is a good idea to group the requirements (_no so detailed_) in a table with each use case as a column. I already captured the requirements for these use cases in terms of latency so we have something to start with.

For the second point, Does SDO refer to Standards Development Organisation? Unfortunately, I didn't understand the point. Could you please give an example?

Karimrabie: > Please add also the editable source of the figure.

I will upload the full slides soon with the use cases architecture. 
I just need some guidance on where to upload it.

TFredberg: @Karimrabie Yes, it was just the articulation of spelling out SDO as Standards Development Organisation to improve readability for us that are nut fluent in the acronym-soup :-)

Karimrabie: @TFredberg Ah now I get it :) I will add the acronym mapping.

Karimrabie:
Dears,

Are there any final comments about the content of use cases section (1,2, & 3). I will be proceeding with use cases 4, &5 following the same structure.

@CsatariGergely I will upload the editable files once finish writing the full content
@TFredberg I will update the acronym mapping at the end of the chapter. For the suggestion of building the table, could you please take the lead in this?

[https://wiki.lfnetworking.org/display/LN/2020-04-07+-+%5BCNTT+EDGE%5D+-+Meeting+Agenda+and+Minutes](url)

@ASawwaf @rabi-abdel @bfcohen @ildikov 

Thanks to approve.

ASawwaf:
Thanks @Karimrabie  , my view on the first 3 use cases, that UPF , vCU and vDU is the most common VNF/CNF that will be hosted in the edge , so i think next step we need to check if these 3 functions need special requirements/profiles or not, which i belive it need :)


So I am ok with what you define for use case 1,2,3 

@TFredberg , Totaly agree and as @Karimrabie mentioned one the agreed point the last meeting 
so IMO, let focus on use cases in this PR , and it will great if you can open a new PR for this common requirements

rabi-abdel:
@ASawwaf, @Karimrabie this PR is marked as WIP, if you want it to be merged, please remove WIP label.

rabi-abdel:
@ASawwaf This PR is trying to merge the content into your own branch. it needs to be updated to merge into master. (otherwise content will not appear in master)

rabi-abdel:
Merging this PR will move the changes into https://github.com/cntt-n/CNTT/pull/1449 so I will do that as the easier solution.

PR #1517, opened 2020-04-28
[RC1] Remove falsy duplicated req tabs

collivier:
They are outdated and then falsy duplicate RM and RA tabs.

### DIFF
--- a/doc/ref_cert/lfn/chapters/chapter03.md
+++ b/doc/ref_cert/lfn/chapters/chapter03.md
@@ -49,93 +49,6 @@ The RTM contains RM config (i.e. .conf) requirements listed “per profile”, f
 
 For detailed information on RM & RA-1 NFVI and VNF requirements, please refer to [RI-1 Chapter 3](https://github.com/cntt-n/CNTT/blob/master/doc/ref_impl/cntt-ri/chapters/chapter03.md).
 
-|#|Requirement|Description|Catalog|Status|MUST|Category|RI-1 Implemented (Yes/No)|RC-1 Coverage (project)|RC-1 Project To On-board|
-|--- |--- |--- |--- |--- |--- |--- |--- |--- |--- |
-|1|req.gen.ost.01|must use OpenStack APIs.|General|in Functest|Must|Functional||Functest||
-|2|req.gen.ost.02|must support dynamic request and configuration of virtual resources through APIs.|General|in Functest|Must|Functional||Functest||
-|3|req.gen.cnt.01|should consist of stateless service components. However, where state is required it must be kept external to the components.|General|||||||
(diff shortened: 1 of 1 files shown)


collivier:
As discussed in RA1 meetings.

pgoyal01:
Cedric, Need @rabi-abdel and @fuqiao123 to weigh in. I agree that these requirements are obsolete as they are not aligned with the RA-1 requirements (some are and some are not).

The question is should RC duplicate the RA-1 requirements? If yes, we will need to copy the requirements from RA-1 Ch02 as I don't know of any way in md to reproduce content by reference.

fuqiao1:
> Cedric, Need @rabi-abdel and @fuqiao123 to weigh in. I agree that these requirements are obsolete as they are not aligned with the RA-1 requirements (some are and some are not).
> 
> The question is should RC duplicate the RA-1 requirements? If yes, we will need to copy the requirements from RA-1 Ch02 as I don't know of any way in md to reproduce content by reference.

I understand the table are not correct with the current RA-1 is because these tables copied RI-1 chp3, and those tables in RI-1 were writen months ago and were basicaly copy content in RA-1 back at that time. I don't think it make sense actually for RI and RC to copy these tables and make it difficult to manage these tables in different WS. In the meantime, these tables can hardly translate in to real test cases. If the TSC care about tracebility of RA requirement in RC, I think it is important that tables in RA translate into exact cases/requirement bullets for RC to check.

rabi-abdel:
@fuqiao1 @collivier @pgoyal01 let us discuss this quickly in the TSC call today.

collivier:
Here we should follow the code development principles.
When code is copied/pasted, we do factorize. When the duplicated code is even falsy, it's become a key issue. Here it should be noted that RC isn't the right source (it's RM and RA).

rabi-abdel:
@rabi-abdel to work with Cedric and fu Qiao to finalise this

collivier:
@rabi-abdel @fuqiao123 @fuqiao1 As discussed in TSC: They are false. They must be removed first and then enhanced. We can't block this PR till the upcoming release.

rabi-abdel:
@collivier I commented them out rather than removing them completely! will merge by end of day. let me know what you think before then,

collivier:
@rabi-abdel I fully disagree to only comment the tabs here because it doesn't fix the bugs.
I don't see why we selected a disruptive approach (they are now still in md files) without CNTT RC lead and Co-lead asking for it and this way hasn't been discussed in TSC. I would appreciate quick clarifications about the role of CNTT RC lead/co-lead.

collivier:
@rabi-abdel In fact it's even worse as it refers to Functest without any Functest contributor involved.
Commenting this tab doesn't fix the bug and more important doesn't match CNTT role and OPNFV meritocraty. No CNTT role allows speaking on behalf of Functest without contributing.

PR #1518, opened 2020-04-28
[RM-CH08] Remove CH08 from the ToC & re-numbering

ASawwaf:
After CH08 moved out , Remove CH08 from the ToC & re-numbering


rabi-abdel:
This change require renaming file names, headings, etc. :) it might be better if we do that during proofreading milestone.

markshostak:
@ASawwaf Hi Ahmed, FYI, RMCH-08 is going to remain in the RM, so at this time we shouldn't have to do any renaming or renumbering. I'd suggest we label this PR "Returned" and close. If things change in the future, we can re/open a new one.  Is that ok? Thanks, -Mark

rabi-abdel:
I think the best way forward is to Close this PR.

PR #1523, opened 2020-04-29
[RA2 Ch4]: Fixing the CNI Multiplexer comparision table

CsatariGergely:
Closes #1522

PR #1524, opened 2020-04-29
[RA2 Ch02] - restructure of RA2 chapters

tomkivlin:
Resolves #1018 
Resolves #1112 
Resolves #1137 
Resolves #1338

PR #1532, opened 2020-04-29
[RA2 Ch4]: Implementing RA2 reorganisation to Ch4

### LINKED ISSUE #1112: [RA2 Ch2] - review CNI and other networking requirements
Question, why is a CNI multiplexer being made a hard requirement as opposed to just setting a requirement to support multiple network interfaces?

> "Because req.inf.ntw.01 requires the architecture to support CNI and req.inf.ntw.16 requires the capability to attach several network interfaces to the pods the architecture must support a CNI metaplugin/CNI multiplexer."

### LINKED ISSUE #1499: [RA-2 Ch04] 4.5 CNI Plugins
[4.5 CNI Plugins](https://github.com/cntt-n/CNTT/blob/master/doc/ref_arch/kubernetes/chapters/chapter04.md#45-cni-plugins) refers to DANM for meta plugin in the text implicitly implying it to be the preferred meta plugin. I don't believe that to be the intent but we should correct any perceptions.

IMO, from a requirements perspective, we should only evaluate features that should be performed by a meta plugin; not compare extraneous features that a particular meta plugin may or may not support. 

With that in mind may I suggest that we evaluate the need for the following requirements to be part of a meta-plugin evaluation:

| Requirement | Requirement Description |
|----------|-----------|
| req.inf.ntw.02 | The Architecture must support intra-node communications, such as between agents on a node and all pods on that node |
| req.inf.ntw.03 | The Architecture must support inter-node communications without NAT, such as communications between pods on a node with all other pods on all nodes |
| req.inf.ntw.04 | The Architecture must support low latency and high throughput traffic needs. |
| req.inf.ntw.10 | The networking solution should be configurable in an automated manner using openly published API definitions. |
| req.inf.ntw.11 | The networking solution should be able to be centrally administrated and configured. |
| req.inf.ntw.14 | The Architecture must support capabilities for integrating SDN controllers. |
| req.inf.ntw.15 | The Architecture should support Service Mesh Interface (SMI). |
|  | Cluster wide IP address management |

CsatariGergely:
This change implement the changes of Ch02 by #1524.

Closes #1523
Closes #1112
Closes #1499

### DIFF
--- a/doc/ref_arch/kubernetes/chapters/chapter04.md
+++ b/doc/ref_arch/kubernetes/chapters/chapter04.md
@@ -30,7 +30,7 @@ Figure 4-1 below shows the architectural components that are described in the su
 In order for a Host OS to be conformant with this Reference Architecture it must meet the following requirements:
 - A version of the Linux kernel that is [compatible with kubeadm](https://kubernetes.io/docs/reference/setup-tools/kubeadm/implementation-details/#kubeadm-init-workflow-internal-design) - this has been chosen as the baseline because kubeadm is focussed on installing and managing the lifecycle of Kubernetes and nothing else, hence it is easily integrated into higher-level and more complete tooling for the full lifecycle management of the infrastructure, cluster add-ons, etc.
 - Windows Server 2019 (this can be used for worker nodes, but be aware of the [limitations](https://kubernetes.io/docs/setup/production-environment/windows/intro-windows-in-kubernetes/#limitations)).
-- In order to support `req.gen.cnt.03` (immutable infrastructure), the Host OS must be easily reproduced, consistent, disposable, will have a repeatable deployment process, and will not have configuration or artifacts that are modifiable in place (i.e. once it is running).
+- In order to support `req.gen.cnt.02` (immutable infrastructure), the Host OS must be easily reproduced, consistent, disposable, will have a repeatable deployment process, and will not have configuration or artifacts that are modifiable in place (i.e. once it is running).
 - The selection of Host OS shall not restrict the selection of the OS used to build container images (container base image).
 
 Table 4-1 lists the Linux kernel versions that comply with this Reference Architecture specification.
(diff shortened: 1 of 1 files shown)


pgoyal01:
Given that we are making changes to RA-2 4.5, IMO, the issue #1499 needs to be addressed.

The gist of Issue #1499 is that if we are discussing meta-plugin then we should only consider requirements applicable to meta-plugin -- multiple interface and multiple CNI support.

4.5 states, "The selected CNI multiplexer/metapulgin may be DANM as it provides the possibility to use several other CNI plugins .." -- I think that since as a meta-plugin MULTUS also meets all of the needs (is also widely used), we should include it in the text as: "The selected CNI multiplexer/metapulgin may be MULTUS or  DANM as they provide the possibility to use several other CNI plugins ..."

w.r.t. the note, "1): Under implementation in the current release." -- I think the table should only reference what exists. If not, we would need to review the MULTUS github. (there is a typo in the current note -- "inthe")

CsatariGergely:
@pgoyal01 

> Given that we are making changes to RA-2 4.5, IMO, the issue #1499 needs to be addressed.

I agree, this pr should also close #1499 

> The gist of Issue #1499 is that if we are discussing meta-plugin then we should only consider requirements applicable to meta-plugin -- multiple interface and multiple CNI support.

The problem is that the metaplugin affects all other networking features in Kubernetes, this is why I've added all relevant requirements to this table. 

> 4.5 states, "The selected CNI multiplexer/metapulgin may be DANM as it provides the possibility to use several other CNI plugins .." -- I think that since as a meta-plugin MULTUS also meets all of the needs (is also widely used), we should include it in the text as: "The selected CNI multiplexer/metapulgin may be MULTUS or DANM as they provide the possibility to use several other CNI plugins ..."

The purpose of Ch4 to select one metaplugin as the related Pod annotations are different. We should avoid the situation when CNF-s need to support two Pod manifests for the two options we define here. 

> w.r.t. the note, "1): Under implementation in the current release." -- I think the table should only reference what exists. If not, we would need to review the MULTUS github. (there is a typo in the current note -- "inthe")

The table itself references only to what exists. It is a feature marked as "Not supported" by both metaplugins. I can remove the note if you think that it is a non-relevant information. (and that would also solve the typo problem)

CsatariGergely:
Closes #1499

pgoyal01:
@CsatariGergely Thanks for the explanation but the fact remains that many of these requirements are met by other components and these don't necessarily must be supported by the selected meta-plugin.

w.r.t. the selection of a single meta plugin if that is the desire then we need to ask all of the operators to be engaged and asked for what they think.

CsatariGergely:
> @CsatariGergely Thanks for the explanation but the fact remains that many of these requirements are met by other components and these don't necessarily must be supported by the selected meta-plugin.

Yes, this is why the table contains some "Supported via another CNI plugin" entries, but I still believe, that this is a better approach than just cherry-pick requirements from Ch2.
 
> w.r.t. the selection of a single meta plugin if that is the desire then we need to ask all of the operators to be engaged and asked for what they think.

This is what CNTT is about, to standardize the cloud infrastructure to reduce the integration cost. The main content of this chapter did not change in the last release, so we can not say that operators (and vendors) had no chance to review it.

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -84,50 +84,51 @@ feature-gates:
...
 
 > Todo: details and RA2 specifications relating to runtimes in order to meet RM features and requirements from RM chapters 4 and 5.
 
 <a name="4.5"></a>
 ## 4.5 CNI plugins
 
-> Editors note: The following chapter lists a set of CNI plugins conformant with the Reference Architecture. In future releases the list of CNI plugins should be refined in a way that there is only one example component selected for each functionality. 
-
-The selected CNI multiplexer/metapulgin may be [DANM](https://github.com/nokia/danm) as it provides the possibility to use several other CNI plugins (`req.inf.ntw.16`) and provides an API based solution to administer the networks (`req.inf.ntw.10`) from a central point (`req.inf.ntw.11`).<br>
+The selected CNI multiplexer/metapulgin may be [DANM](https://github.com/nokia/danm) as it provides the possibility to use several other CNI plugins (`req.inf.ntw.06`) and provides an API based solution to administer the networks from a central point (`req.inf.ntw.03`).<br>
```
taylor, line 105: If the API is required by a CNI plugin, I suggest the language be updated to be more neutral:

> The selected CNI multiplexer/metapulgin should support other CNI plugins (req.inf.ntw.06) and provide an API based solution to administer the networks from a central point (req.inf.ntw.03).

Otherwise make the management API be optional for a CNI:

>  The selected CNI multiplexer/metapulgin MUST support other CNI plugins (req.inf.ntw.06) and MAY provide an API based solution to administer the networks from a central point (req.inf.ntw.03).

tomkivlin:
Concerns about suggesting one or the other. 
Action Point: discuss the approach in https://github.com/cntt-n/CNTT/issues/1499 to how to resolve the Multus v DANM debate (suggestion: describe the use-cases, create more requirements if we need, make the decision clear)
Action Point: create issue to develop content in "Gaps" chapter re. the ability to manage a multiplexer via a consistent mechanism / API

doc/ref_arch/kubernetes/chapters/chapter04.md
CsatariGergely: Thanks I've updated the wording based on this.

tomkivlin: @taylor @pgoyal01 - could one/both of you review and approve the PR if you're happy (given the points made in the call about going into more detail and potentially changing the recommendation in future iterations). Thanks!

pgoyal01:
The new language states:
As the selected CNI multiplexer/metapulgin MUST support other CNI plugins (req.inf.ntw.06) and SHOULD provide an API based solution to administer the networks from a central point (req.inf.ntw.03) the selected CNI multiplexer/metapulgin may be DANM.

The MUST requirement is met by both MULTUS and DANM -- the above language seems to indicate that is not the case.

rabi-abdel:
is this to be merged for Baldy? @CsatariGergely

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -84,50 +84,51 @@ feature-gates:
...
 
 > Todo: details and RA2 specifications relating to runtimes in order to meet RM features and requirements from RM chapters 4 and 5.
 
 <a name="4.5"></a>
 ## 4.5 CNI plugins
 
-> Editors note: The following chapter lists a set of CNI plugins conformant with the Reference Architecture. In future releases the list of CNI plugins should be refined in a way that there is only one example component selected for each functionality. 
-
-The selected CNI multiplexer/metapulgin may be [DANM](https://github.com/nokia/danm) as it provides the possibility to use several other CNI plugins (`req.inf.ntw.16`) and provides an API based solution to administer the networks (`req.inf.ntw.10`) from a central point (`req.inf.ntw.11`).<br>
+As the selected CNI multiplexer/metapulgin MUST support other CNI plugins (`req.inf.ntw.06`) and SHOULD provide an API based solution to administer the networks from a central point (`req.inf.ntw.03`) the selected CNI multiplexer/metapulgin may be [DANM](https://github.com/nokia/danm).<br>
```
taylor, line 105: Suggested update to wording:

> The selected CNI multiplexer/metapulgin MUST be conformant with the [CNI spec](https://github.com/containernetworking/cni/blob/master/SPEC.md), MUST support other CNI plugins (`req.inf.ntw.06`), and SHOULD provide an API based solution to administer the networks from a central point (`req.inf.ntw.03`).  An example CNI implementation which meets these requirements is [DANM](https://github.com/nokia/danm).

taylor: Alternative wording:

The selected CNI multiplexer/metapulgin MUST be conformant with the [CNI spec](https://github.com/containernetworking/cni/blob/master/SPEC.md), MUST support other CNI plugins (`req.inf.ntw.06`), and SHOULD provide an API based solution to administer the networks from a central point (`req.inf.ntw.03`).  Example CNI implementations which meet these requirements include:
- [DANM](https://github.com/nokia/danm/#danm)
- [Multus](https://github.com/intel/multus-cni#multus-cni)

taylor: It may make sense to have a section discussing the APIs which are provided by additional items outside of the CNI. Even in the DANM project the API support is not part of the CNI plugin itself it is additional components (eg. netwatcher, svcwatcher.  see [multus doc](https://github.com/nokia/danm/#install-an-akraino-rec-and-get-danm-for-free)  for more)-- though I'm not clear if the DANM CNI plugin can be used w/o the other components independently.

pgoyal01: @taylor Thanks. Also, if we look at the direction stated in the [DANM documen](https://github.com/nokia/danm/#install-an-akraino-rec-and-get-danm-for-free)t, "Fakeipam is a temporary solution, the long-term aim is to separate DANM's IPAM component into a full-fledged, standalone IPAM solution."

PR #1547, opened 2020-04-30
[RI-1 Ch06] Change the requirement level from must to may in req.gen.ins.01

jgu17:
Based on the feedback, most installers except one is not interested to support another format of descript files beyongs its own native format. The one that supports PDF incurred loss of functionality. Openstack Airship community also has no interest to support PDF. The estimate from Sridhar to implement such conversion tool is 3 developers/2months for the initial implementation based on Airship Treasuremap 1.7, then needs a redo when AS 2.0 is released late this year. PDF/IDF also needs to be drastically expanded to support the Airship supported configurations. So after discussion with Sridhar and Al, I am proposing to change this requirement from "must" to "may".

### DIFF
--- a/doc/ref_impl/cntt-ri/chapters/chapter06.md
+++ b/doc/ref_impl/cntt-ri/chapters/chapter06.md
@@ -47,7 +47,7 @@ Thanks to the descriptor file, the NFVi infrastructure deployment could be compl
 
 | Ref #            | sub-category | Description                                                                              |
 |------------------|--------------|------------------------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


acmacm:
I suggest to change **should** to **may**, the chapter intro only allows **may* for options. We have discussed the work estimate to meet this requirement - on the order of months...

jgu17:
@rrajesh70 @opensource-tnbt @nickchase Can you please review and approve by May 1st, noon PST ? This is urgent. Thanks!

rabi-abdel:
@jgu17 can you please send them a direct email requesting them to do that please.

rabiabdel:
one more approval needed for it to be merged for baldy

PR #1553, opened 2020-05-01
[RC1] Exclude Octavia testing in tempest_heat

collivier:
Octavia is optional from the time being.
It has to be over protected even if tempest config may apply
the exclusion.


rabi-abdel:
this will be merged by end of today to make it to baldy if no further input received

PR #1559, opened 2020-05-05
[RC] Enable test_network_v6 (linked to req.inf.ntw.16)

### LINKED ISSUE #1556: [RA1][RC1] mismatch about IPv6
I think there is a mismatch between RA and RC about IPv6.
A few IPv6 tests are currently skipped in RC and I think I do rather enable it in Functest (CNTT Compliance) and RC.

https://github.com/cntt-n/CNTT/blob/master/doc/ref_arch/openstack/chapters/chapter02.md
| `req.inf.ntw.16` | Network | The Architecture **must** support dual stack IPv4 and IPv6 for tenant networks and workloads.| |

https://github.com/cntt-n/CNTT/blob/master/doc/ref_cert/lfn/chapters/chapter03.md
.*test_network_v6 is excluded as they were considered as out of CNTT RC.

Do you agree enabling all the following tests?
tempest.scenario.test_network_v6.TestGettingAddress.test_dhcp6_stateless_from_os         
tempest.scenario.test_network_v6.TestGettingAddress.test_dualnet_dhcp6_stateless_from_os      
tempest.scenario.test_network_v6.TestGettingAddress.test_dualnet_multi_prefix_dhcpv6_stateless            
tempest.scenario.test_network_v6.TestGettingAddress.test_dualnet_multi_prefix_slaac      
tempest.scenario.test_network_v6.TestGettingAddress.test_dualnet_slaac_from_os              
tempest.scenario.test_network_v6.TestGettingAddress.test_multi_prefix_dhcpv6_stateless              
tempest.scenario.test_network_v6.TestGettingAddress.test_multi_prefix_slaac         
tempest.scenario.test_network_v6.TestGettingAddress.test_slaac_from_os

They mostly ping6 the router via a VM (the initial ssh connection is done via IPv4) which gets its IPv6 in all combinations.
I know the CNTT requirement is quite unclear compared to the tempest tests but I would say nothing specific here (and it’s not the worst case: see booting VM).

They all pass in IaaS verification. It’s only about adding them in RC to cover this “MUST” requirement (I’m considering this scenario if no quick feedback).
If req.inf.ntw.16 becomes “MAY” or “SHOULD” requirement, RC will be fine as these tests would remain out CNTT RC (IaaS verification).
http://artifacts.opnfv.org/functest/NAEGSVQB2IN6/functest-opnfv-functest-smoke-hunter-tempest_slow-run-189/tempest_slow/tempest-report.html

collivier:
Linked to req.inf.ntw.16 [1]
The Architecture must support dual stack IPv4 and IPv6 for tenant
networks and workloads

Closes #1556

[1] https://github.com/cntt-n/CNTT/blob/master/doc/ref_arch/openstack/chapters/chapter02.md

### DIFF
--- a/doc/ref_cert/lfn/chapters/chapter03.md
+++ b/doc/ref_cert/lfn/chapters/chapter03.md
@@ -53,7 +53,7 @@ For detailed information on RM & RA-1 NFVI and VNF requirements, please refer to
 <a name="3.3.1"></a>
 ### 3.3.1 Architecture and OpenStack Requirements
 
-<!-- 
+<!--
 |#|Requirement|Description|Catalog|Status|MUST|Category|RI-1 Implemented (Yes/No)|RC-1 Coverage (project)|RC-1 Project To On-board|
 |--- |--- |--- |--- |--- |--- |--- |--- |--- |--- |
 |1|req.gen.ost.01|must use OpenStack APIs.|General|in Functest|Must|Functional||Functest||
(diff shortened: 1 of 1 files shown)


pgoyal01:
@collivier Does the test_network_v6 test for dual IP stack or just test support for IPv6? The requirement req.inf.ntw.16 is for teh support of "dual stack IPv4 and IPv6 for tenant networks and workloads.

collivier:
Yes these tests verify both
  - ipv4: ssh connection from Functest to VMs (though the floating ip)
  - ipv6: ping from VMs to router

It doesn't cover an IPv6-only deployment (e.g. testing container using IPv6, OS endpoint in IPv6, etc.). They exactly match the requirement. But regarding the way how tempest tests are designed, we won't cover the 2in1 requirements as here in most of the cases.

They are other tests already-in-use at the API levels (e.g. subnet operations).

collivier:
It's worth mentioning that 8 tests have been developed to cover the different addressing schemes
https://docs.openstack.org/neutron/pike/admin/config-ipv6.html

PR #1564, opened 2020-05-08
RM Ch3 Netw Model png diagram

walterkozlowski:
RM-Chapter4-Network-Model
Figure 3-5-1


ulikleber:
I don't have an issue with the diagram, but what we do here is nothing for code freeze phase. What is the explaining text around it? Does the text explain what are the management clients?

walterkozlowski:
Replaced by PR #1568

PR #1567, opened 2020-05-08
[Tech Glossary] Clarifying terminology of CPU and threads

### LINKED ISSUE #1566: [TSC Glossary] Pcpu and multithreading terms used incorrectly

visnyei:
Closes #1566.

### DIFF
--- a/doc/tech/glossary.md
+++ b/doc/tech/glossary.md
@@ -8,7 +8,7 @@
   * [1.2 Hardware layers terminology](#1.2)
   * [1.3 Operational and administrative terminology](#1.3)
   * [1.4 Container Related Terminology](#1.4)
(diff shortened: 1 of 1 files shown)


petorre:
Changing

> but is integrated with other cores on a multiprocessor chip (integrated circuit die). Please note that the multiprocessor chip is also referred to as a CPU that is placed in a socket of a computer motherboard.


to

> and is integrated with other cores on a multicore processor (chip, integrated circuit die). Please note that the multicore processor is also referred to as a CPU that is placed in a socket of a computer motherboard.

would make it more precise for multicore processors vs multiprocessor packages.

visnyei:
Thank you! Comments have been implemented. :)

PR #1579, opened 2020-05-14
[RC1] Blacklist tempest tests newer than Nova microversion 2.53

collivier:
RA1 chapter5 select Pike's microversions and then a few tests should have been skipped in CNTT.
It's linked to https://gerrit.opnfv.org/gerrit/c/functest/+/70186

### DIFF
--- a/doc/ref_cert/lfn/chapters/chapter03.md
+++ b/doc/ref_cert/lfn/chapters/chapter03.md
@@ -53,7 +53,7 @@ For detailed information on RM & RA-1 NFVI and VNF requirements, please refer to
 <a name="3.3.1"></a>
 ### 3.3.1 Architecture and OpenStack Requirements
 
-<!-- 
+<!--
 |#|Requirement|Description|Catalog|Status|MUST|Category|RI-1 Implemented (Yes/No)|RC-1 Coverage (project)|RC-1 Project To On-board|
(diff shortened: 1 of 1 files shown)


pgoyal01:
@collivier I will open a bug as in RA-1 Chapter 5 we refer to "Minimal microversion" but should be the "maximum microversion"

collivier:
@pgoyal01 I see your point in the other PR but I would consider the first proposal as better (it depends on how you interpret the tab).
Here the main issue is that I run this 5 tests (out of 2000+) by error because they cover features newer than Pike. It's not so simple because the SUT is based on Rocky and this test were successfully passing (Thank you Orange for the feedbacks). It's enforced now because I set all tempest max_version to Pike in all CNTT containers (including Functest master).

Thank you for the approvals!

collivier:
In tempest, max_microversion is about the tests executed. All tests asking for a newer version are skipped even if the SUT supports the capabilities
https://git.opnfv.org/functest/tree/docker/smoke-cntt/tempest_conf.yaml?h=stable/hunter#n3

The default tempest conf in Hunter selects the Rocky microversions because Hunter is designed to test Rocky and older.
https://git.opnfv.org/functest/tree/functest/opnfv_tests/openstack/tempest/custom_tests/tempest_conf.yaml#n3

PR #1587, opened 2020-05-14
[RA-1 Ch05] Change to "Maximal API microversion"

### LINKED ISSUE #1586: [RA-1 Ch05] Change "Minimal microversion" to "Maximal"

pgoyal01:
The microversions specified are the maximal versions for Pike -- correcting typo ("minimal")

Fixes #1586

### DIFF
--- a/doc/ref_arch/openstack/chapters/chapter05.md
+++ b/doc/ref_arch/openstack/chapters/chapter05.md
@@ -32,7 +32,7 @@ For the purpose of conformance tests, this chapter also identifies the set of th
 
 ### 5.2.1. Keystone
 
-| **OpenStack Service** | **API Version** | **Minimal API Microversion** |
+| **OpenStack Service** | **API Version** | **Maximal API Microversion** |
 |-----------------------|-----------------|------------------------------|
 | Identity: Keystone    | v3              | 3.8                          |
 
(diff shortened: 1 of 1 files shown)


pgoyal01:
@collivier As you know the Openstack specifies the maximum microversion that can be utilised with a particular OpenStack release and this change is trying to capture that independent of how we test.

collivier:
I think we all agree on the same statement and the fact that OpenStack Rocky can be compliant with RA1 Chapter5 even by deploying newer extensions. max is more for the testing part (RC1), because we can't test something newer.

I think the tab is better asis (but it could be more explained before) else Queens and newer can't be compliant with CNTT.

collivier:
BYW, nova and cinder return min_version and version in API. 
      "version": "2.72",
      "min_version": "2.1",

      "version": "3.59",
      "min_version": "3.0",

collivier:
If you want to switch to max microversion, would it be possible to add an introduction making clear that CNTT selects features/capabilities/operations according the next max_microversions (Pike) but the SUT can offer newer versions thanks to the OpenStack microversion policies.

collivier:
It's worth mentioning that we are running RC vs OpenStack Queens for the fiel trial.

pgoyal01:
@collivier Thanks.

Added the following to Section 5.2:
Please note that OpenStack provides a maximum microversion to be used with an OpenStack release. In the following sections the "Maximal API Version" refers to this maximum microversion specified for the OpenStack Pike release. Please note that in Reference Conformance (RC-1) testing, the System Under Test (SUT) can utilise newer microversions because of the OpenStack microversion policies. As per multiple OpenStack services documentation, for example the [Compute Service](https://docs.openstack.org/api-guide/compute/microversions.html), "A cloud that is upgraded to support newer microversions will still support all older microversions to maintain the backward compatibility for those users who depend on older microversions."

rabi-abdel:
@collivier please re-review

rabi-abdel:
@collivier please re-review

PR #1620, opened 2020-05-21
[RA-1 Ch02] Clean 2.3 -- "should" and "may" reqmts

### LINKED ISSUE #1618: [RA-1 Ch02] Move "should" and "may" requirements from 2.3 to 2.4
Section 2.3 should only contain "must" requirements but it still contains 10 "should" and 2 "may" requirements

Change has to be made in both Baldy and Baraque releases

pgoyal01:
Move "should" and "may" requirements from 2.3 to 2.4

Fixes #1618


karinesevilla:
@goyal For Req.inf.stg.11 and Req.inf.ntw.18, I agree to move them into the recommendations section considering that they can cover non mandatory capabilities. But in this case, the link to chapter 5 in the Traceability column is no longer valid, it must be removed.

rabi-abdel:
@pgoyal01 can you please address @collivier and @karinesevilla above?

pgoyal01:
@collivier @karinesevilla 

| Ref # | sub-category | Description |  Comment |
|----|----------------|----------------------|-----------|
| `req.inf.stg.06` | Storage | The Architecture **should** make the immutable images available via location independent means. | This is about images and not about storage. |
| `req.inf.stg.11` | Storage | The Architecture **should** support the Block storage capabilities specified in https://docs.openstack.org/api-ref/block-storage/. | The problem with this requirement is that it refers to quite a large # of block storage features and not all of these are mandatory. The reference can be changed to that in Chapter 4 but we need some granular mandatory requirements for block storage to justify the selections in Chapter 05. |
| `req.inf.ntw.18` | Network | The Architecture **should** support the network extensions specified in https://docs.openstack.org/api-ref/network/v2/.| Comments similar to those for `req.inf.stg.11` above |

Primary disconnect is that in Chpater05 we have mandatory features but no requirements driving the need for those features.

rabi-abdel:
@collivier please re-review

karinesevilla:
@pgoyal01 Regarding your comments,
req.inf.stg.06: I think it must be rephrased to be clarified
req.inf.stg.11 and req.inf.ntw.18: I agree,  these 2 requirements are too large and for me not useful.

rabi-abdel:
this PR needs resolving. @pgoyal01 will be good to get this discussed during next RA-1 call.

rabi-abdel:
@pgoyal01 would you mind resolving the conflicts above (i.e. rebasing) so this can be merged?

PR #1627, opened 2020-05-27
[Edge] Source File (PPT) for Edge Use Cases

Karimrabie:
These are the slides presented at the first Edge group call.
It includes the source files for the use cases figures.


pgoyal01:
@Karimrabie 
1. Each source figure needs to be its own file and there is a naming scheme.
2. Please remove WIP label when ready for review.

Karimrabie:
> @Karimrabie
> 
> 1. Each source figure needs to be its own file and there is a naming scheme.
> 2. Please remove WIP label when ready for review.

1- Each source figure should be a separate ppt file?
2- Done

PR #1628, opened 2020-05-28
[RA1]RC1] Skip manage_snapshot in CNTT

collivier:
As detected by Orange's RC Field trial, Ceph doesn't support this
feature before Rocky [1].
"AttributeError: 'RBDDriver' object has no attribute 'unmanage_snapshot'"

manage_snapshot testing is fully passing in Functest SUT because CNTT
is verified vs OpenStack Rocky and newer.

Ceph is selected by RI as backend and is widely reused by the ecosystem.
It's similar to [2].

manage_snapshot will be mandatory back once CNTT selects Rocky or newer.

[1] https://bugs.launchpad.net/cinder/+bug/1645288
[2] https://gerrit.opnfv.org/gerrit/c/functest/+/69888

### DIFF
--- a/doc/ref_arch/openstack/chapters/chapter05.md
+++ b/doc/ref_arch/openstack/chapters/chapter05.md
@@ -72,7 +72,7 @@ Image Service Versions: https://docs.openstack.org/api-ref/image/versions/index.
 | clone                  | X             |
 | consistency_group      |               |
 | extend_attached_volume |               |
-| manage_snapshot        | X             |
+| manage_snapshot        |               |
 | manage_volume          | X             |
 | multi_backend          |               |
 | snapshot               | X             |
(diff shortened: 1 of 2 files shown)


pgoyal01:
@collivier Since we are going to upgrade to Rocky or newer, I suggest we leave this in place and only commit PR #1629 (Baldy). Avoids unnecessary work. Or we open an issue now to reinstate so that we do not forget. My preference would be to not make the change.

collivier:
@rabi-abdel @rabiabdel Would you mind reviewing and merging this PR? The related change was applied in Functest (Field trial is coming)

pgoyal01:
@collivier @rabi-abdel PR #1629 is for Baldy release and makes sense to make the change. PR #1628 is for the current viz., Baraque release and I think it would be a waste of effort to make the change now and undo later..

collivier:
@pgoyal01 @rabi-abdel The revert is inappropriate here (and even more when the author is not aware) and frankly disconnected from the classical process. Until we switched to Train and the doc is simply false. The revert should go through the same process and not applied by admin without concensus.
I agree regarding the overall target but what we try to avoid is less disruptive somehow than the mess done here (+now RC is unsynced with RA1).

PR #1668, opened 2020-06-10
Adding readthedocs style build scripts

CsatariGergely:
This change adds the possibility to build the documentation locally
using the same toolset what is used by readthedocs.io.
The build can be started with the `command tox -e docs` and
requires the existence of python and tox.


collivier:
@CsatariGergely I would prefer us to apply the PBR model even if here there is no py package. In that case requirements.txt should manage abstract requirements (e.g. <=) and upper-constraints.txt concrete requirements (e.g. ===). Not a big deal right now.

collivier:
@rabi-abdel @rabiabdel @CsatariGergely we may leverage on this first commit to verify links and spelling across the project. As far as I know sphinx can handle markdown files as inputs.

rabi-abdel:
@collivier is the proposal to expand the scripts to add more checks? can you propose a PR for this?

collivier:
@rabi-abdel Yes it could be another PR. It's about defining a few tox entries which would be called by verification bots.

Issue #1569, opened 2020-05-10
[Edge-RM CH02] New Profile for Edge ( vRAN ,..)


ASawwaf:
Dears, this issue for more decision about how can we cover the vRAN or/and any workload required the same requirements ( FPGA ) at the edge, we agreed during Edge meeting that we need to have a new profile so 

we have 2 options : 
1- Re-activate the Compute intensive 
2- Create a new profile

kedmison:
Compute-optimized was intended as a 'better' version of basic; either faster vCores, or more vCores.  FPGAs are an entirely different computational paradigm, using entirely different binaries (FPGA bitstreams), rather than x86 instruction sets.

For this reason, I would recommend introducing this as a FPGA profile.  
This profile should also explore the required vCPU count, RAM, etc and also address how many of these FPGAs may be installed in the system, and how they are associated to workloads.

kedmison:
RM meeting 2020-05-13: We are not opposed to a workload profile for FPGA, but need vendor-independent characterization of the  FPGA adapter card capabilities prior to defining the profile.

ASawwaf:
thanks all for our discussion today, we will continue to cover this in edge meeting 2020-05-14
but it makes my life more complicated, i need big support from @petorre @trevgc if we can find some characterization, which I believe the FPGAs as Technologies by nature has independent characterization due to the Fact of the use case that it used for

trevgc:
I was told that even today it is possible to design a very low capacity FlexRAN solution without FPGA. To be verified what this really means but I think highlights that we need to be careful about assuming a profile is dependent on any particular technology.

TFredberg:
I agree with @trevgc that CNTT should not assume that it is impossible to run a specific workload without some special HW that might make the workload run more efficient or live up to hard characteristics requirements.

There is however a need in all cases to specify the characteristics needs from the workloads and the characteristics capabilities from the infrastructure resource instances in any deployed Cloud Infrastructure. These characteristics needs to be expressed in a way that make sense both for the workload and the resource provider. 

It think it would be very good if the involved subject matter expert around the 4 use cases could have a first suggestion about what characteristics matter and see if they could express them in a technology independent way, but still be useful for the resource provisioning when inserting an FPGA, GPU or SmartNIC PCIe card.

lucisuciu:
Dear All, 

IMHO, I'll go for the creation of a new edge profile, and it should be techology-independent if possible... 

[...]

PR #1692, opened 2020-06-14
[Edge] Define a new profile "Edge Profile"

ASawwaf:
Fixes https://github.com/cntt-n/CNTT/issues/1569

### DIFF
--- a/doc/ref_model/chapters/chapter02.md
+++ b/doc/ref_model/chapters/chapter02.md
@@ -148,15 +148,21 @@ By trying to sort workloads into different categories based on the requirements 
     - Predictable computing
     - High network throughput
     - Low network latency
-<!--
+    
 - **Profile Three**
   - Workload types
-    - Control plane functions with specific computing needs
-    - _Examples: MME, AMF, IMS-CSCF_
+    - Edge functions with very specific computing and networking needs
+    - _Examples: vRAN ( vDU , vCU ) **First Edge use cases** https://github.com/cntt-n/CNTT/blob/master/doc/tech/usecases.md
   - Requirements
-    - Predictable computing
-    - Low network latency    
+    - very Low network latency
+    - Offload some Radio requirments
+       - Hi-PHY processing, FPGA accelerators are used for enhancing performance. 
+       - FEC offload at L1.
+       - 5G NR, PHY is offloaded to the FPGA
+       - Real Time Signal Processing 
+       - Fronthaul Acceleration & Conversion
 
+<!--
 - **Profile Four**
   - Workload types
     - Control plane functions with specific storage needs
@@ -172,6 +178,7 @@ By trying to sort workloads into different categories based on the requirements 
 Based on the above analysis, following cloud infrastructure profiles are proposed (also shown in **Figure 2-1** below)
 - **Basic**: for Workloads that can tolerate resource over-subscription and variable latency.
 - **Network Intensive**: for Workloads that require predictable computing performance, high network throughput and low network latency.
+- **Edge** : For Edge use cases that requirment sepical requirements, in terms of offloading some functions and with very low network latency.
 
 <p align="center"><img src="../figures/ch02_infra_profiles.PNG" alt="infra_profiles" title="Infrastructure Profiles" width="100%"/></p>
 <p align="center"><b>Figure 2-1:</b> Infrastructure profiles proposed based on VNFs categorisation.</p>
--- a/doc/ref_model/chapters/chapter02.md
+++ b/doc/ref_model/chapters/chapter02.md
@@ -180,7 +180,7 @@ By trying to sort workloads into different categories based on the requirements 
 Based on the above analysis, following cloud infrastructure profiles are proposed (also shown in **Figure 2-1** below)
 - **Basic**: For Workloads that can tolerate resource over-subscription and variable latency.
 - **Network Intensive**: For Workloads that require predictable computing performance, high network throughput and low network latency.
-- **Edge** : For Edge use cases that have special requirements, in terms of offloading some functions and and very low network latency.
+- **Edge** : For Edge use cases that have special requirements, in terms of offloading some functions , very low network latency ,very high storage requirements and very high computing requirements "GPU" .
 
 <p align="center"><img src="../figures/ch02_infra_profiles.PNG" alt="infra_profiles" title="Infrastructure Profiles" width="100%"/></p>
 <p align="center"><b>Figure 2-1:</b> Infrastructure profiles proposed based on VNFs categorisation.</p>
--- a/doc/ref_model/chapters/chapter02.md
+++ b/doc/ref_model/chapters/chapter02.md
@@ -178,7 +178,7 @@ By trying to sort workloads into different categories based on the requirements 
 Based on the above analysis, following cloud infrastructure profiles are proposed (also shown in **Figure 2-1** below)
 - **Basic**: For Workloads that can tolerate resource over-subscription and variable latency.
 - **Network Intensive**: For Workloads that require predictable computing performance, high network throughput and low network latency.
-- **Edge** : For Edge use cases that requirment sepical requirements, in terms of offloading some functions and with very low network latency.
+- **Edge** : For Edge use cases that requirment special requirements, in terms of offloading some functions and with very low network latency.
 
 <p align="center"><img src="../figures/ch02_infra_profiles.PNG" alt="infra_profiles" title="Infrastructure Profiles" width="100%"/></p>
 <p align="center"><b>Figure 2-1:</b> Infrastructure profiles proposed based on VNFs categorisation.</p>
--- a/doc/ref_model/chapters/chapter02.md
+++ b/doc/ref_model/chapters/chapter02.md
@@ -180,7 +180,7 @@ By trying to sort workloads into different categories based on the requirements 
 Based on the above analysis, following cloud infrastructure profiles are proposed (also shown in **Figure 2-1** below)
 - **Basic**: For Workloads that can tolerate resource over-subscription and variable latency.
 - **Network Intensive**: For Workloads that require predictable computing performance, high network throughput and low network latency.
-- **Edge** : For Edge use cases that have special requirements, in terms of offloading some functions , very low network latency ,very high storage requirements and very high computing requirements "GPU" .
+- **Edge** : For Edge use cases that have special requirements, in terms of offloading some functions, very low network latency, very high storage requirements and very high computing requirements "GPU".
 
 <p align="center"><img src="../figures/ch02_infra_profiles.PNG" alt="infra_profiles" title="Infrastructure Profiles" width="100%"/></p>
 <p align="center"><b>Figure 2-1:</b> Infrastructure profiles proposed based on VNFs categorisation.</p>
--- a/doc/ref_model/chapters/chapter02.md
+++ b/doc/ref_model/chapters/chapter02.md
@@ -176,8 +176,8 @@ By trying to sort workloads into different categories based on the requirements 
 # 2.3 Cloud Infrastructure Profiles
 
 Based on the above analysis, following cloud infrastructure profiles are proposed (also shown in **Figure 2-1** below)
-- **Basic**: for Workloads that can tolerate resource over-subscription and variable latency.
-- **Network Intensive**: for Workloads that require predictable computing performance, high network throughput and low network latency.
+- **Basic**: For Workloads that can tolerate resource over-subscription and variable latency.
+- **Network Intensive**: For Workloads that require predictable computing performance, high network throughput and low network latency.
 - **Edge** : For Edge use cases that requirment sepical requirements, in terms of offloading some functions and with very low network latency.
 
 <p align="center"><img src="../figures/ch02_infra_profiles.PNG" alt="infra_profiles" title="Infrastructure Profiles" width="100%"/></p>
--- a/doc/ref_model/chapters/chapter02.md
+++ b/doc/ref_model/chapters/chapter02.md
@@ -180,7 +180,7 @@ By trying to sort workloads into different categories based on the requirements 
 Based on the above analysis, following cloud infrastructure profiles are proposed (also shown in **Figure 2-1** below)
 - **Basic**: For Workloads that can tolerate resource over-subscription and variable latency.
 - **Network Intensive**: For Workloads that require predictable computing performance, high network throughput and low network latency.
-- **Edge** : For Edge use cases that requirment special requirements, in terms of offloading some functions and with very low network latency.
+- **Edge** : For Edge use cases that requirment special requirements, in terms of offloading some functions and and very low network latency.
 
 <p align="center"><img src="../figures/ch02_infra_profiles.PNG" alt="infra_profiles" title="Infrastructure Profiles" width="100%"/></p>
 <p align="center"><b>Figure 2-1:</b> Infrastructure profiles proposed based on VNFs categorisation.</p>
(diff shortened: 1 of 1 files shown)


doc/ref_model/chapters/chapter02.md
```diff
@@ -148,15 +148,21 @@ By trying to sort workloads into different categories based on the requirements
...
     - High network throughput
     - Low network latency
-<!--
+    
 - **Profile Three**
   - Workload types
-    - Control plane functions with specific computing needs
-    - _Examples: MME, AMF, IMS-CSCF_
+    - Edge functions with very specific computing and networking needs
+    - _Examples: vRAN ( vDU , vCU ) **First Edge use cases** https://github.com/cntt-n/CNTT/blob/master/doc/tech/usecases.md
```
CsatariGergely, line 155:
```suggestion
    - _Examples: vRAN ( vDU , vCU ) [First Edge use cases](../../tech/usecases.md)
```

doc/ref_model/chapters/chapter02.md
```diff
@@ -170,8 +176,9 @@ By trying to sort workloads into different categories based on the requirements
 # 2.3 Cloud Infrastructure Profiles
 
 Based on the above analysis, following cloud infrastructure profiles are proposed (also shown in **Figure 2-1** below)
-- **Basic**: for Workloads that can tolerate resource over-subscription and variable latency.
-- **Network Intensive**: for Workloads that require predictable computing performance, high network throughput and low network latency.
+- **Basic**: For Workloads that can tolerate resource over-subscription and variable latency.
+- **Network Intensive**: For Workloads that require predictable computing performance, high network throughput and low network latency.
+- **Edge** : For Edge use cases that requirment sepical requirements, in terms of offloading some functions and with very low network latency.
```
CsatariGergely, line 181: Will this be exclusive for edge use cases?

ASawwaf: @CsatariGergely Yes

msidana, line 181: For Edge use cases that have special~~requirment sepical~~ requirements, in terms of offloading some functions and with very low network latency.

doc/ref_model/chapters/chapter02.md
```diff
@@ -170,8 +176,9 @@ By trying to sort workloads into different categories based on the requirements
 # 2.3 Cloud Infrastructure Profiles
 
 Based on the above analysis, following cloud infrastructure profiles are proposed (also shown in **Figure 2-1** below)
-- **Basic**: for Workloads that can tolerate resource over-subscription and variable latency.
-- **Network Intensive**: for Workloads that require predictable computing performance, high network throughput and low network latency.
+- **Basic**: For Workloads that can tolerate resource over-subscription and variable latency.
```
msidana, line 179: Any category you see for compute intensive or special purpose compute like GPUs etc. ?

doc/ref_model/chapters/chapter02.md
```diff
@@ -170,8 +176,9 @@ By trying to sort workloads into different categories based on the requirements
...
+- **Network Intensive**: For Workloads that require predictable computing performance, high network throughput and low network latency.
```
msidana, line 180: Do we need to include **predictable computing** performance for network intensive ? This category should stick to networking requirements only. Thoughts ?

doc/ref_model/chapters/chapter02.md
ASawwaf: @msidana , the initial profiles include compute-intensive but we removed it as there is no use cases for it that time

doc/ref_model/chapters/chapter02.md
ASawwaf: @msidana , the initial profiles include compute-intensive but we removed it as there is no use cases for it that time

doc/ref_model/chapters/chapter02.md
```diff
@@ -170,8 +176,9 @@ By trying to sort workloads into different categories based on the requirements
...
+- **Edge** : For Edge use cases that requirment special requirements, in terms of offloading some functions and with very low network latency.
```
pgoyal01, line 181: suggest
```suggestion
- **Edge** : For Edge use cases that have special requirements, in terms of offloading some functions and with very low network latency.
```

pgoyal01:
General, all acronyms need the expansion when first used

ASawwaf:
@pgoyal01 Can you pin out the new acronyms 
-PHY ?
- FPGA ?

petorre:
In Requirements, do you like to mention something on
- Timing Accuracy
- Edge-limited power envelope

?

petorre:
"Offload some Radio requirments" -> "Accelerate Radio functionality/algorithms like", followed by lines you have just from those I would take the How part (currently FPGA) into separate line and write it more generically as "Acceleration implemented as Programmable (like FPGA) or Fixed Function (for stable algorithms)" or something like that, which we discussed on one of the calls.

doc/ref_model/chapters/chapter02.md
pgoyal01: "...  and with very low network latency." --> "...  and very low network latency."

ASawwaf:
> In Requirements, do you like to mention something on
> 
> * Timing Accuracy
> * Edge-limited power envelope
> 
> ?

@petorre , Yes please

doc/ref_model/chapters/chapter02.md
ASawwaf: done @pgoyal01 thx

ASawwaf:
@pgoyal01 regarding acronyms , what did you think for best place to define them 
- Within the new profile  (RM 02 )
- CNTT- Use cases - Edge 
- Glossary

pgoyal01:
Couple of observations:

1. At the beginning of the write-up on profiles (Section 2.2), the sentence "By trying to sort workloads into different categories based on the requirements observed, below are the different profiles concluded, which are mainly driven by expected performance levels:" needs to corrected. One suggestion, but there maybe other better ones, is: By categorizing workloads by their requirements, the following profiles, primarily driven by expected performance, are defined:"

2. In Section 2.3, there is an unlabeled diagram that now needs to include the Edge profile. The diagram needs to be scaled down too.

3. Tom from GSMA, if I understood him correctly, wants us to include all referred to content within the RM chapters. In this case, it would be "Edge use cases". Which chapter would the use cases be best suited for inclusion in?

ASawwaf:
@pgoyal01 , Thanks for this comment it really makes sense,  let me address it 


1-At the beginning of the write-up on profiles (Section 2.2), the sentence "By trying to sort workloads into different categories based on the requirements observed, below are the different profiles concluded, which are mainly driven by expected performance levels:" needs to corrected. One suggestion, but there maybe other better ones, is: By categorizing workloads by their requirements, the following profiles, primarily driven by expected performance, are defined:"

It is valid, and as the same time it is very critical as it is one of the cores of RM./CNTT
i would like to discuss to have a consistency to the reader 


In Section 2.3, there is an unlabeled diagram that now needs to include the Edge profile. The diagram needs to be scaled down too.

2-it is my bad as I tried to upload the image but failed 


3-Tom from GSMA, if I understood him correctly, wants us to include all referred to content within the RM chapters. In this case, it would be "Edge use cases". Which chapter would the use cases be best suited for inclusion in? I think RM Chapter 8 is now empty.


totally agreed, we have a discussion before in edge meeting that does not make to Edge in a high level in ToC and it should under anywhere in RM , but there is consensus to leave it at AS IS 
but I will trigger it again , and I like to utilize CH08 @rabi-abdel @bfcohen  what do you think again ?

doc/ref_model/chapters/chapter02.md
```diff
@@ -170,8 +178,9 @@ By trying to sort workloads into different categories based on the requirements
 # 2.3 Cloud Infrastructure Profiles
 
 Based on the above analysis, following cloud infrastructure profiles are proposed (also shown in **Figure 2-1** below)
-- **Basic**: for Workloads that can tolerate resource over-subscription and variable latency.
-- **Network Intensive**: for Workloads that require predictable computing performance, high network throughput and low network latency.
+- **Basic**: For Workloads that can tolerate resource over-subscription and variable latency.
+- **Network Intensive**: For Workloads that require predictable computing performance, high network throughput and low network latency.
+- **Edge** : For Edge use cases that have special requirements, in terms of offloading some functions , very low network latency ,very high storage requirements and very high computing requirements "GPU" .
```
CsatariGergely, line 183: Do we really need GPU-s on the edge? For me GPU-s are more for ML training and I do not see an use case for this on the edge.

CsatariGergely, line 183:
```suggestion
- **Edge** : For Edge use cases that have special requirements, in terms of offloading some functions, very low network latency, very high storage requirements and very high computing requirements "GPU".
```

Karimrabie: @CsatariGergely Hello Gergely, we have the use case of the remote Edge Survellience with video analytics. Where the live streams need to be analyzed at the Edge (ML-assisted) and alarms are raised for the suspected scenes. It is one of the 5G use cases for Mission Critical IoT.

doc/ref_model/chapters/chapter02.md
```diff
@@ -180,7 +180,7 @@ By trying to sort workloads into different categories based on the requirements
 Based on the above analysis, following cloud infrastructure profiles are proposed (also shown in **Figure 2-1** below)
 - **Basic**: For Workloads that can tolerate resource over-subscription and variable latency.
 - **Network Intensive**: For Workloads that require predictable computing performance, high network throughput and low network latency.
-- **Edge** : For Edge use cases that have special requirements, in terms of offloading some functions , very low network latency ,very high storage requirements and very high computing requirements "GPU" .
+- **Edge** : For Edge use cases that have special requirements, in terms of offloading some functions, very low network latency, very high storage requirements and very high computing requirements "GPU".
```
pgoyal01, line 183: Does Edge have "very high storage requirements"? Many Edge use cases do not have even medium storage requirements. 
I think "GPU" needs to be deleted.

```suggestion
- **Edge** : For Edge use cases that have special requirements, in terms of offloading some functions, very low network latency and very high computing performance requirements.
```

ASawwaf: @pgoyal01 for vRAN use cases there is no need for GPU , but for the edge to make it more wider i belive GPU required for other use cases 

i will copy karim comments 

karim quote
"
Hello Gergely, we have the use case of the remote Edge Survellience with video analytics. Where the live streams need to be analyzed at the Edge (ML-assisted) and alarms are raised for the suspected scenes. It is one of the 5G use cases for Mission Critical IoT."

what do u think

pgoyal01: @ASawwaf we need to reword the sentence. There is also the issue about the " very high storage requirements" for the Edge. 

```suggestion
- **Edge** : For use cases that have special requirements, in terms of offloading some functions, very low network latency, and very high computing performance requirements. For example, the use of GPU for remote surveillance with video analytics where the live streams are, ML-assisted, analyzed at the Edge and alarms are raised for the suspected scenes. 
```

kedmison: GPUs should be in a different workload profile. One could request that hostgroups supporting the GPU workload profile be deployed in edge datacenters.

kedmison: Similar for storage requirements.  If high storage requirements are needed, they should be identified as a separate workload profile, with a supporting host aggregate, and then that can be deployed at the edge if needed.

TFredberg:
I fully support @kedmison comment of a more neutral name of the profiles. I would also recomend that we start thinking about naming/denomination rules for profiles possibly in the RM.
I would think that in general a profile name should be agnostic from the function it could implement, use case it could be used for and the deployment environment in where it could be deployed.
This would make the Network a boarderline name, but likely OK if we view that as an area or class of offered HW functionality and characteristics we like to express.

ASawwaf:
@pgoyal01 @kedmison @kedmison 

let address it , it is my bad to add " very high storage requirements and very high computing requirements "GP" **but my intention to make it wider not focus on vRAN only (FPGA)**

as Edge can fit many use cases and each use case has different requirements ( discussed in last Edge meeting  )
vRAN==> FPGA
CDN==>  persistent memory and huge storage 
Analytics, AI ==>Vision Processing Unit (VPU).
Video & Gaming ==>GPU

And for FPGA profile, FPGA can be used at the edge and non-edge @kedmison 


So we have either focus on vRAN or make it more generic to cover Edge use cases

kedmison:
@ASawwaf The edge location can contain multiple different host aggregates that each satisfy one or more of the workloads you are identifying
vRAN -> NF (network-FPGA) hostgroup
CDN -> new storage profile and hostgroup. 
Video, Gaming, AI -> GPU

An edge cloud could contain multiple NF hosts, multiple storage hosts, multiple, GPU hosts, but not necessarily require that they all be within the same server or host aggregate.

ASawwaf:
@kedmison , Yup , so you answer me question, " we have either focus on vRAN or make it more generic to cover Edge use cases"

so I will prefer to revert back to be on the use case driven , so let cover whatever in our hand which vRAN which mandated FPGA

for the profile name, although we have consensus over the "edge profile" name, I open for any suggestion

TFredberg:
The use of "host aggregates" at least in OpenStack denominate a way to group hosts which I think consist of both HW and (host-OS) SW, but not its use case: 
"Host aggregates are a mechanism for partitioning hosts in an OpenStack cloud, or a region of an OpenStack cloud, based on arbitrary characteristics. Examples where an administrator may want to do this include where a group of hosts have additional hardware or performance characteristics."

This is along the lines I talked about above that profiles and "host aggregates" should be naming and grouping of HW and hosts with an offered characteristics and not a use case, loaded function nor a deployment scenario.

@kedmison I am not shure we should call a grouping of HW with the same HW profile a hostgroup since a host is HW + OS.

FPGA would then work, but not RAN nor Edge, but I think we first must agree on the principles on which the HW Profiles are to be named. Do we?

kedmison:
@tfredberg you are correct, a host aggregate can be created for many reasons.  
I think we're still talking about workload profiles, which would then be mapped to a hardware profile (new or existing) that is required to support it.

I'm reluctant to use FPGA directly as that would also encompass the potential case (observed in public cloud) where a FPGA card is used and does not have any NIC attached to that card.  
The key feature hardware enabler at issue seems to be a FPGA-enabled NIC card, whose most clearly identified use case is a vRAN application.  For that reason, I would want any name here to associate those characteristics (network and FPGA) together.

TFredberg:
@kedmison We seems to be very aligned on most areas:
- workload profiles are mapped onto HW profiles
- HW profiles should only state its HW capabilities and not deployment or application suitability use cases
- HW profiles should be specific enough to enable mapping of applications with specific type (but not instance of) HW requests

I would think (but that is my guess) that the specific characteristics of an FPGA based HW that is requested for xRAN would be an FPGA in-line between the CPU and Ethernet ports. If then the FPGA has a soft NIC inside or a hard NIC outside should be of less importance. In that sence it would still be a NetWorking-Fpga instance. If we claim that it has to be a NIC on the PCI card we also have to get into what brand and type of NIC it is and that becomes messy.

karinesevilla:
@ASawwaf  I think that the name "Edge" is not refecting a unique set of requirements. 
Edge use cases don't have all the same constraints.
The major and common characteristics of the profile we  want to define are:
Network intensive + Compute intensive + time constraints (low latency + time synchro).
To capture these requirements I would prefer to name this profile "High Performances" profile rather than a name mentioning a techno like FPGA which could be replaced by another techno.

rabi-abdel:
@ASawwaf please advice what would you like to do with this PR. based on the conclusion of last meeting to add extensions to NI profile, shall this PR be closed now?

ASawwaf:
Thanks for your contribution as agreed last week, we will look on Profile capabilities in ch04 and figure out which one will be used in Edge use cases so no need for edge profiles

PR #1710, opened 2020-06-18
[ROOT] Add navigation tips on top document

collivier:
It would ease the different type of users to go the best target.
The list will be updated according to the next updates (e.g. RC2).


collivier:
All credits to @scottsteinbrueck for the original idea

PR #1773, opened 2020-07-10
[RM Ch04] Hardware and Virtual Infrastructure Layer concepts

TFredberg:
Added an updated version of Hardware and Virtual Infrastructure Layer concepts from the Baldy approved Tech/approach.md

### DIFF
--- a/doc/ref_model/chapters/chapter04.md
+++ b/doc/ref_model/chapters/chapter04.md
@@ -623,6 +623,20 @@ Before we start developing a deep model we need to agree on some foundational co
  - Software Defined Underlay and Overlay Networking concepts 
  - Programmable Networking Fabric concept
 
+<a name="4.3.1.1"></a>
+#### 4.3.1.1 Underlay and Overlay Networking concepts
+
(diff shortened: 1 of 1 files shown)


doc/ref_model/chapters/chapter04.md
```diff
@@ -636,6 +636,19 @@ In some instances, the Virtualisation Tenants can bypass the Overlay Networking
...
+
+These concepts are very similar to how the Hyperscaler Cloud Providers (HCP) offer Virtual Private Clouds for users of Bare Metal deployment on the HCP shared pool of servers, storage and networking resources.
+
+The separation of Hardware and Virtual Infrastructure Layers administrative domains makes it important that the Reference Architectures do not include direct management or dependencies of the pooled physical hardware resources in the HW Infrastructure Layer e.g. servers, switches and underlay networks from within the Virtual Infrastructure Layer. All automated interaction from the Virtual Infrastructure Layer implementations towards the HW Infrastructure with its shared networking resources in the HW Infrastructure Layer must go through a common abstracted Reference Model interface.
```
pgoyal01, line 650: for an implementation it has to something concrete and not just as abstracted Reference Model interface.
```suggestion
The separation of Hardware and Virtual Infrastructure Layers administrative domains makes it important that the Reference Architectures do not include direct management or dependencies on the pooled physical hardware resources in the HW Infrastructure Layer e.g. servers, switches and underlay networks from within the Virtual Infrastructure Layer. All interactions from the Virtual Infrastructure Layer implementations with the HW Infrastructure, with its shared resources, must be through an abstracted interface layer. 
```

PR #1776, opened 2020-07-10
[RM Ch04] Networking Reference Model

TFredberg:
Added the updated Networking Reference Model from the approved contant of the Baldy release in Tech/approach.md


ASawwaf:
@TFredberg, thanks for the great effort, just quick question, 1st, is this one aligned with @tomkivlin's diagram 
https://wiki.lfnetworking.org/display/LN/CNTT+RA2+-+Kubernetes+-+Diagrams?src=contextnavpagetreemode 

2nd , few comments
- The interface between CaaS and Bare Metal  should not be ( VI-HA ) as ( VI-HA ) is ETSI 
- 2 doted lines from NFVO's should be non-CNTT/ETSI NFV
- VNF-M to VNF and CNFM to CNF
- missing  EMS

PR #1782, opened 2020-07-13
[RA1] Update Chapter 5 to Train

collivier:
It updates microversions and links according to Train.
It should be noted that the last hot version is 2018-08-31 (rocky).

### DIFF
--- a/doc/ref_arch/openstack/chapters/chapter05.md
+++ b/doc/ref_arch/openstack/chapters/chapter05.md
@@ -316,7 +316,7 @@ Please note that the initial version of the [Cyborg API v1.0](https://docs.opens
 |---------------------------------|-----------------|
 | Key Manager: Barbican           | v1              |
 
-Barbican API Documentation: https://docs.openstack.org/barbican/pike/api/
+Barbican API Documentation: https://docs.openstack.org/barbican/latest/api/
 
 <!--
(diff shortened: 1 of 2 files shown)


pgoyal01:
[5.3.4 Barbican](https://github.com/cntt-n/CNTT/blob/master/doc/ref_arch/openstack/chapters/chapter05.md#534-barbican)
Barbican API Documentation: https://docs.openstack.org/barbican/pike/api/

The last URL needs to be replaced with https://docs.openstack.org/barbican/train/api/.

pgoyal01:
We need to add  Ironic with an API version of 1.58

collivier:
@pgoyal01 why not adding barbican and ironic but I would advice us to add them in their dedicated PRs. Here it's only about increasing the existing microversions and the capabilities optional due to partial supports.

collivier:
@pgoyal01 I'm still unable to add you as reviewer of the relevant Functest changes. Would you mind adding yourself as reviewer (you're not obliged to vote)? It may sync gerrit and LFN DB.
https://gerrit.opnfv.org/gerrit/c/functest/+/70542
https://gerrit.opnfv.org/gerrit/c/functest/+/70543
https://gerrit.opnfv.org/gerrit/c/functest/+/70544

PR #1801, opened 2020-07-15
[RA1 Ch2]: Fix link to chapter 5.3

CsatariGergely:


pgoyal01:
@CsatariGergely the original reference also works. You have taken the reference form the TOC while the original is from the actual section within the text. Given that there is no error, do we still want this change?

rabi-abdel:
@pgoyal01 please approve

pgoyal01:
@rabi-abdel Thanks. BTW This change wasn't needed. Both references are correct using the link specified in the ToC or the actual reference to the section in Github.

Issue #1637, opened 2020-05-28
[RA2 Ch04] Define configuration parameters of CNI plugins


CsatariGergely:
As it was discussed in the [RA2 meeting on 2020.05.28](https://wiki.lfnetworking.org/display/LN/2020-05-28+-+%5BCNTT+-+RA2%5D+-+Meeting+Agenda+and+Minutes) RA2 should specify the CNI plugin parameters what are mandatory to set for an RA2 compliant infrastructure.

tomkivlin:
Some suggested write-up for networking:
[RA2-Ch4-Networking-Specs-2020-07-17.docx](https://github.com/cntt-n/CNTT/files/4938945/RA2-Ch4-Networking-Specs-2020-07-17.docx)

_Originally posted by @pgoyal01 in https://github.com/cntt-n/CNTT/pull/1805#issuecomment-660207366_

PR #1805, opened 2020-07-16
[RA2 Ch4] Add Host OS and Kubernetes specs

### LINKED ISSUE #1634: [RA2 Ch04] Define configuration parameters of Host OS
As it was discussed in the [RA2 meeting on 2020.05.28](https://wiki.lfnetworking.org/display/LN/2020-05-28+-+%5BCNTT+-+RA2%5D+-+Meeting+Agenda+and+Minutes) RA2 should specify the host OS parameters what are mandatory to set for an RA2 compliant infrastructure.

### LINKED ISSUE #1635: [RA2 Ch04] Define configuration parameters of Kubernetes
As it was discussed in the [RA2 meeting on 2020.05.28](https://wiki.lfnetworking.org/display/LN/2020-05-28+-+%5BCNTT+-+RA2%5D+-+Meeting+Agenda+and+Minutes) RA2 should specify the Kuberntes parameters what are mandatory to set for an RA2 compliant infrastructure.

tomkivlin:
Also wrapped lines to 80 chars (except those in tables).

Fixes #1634 
Fixes #1635


doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -18,26 +18,42 @@
...
+virtual machine and an installed Operating System. In order for a Container Host
+to be conformant with the Reference Architecture it must be implemented as per
+the following specifications:
 
 |Ref|Specification|Details|Requirement Trace|
 |---|---|---|---|
-|`ra2.ch.001`||||
-|`ra2.ch.002`||||
-|`ra2.ch.003`||||
+|`ra2.ch.001`|Huge Pages|Huge Pages must be enabled within the Container Host OS (e.g. to verify run `cat /proc/meminfo ¦ grep Huge`)|[infra.com.cfg.004](./chapters/chapter02.md#223-cloud-infrastructure-software-profile-requirements)|
```
pgoyal01, line 50: do we need to specify the huge page sizes, number, etc.?

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -18,26 +18,42 @@
...
+|`ra2.ch.002`|SR-IOV capable NICs|The physical machine on which the Container Host runs must be equipped with NICs that are SR-IOV capable (e.g. [this list from Intel](../../../../../../../../intel/sriov-network-device-plugin#supported-sr-iov-nics)).|[e.cap.013](./chapters/chapter02.md#223-cloud-infrastructure-software-profile-requirements)|
```
pgoyal01, line 51: do we need a requirement to enable the SR-IOV plugin?

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -18,26 +18,42 @@
...
+|`ra2.ch.003`|SR-IOV Virtual Functions|SR-IOV virtual functions (VFs) must be configured within the Container Host OS, as the SR-IOV Device Plugin does not manage the creation of these VFs. (e.g. [vf-setup.md](../../../../../../../../intel/sriov-network-device-plugin/blob/master/docs/vf-setup.md)).|[e.cap.013](./chapters/chapter02.md#223-cloud-infrastructure-software-profile-requirements)|
+|`ra2.ch.004`||||
```
pgoyal01, line 53: In this section, do you want to add info CPU isolation, CPU dedication, CPU pinning (implemented in two parts: ensuring that guests can only run on dedicated cores; and ensuring that common host processes do not run on those dedicated cores)

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: I did wonder about that.  There isn't anything in the RM that specifies that, only that huge pages needs to be supported.  I guess for K8s we can narrow it down to 2Mi and 1Gi, but otherwise there's nothing that tells us we need to include a quantity of each of them.  I guess for the purposes of the specification, we need to ask that at least one of each is configured and the success criteria for a test would be that at least one of each is shown as allocatable.  I'll modify as such.

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: So the way my brain has been thinking about this is:
- There are a bunch of requirements relating to SR-IOV (and any other Device Plugin) that belong in the Kubernetes section (kubelet options, etc.)
- Then there are a bunch of requirements that are more directly related to the configuration of the Networking part - i.e. use Multus/DANM/neither, configure your CNI like this, configure the SR-IOV Device Plugin like this, etc.

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: Yes absolutely.  My knowledge on these subjects is not great, so help would be appreciated.  Need to consider for both bare metal and virtualised nodes.

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: I've updated in https://github.com/cntt-n/CNTT/pull/1805/commits/2bf586b978a739fe6d39643c498d46a09c2cc6cd to read:

> Huge Pages must be enabled within the Container Host OS, with at least one each of 2048KiB and 1048576KiB page sizes, exposing schedulable resources `hugepages-2Mi` and `hugepages-1Gi`.

doc/ref_arch/kubernetes/chapters/chapter04.md
pgoyal01: @tomkivlin Will send you a file --  something that could be adapted; the last page has the .md table format. Tried to attach file here but it failed.

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -18,26 +18,45 @@
...
+to be conformant with the Reference Architecture it must be implemented as per
+the following specifications:
 
 |Ref|Specification|Details|Requirement Trace|
 |---|---|---|---|
-|`ra2.ch.001`||||
-|`ra2.ch.002`||||
-|`ra2.ch.003`||||
+|`ra2.ch.001`|Huge Pages|Huge Pages must be enabled within the Container Host OS, with at least one each of 2048KiB and 1048576KiB page sizes, exposing schedulable resources `hugepages-2Mi` and `hugepages-1Gi`.|[infra.com.cfg.004](./chapter02.md#223-cloud-infrastructure-software-profile-requirements)|
+|`ra2.ch.002`|SR-IOV capable NICs|The physical machine on which the Container Host runs must be equipped with NICs that are SR-IOV capable (e.g. [this list from Intel](../../../../../../../../intel/sriov-network-device-plugin#supported-sr-iov-nics)).|[e.cap.013](./chapter02.md#223-cloud-infrastructure-software-profile-requirements)|
```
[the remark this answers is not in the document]
tomkivlin: Seems to work from here: https://github.com/cntt-n/CNTT/blob/tomkivlin-ra2-ch4-fixes1634/doc/ref_arch/kubernetes/chapters/chapter04.md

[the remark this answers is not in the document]
tomkivlin: So should work when merged with master as /blob/.../doc/ is the same for both variants.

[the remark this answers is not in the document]
tomkivlin: Or is this more a problem for e.g. read-the-docs work you're doing?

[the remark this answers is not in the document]
CsatariGergely: Yes, this will be a problem for the readthedocs work. We should either use absolult links, like https://github.com/intel/sriov-network-device-plugin#supported-sr-iov-nics , or I can craft just an other Shinx plugin to fix these during the compilation.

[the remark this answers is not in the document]
tomkivlin: I'll use absolute links, no problem.

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -18,26 +18,45 @@
...
+virtual machine and an installed Operating System. In order for a Container Host
+to be conformant with the Reference Architecture it must be implemented as per
+the following specifications:
 
 |Ref|Specification|Details|Requirement Trace|
 |---|---|---|---|
-|`ra2.ch.001`||||
-|`ra2.ch.002`||||
-|`ra2.ch.003`||||
+|`ra2.ch.001`|Huge Pages|Huge Pages must be enabled within the Container Host OS, with at least one each of 2048KiB and 1048576KiB page sizes, exposing schedulable resources `hugepages-2Mi` and `hugepages-1Gi`.|[infra.com.cfg.004](./chapter02.md#223-cloud-infrastructure-software-profile-requirements)|
```
Levovar, line 50: are you trying to say that all containers hosts must have hugepages and SR-IOV enabled all the time, or that it should be possible to enable these features on all hosts?
the former isn't really true, as majority of the applications do not require hugepages, or SR-IOV. if you enable hugepages on a host which does not hold DPDK apps that memory is effectively wasted

it would be better to say these feature must be possible to be configured for any host profiles IMO

doc/ref_arch/kubernetes/chapters/chapter04.md
Levovar, line 51: might want to include a Mellanox link too for proper vendor representation :)

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -18,26 +18,45 @@
...
+|`ra2.ch.002`|SR-IOV capable NICs|The physical machine on which the Container Host runs must be equipped with NICs that are SR-IOV capable (e.g. [this list from Intel](../../../../../../../../intel/sriov-network-device-plugin#supported-sr-iov-nics)).|[e.cap.013](./chapter02.md#223-cloud-infrastructure-software-profile-requirements)|
+|`ra2.ch.003`|SR-IOV Virtual Functions|SR-IOV virtual functions (VFs) must be configured within the Container Host OS, as the SR-IOV Device Plugin does not manage the creation of these VFs. (e.g. [vf-setup.md](../../../../../../../../intel/sriov-network-device-plugin/blob/master/docs/vf-setup.md)).|[e.cap.013](./chapter02.md#223-cloud-infrastructure-software-profile-requirements)|
```
Levovar, line 52: same as with hugepages. SR-IOV is only needed on some host profiles, not on all

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -18,26 +18,45 @@
...
+|`ra2.ch.004`|CPU Simultaneous Multi-Threading (SMT)|SMT must be enabled in the BIOS on the physical machine on which the Container Host runs.|[infra.hw.cpu.cfg.004](./chapter02.md#224-cloud-infrastructure-hardware-profile-requirements)|
+|`ra2.ch.005`|CPU Allocation Ratio|For Kubernetes nodes running as Virtual Machines, ensure the CPU allocation ratio between vCPU and physical CPU is 1:1 (note that it is not possible for the kubelet to over-allocate CPU resources.)|[infra.com.cfg.001](./chapter02.md#223-cloud-infrastructure-software-profile-requirements)|
```
Levovar, line 54: 1: I guess you wanted to say thread, not core. In case HT is enabled 1 vCPU is 1 thread usually
2: not just that it is possible, it is the default setting :) deploy any Pods without limits, start up some infinite loops and watch them eat the resources
requests cannot be overallocated sure, but limits very much can be more than the available physical CPU shares

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -79,8 +98,16 @@ In order for the Kubernetes components to be conformant with the Reference Archi
...
+|`ra2.k8s.007`|DevicePlugins Feature Gate|In order to support the various acceleration devices required, the DevicePlugins feature gate must be enabled (note, this is enabled by default in Kubernetes v1.10 or later). `--feature-gates="...,DevicePlugins=true,..."`|Various, e.g. [e.cap.013, e.cap.014](./chapters/chapter02.md#221-cloud-infrastructure-software-profile-capabilities)|
+|`ra2.k8s.008`|System Resource Reservations|To avoid resource starvation issues on nodes, reserve compute resources for system daemons and Kubernetes system daemons such as kubelet, container runtime, etc. (requires Kubernetes version 1.17 or later). Use the following kubelet flags: `--reserved-cpus=[0-3]`|NB: There doesn't appear to be a requirement that this is for?  Do we need to be specific about the values used?  The guidelines [here](https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/#general-guidelines) suggest a more nuanced approach is recommended?|
```
Levovar, line 103: number of CPUs to be reserved always depends on the stuff running on the specific infrastructure. even within the same product differetn configurations can be imagined, requiring different numbers (with / without Ceph etc.)

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -79,8 +98,16 @@ In order for the Kubernetes components to be conformant with the Reference Archi
...
+|`ra2.k8s.009`|CPU Pinning|In order to support CPU Pinning, the kubelet must be started with the `--cpu-manager-policy=static` option. (Note, only containers in `Guaranteed` pods - where CPU resource `requests` and `limits` are identical - and configured with positive-integer CPU `requests` will take advantage of this. All other Pods will run on CPUs in the remaining shared pool.)|[infra.com.cfg.003](./chapter02.md#223-cloud-infrastructure-software-profile-requirements)|
+|`ra2.k8s.010`||||
```
Levovar, line 105: reserving CPU is nice, how about system reserving memory too?

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -152,7 +180,8 @@ To support the isolation of the resources used by the infrastructure from the re
 <a name="4.5"></a>
 ## 4.5 Networking solutions
```
Levovar, line 181: any requirements regarding a proper network blueprint? separate VLANs for infra, proper bonding, physical separation of infra and app traffic etc.?

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: That's a great point - I will re-word this and other statements.

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -79,8 +98,16 @@ In order for the Kubernetes components to be conformant with the Reference Archi
...
-|`ra2.k8s.006`||||
-|`ra2.k8s.007`||||
+|`ra2.k8s.006`|NUMA Support|The `TopologyManager` and `CPUManager` feature gates must be enabled and configured on the kubelet (note, TopologyManager is enabled by default in Kubernetes v1.18 and later, with CPUManager enabled by default in Kubernetes v1.10 and later). `--feature-gates="...,TopologyManager=true,CPUManager=true" --topology-manager-policy=single-numa-node --cpu-manager-policy=static`|[e.cap.007](./chapter02.md#221-cloud-infrastructure-software-profile-capabilities) [infra.com.cfg.002](./chapter02.md#223-cloud-infrastructure-software-profile-requirements) [infra.hw.cpu.cfg.004](./chapter02.md#224-cloud-infrastructure-hardware-profile-requirements)|
+|`ra2.k8s.007`|DevicePlugins Feature Gate|In order to support the various acceleration devices required, the DevicePlugins feature gate must be enabled (note, this is enabled by default in Kubernetes v1.10 or later). `--feature-gates="...,DevicePlugins=true,..."`|Various, e.g. [e.cap.013, e.cap.014](./chapters/chapter02.md#221-cloud-infrastructure-software-profile-capabilities)|
```
gkunz, line 102: broken link

```suggestion
|`ra2.k8s.007`|DevicePlugins Feature Gate|In order to support the various acceleration devices required, the DevicePlugins feature gate must be enabled (note, this is enabled by default in Kubernetes v1.10 or later). `--feature-gates="...,DevicePlugins=true,..."`|Various, e.g. [e.cap.013, e.cap.014](./chapter02.md#221-cloud-infrastructure-software-profile-capabilities)|
```

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: I agree!  Ideally there would be a vendor-agnostic list somewhere.  The Intel link above includes Mellanox cards, but perhaps not all of them.

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: Interesting, I was under the impression that the "Allocatable" was a hard limit, regardless of the requests made by Pods (i.e. allocatable would get used, next Pod would be stuck in pending state?)

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: Yeah I'm a little nervous about including this tbh (partly as you've described, partly because there is nothing in chapter02.md suggesting it's a requirement, partly that it feels a little out of scope of what we're trying to focus on in CNTT).

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: Don't think it's possible in Kubernetes, but I agree - it's as important.

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: The requirements are in chapter02.md - what we need to add into this table is the configuration specification for the Kubernetes networking solution.  So yes to what you've said - but we need to define the configuration spec in this chapter, based on requirements defined in chapter02.md

doc/ref_arch/kubernetes/chapters/chapter04.md
Levovar: yeah my bad, haven't cehcked the link. but for example ConnectX6 series is missing from the list, yep :)
I guess any such list is destined to go out of date so might be better to just state he generic requirement

doc/ref_arch/kubernetes/chapters/chapter04.md
Levovar: only if your Pod state "requests". if you don't state a request you will be scheduled

Levovar: from a live system:
Allocated resources:
  (Total limits may be over 100 percent, i.e., overcommitted.)
  Resource                        Requests     Limits
  --------                        --------     ------
  cpu                             1628m (3%)   7498m (17%)

The help text itself says limits might be overcommited :)

tomkivlin: thanks - will tweak the wording

doc/ref_arch/kubernetes/chapters/chapter04.md
Levovar: yeah I would suggest to keep it on the "configure enough CPU cores for your host platform via --reserved-cpus"
additional note: you separately mention static policym ust be enabled so functionally speaking --reserved-cpus will work, but it worth noting --reserved-cpus **only works** with static CPU policy

doc/ref_arch/kubernetes/chapters/chapter04.md
Levovar: --system-reserved Kubelet flag :)

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: OK I think I understand now - please correct any mistakes!  
- Sum(`limits`) > `allocatable` is valid.
- Sum(`requests`) > `allocatable` is not valid.
- If you do specify `limits` but don't specify `requests` then `requests`==`limits` and over-commitment isn't possible (because sum(`requests`) must be < `allocatable`).
- If you do specify `requests` < `limits` then you are in the shared CPU pool, and over-commitment (i.e. sum(`limits`) > allocatable) is possible as long as sum(`requests`) < `allocatable`.

Is that right?

Levovar: not sure if request = limit is added to a Pod wrt CPU, but otherwise correct, yes

doc/ref_arch/kubernetes/chapters/chapter04.md
pgoyal01: We need to configure [hugepages](https://kubernetes.io/docs/tasks/manage-hugepages/scheduling-hugepages/).
OVS-DPDK: need to pre-allocate from hugepages on specific sockets for running OvS-DPDK processes; I understand, we can specify up to 8 1G hugepages using "dpdk-socket-mem=".

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: The last example [here](https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy) suggests:

> This pod runs in the Guaranteed QoS class because only limits are specified and requests are set equal to limits when not explicitly specified.

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: @pgoyal01 - `ra2.ch.001` should cover configuration within the host OS.  Let me know if that's not quite right.  The requirements I pulled through into chapter02.md didn't include a number of each hugepage size, but if there are other requirements that require a particular number of them then we need to include that and reference those reqs from chapter02.md as well.

> |`ra2.ch.001`|Huge Pages|It must be possible to enable Huge Pages (2048KiB and 1048576KiB) within the Container Host OS, exposing schedulable resources `hugepages-2Mi` and `hugepages-1Gi`.|

pgoyal01: @tomkivlin I was just replying to the previous comments about memory reservation.  Ra2.ch.001 is correct. Thanks

pgoyal01:
@tomkivlin Some suggested write-up for networking:
[RA2-Ch4-Networking-Specs-2020-07-17.docx](https://github.com/cntt-n/CNTT/files/4938945/RA2-Ch4-Networking-Specs-2020-07-17.docx)

tomkivlin:
> @tomkivlin Some suggested write-up for networking:
> [RA2-Ch4-Networking-Specs-2020-07-17.docx](https://github.com/cntt-n/CNTT/files/4938945/RA2-Ch4-Networking-Specs-2020-07-17.docx)

Thanks Pankaj - I've shared this in #1637 as that will focus on the networking part.

doc/ref_arch/kubernetes/chapters/chapter04.md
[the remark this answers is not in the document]
tomkivlin: Updated

tomkivlin: Removed link to Intel doc in https://github.com/cntt-n/CNTT/commit/b90ac99e4a8e53cb1ab94949f32ab66b5578f75d.

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: Updated in https://github.com/cntt-n/CNTT/commit/28f8ac972d3cca675e95bf11220c1c4317be2a74

tomkivlin:
@gkunz thanks, but this is still WIP - plenty more changes to come yet :)

tomkivlin:
@rabi-abdel this isn't ready for review yet - PR is still in draft state.

rabi-abdel:
> @rabi-abdel this isn't ready for review yet - PR is still in draft state.

Yes, apologies, I missed that.

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: Levente, does this new wording work for you? https://github.com/cntt-n/CNTT/pull/1805/commits/a2904850d7710add072f05421b377c069eac190e

Levovar: so yeah it is factually correct now. remaining question is semantic based :)
if we truly want to control overcommitment ratio, I think a similar sentence for the summarized limits might be also required
i.e. requests and limits both must be defined for all containers, and summa requests must be lower than allocatable, and summa limits must be lower than 2Xallocatable -> guaranteed 1:2 overcommitment ratio

if this RQ is about controlling overcommitment ratio then limits should be taken into account too IMO; if not then nevermind :)

tomkivlin: > i.e. requests and limits both must be defined for all containers, and summa requests must be lower than allocatable, and summa limits must be lower than 2Xallocatable -> guaranteed 1:2 overcommitment ratio

Except that the RM states that allocation ratio must be 1:1 for all profiles.  I'll update to reflect what you've mentioned about limits - agreed.

Levovar: yeah then the same, but summa limits must equal summa requests

tomkivlin: OK, updated in https://github.com/cntt-n/CNTT/pull/1805/commits/13b0d5834bea31bd72b5951116734dd273dbdb58

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -79,8 +99,16 @@ In order for the Kubernetes components to be conformant with the Reference Archi
...
+|`ra2.k8s.008`|System Resource Reservations|To avoid resource starvation issues on nodes, reserve compute resources for system daemons and Kubernetes system daemons such as kubelet, container runtime, etc. (requires Kubernetes version 1.17 or later). Use the following kubelet flags: `--reserved-cpus=[0-3]`|NB: There doesn't appear to be a requirement that this is for?  Do we need to be specific about the values used?  The guidelines [here](https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/#general-guidelines) suggest a more nuanced approach is recommended?|
+|`ra2.k8s.009`|CPU Pinning|In order to support CPU Pinning, the kubelet must be started with the `--cpu-manager-policy=static` option. (Note, only containers in `Guaranteed` pods - where CPU resource `requests` and `limits` are identical - and configured with positive-integer CPU `requests` will take advantage of this. All other Pods will run on CPUs in the remaining shared pool.)|[infra.com.cfg.003](./chapter02.md#223-cloud-infrastructure-software-profile-requirements)|
```
tomkivlin, line 105: @michaelspedersen - this is the spec I've written for CPU pinning so far.  As mentioned in the other PR, it may be that there are more specific requirements we need to add into chapter02.md in order to drive a more specific decision about using CMK over the native CPU manager.

tomkivlin: i.e. do we need something in chapter02.md along the lines of "The architecture must support the capability to ensure data plane CNFs are pinned to specific CPU cores"?  To me, the generic "CPU Pinning" requirement from the RM isn't specific enough to help (me anyway) determine whether the native CPU manager is sufficient.

Levovar: in a K8s environment apps should never assume the capability of requesting specific cores
capability to request NUMA aligned cores? sure. and this is supported by K8s CPU Manager
capability to request cores from a special, optimized pool? now this is where it gets tricky :) because DPDK apps might legit need this, but upstream legit does not support this RQ.
ofc you can always separate these apps to their own nodes, and apply low latency tunings to **all** application cores, but that is al ess dynamic approach. and has some OPEX aspect to it (increased consumption, inability to collocate different apps therefore need for more physical servers in the DC etc.)

tomkivlin: aye - that's what I was starting to think.  Feels like the native CPU manager meets the requirements that are currently defined.

michaelspedersen: I'm wondering what the best approach would be.

The requirement in the RM seems sufficient for VNFs, as the workload only sees the resources that are provided to the VM. i.e. If you assign 3 CPUs (or cores) to the VM, `lscpu` will only show you that many resources. For CNFs the view from inside the container will look like the host, so even with 3 CPUs assigned you might still see 48 CPUs through `lscpu`.

Using [CMK](https://github.com/intel/CPU-Manager-for-Kubernetes) as an example, it adds an extra layer of pinning/isolation through CPU affinity, compared to the native CPU Manager which only limits the processor cycles that each pod has access to. 
My concern is the way that the pinning is enforced. CMK provides a list of assigned cores/threads to the container through environment variables, and these can then be used through either 1) a provided wrapper that sets the affinity of the application or workload running in the container, or 2) to map into an configuration file that accepts a specific cores/threads for processing (i.e. VPP, OVS or generic DPDK applications). This puts responsibility on the workload to abide by these assignments, as I believe it will still be possible to override this assignment.

It might be good to add a requirement to Ch2 that specifies pinning to specific CPUs, but I wonder if it should be a MUST or a SHOULD (especially if it relies on the workload to function properly)

michaelspedersen: > 
> 
> in a K8s environment apps should never assume the capability of requesting specific cores
> capability to request NUMA aligned cores? sure. and this is supported by K8s CPU Manager
> capability to request cores from a special, optimized pool? now this is where it gets tricky :) because DPDK apps might legit need this, but upstream legit does not support this RQ.
> ofc you can always separate these apps to their own nodes, and apply low latency tunings to **all** application cores, but that is al ess dynamic approach. and has some OPEX aspect to it (increased consumption, inability to collocate different apps therefore need for more physical servers in the DC etc.)

I agree that apps shouldn't be able to request specific cores, but as you also mention DPDK apps usually need specific, and typically isolated, cores.

For me the requirement will depend a lot on the target users. For computing this would (almost) never be an issue, but for telcos I can see this being a requirement due to SLAs, in particular ones related to jitter/latency.

pgoyal01: My understanding (please correct), the native CPU Manager doesn't really offer true isolation among user pods; exclusion is for the Kubernetes processes. CMK, on the other hand, provides isolation even among user pods.

tomkivlin: > My understanding (please correct), the native CPU Manager doesn't really offer true isolation among user pods; exclusion is for the Kubernetes processes. CMK, on the other hand, provides isolation even among user pods.

Perhaps (I don't know).  Point is, we don't have a requirement stating that is required, we just have one that says CPU pinning is required.  Do we need more detailed requirements such as those I suggest above, or e.g. "The architecture must support the capability for workloads to be scheduled to specific and isolated CPU cores."?

pgoyal01: @tomkivlin There are additional requirements that do not specify the level of isolation but could still be applicable, for example the following:

Reference | Description | Requirement (common to all Profiles)
-- | -- | --
e.man.004 | Capability to isolate resources between tenants | Must support
sec.wl.006 | Workloads must be separable by their categorisation (for example, payment card information, healthcare, etc.) |

Levovar: IMO you are overmistiyfing this a little bit. as someone who authored one CPU Manager solution, I think the situation is actually simple
if you have any of the following 2 requirements defined:
- possibility for apps to request special exclusive cores, different from other cores (differently tuned than rest on the node)
- possibility for a container to request exclusive **and** shared resources at the same time
than, and only than you need a custom CPU manager.
If you don't have any of these requirements, you don't need a custom CPU manager.

Sidenote: if you need a custom CPU Manager, CMK is not the only option. if you ask me it is not even the best workaround for multiple reasons, but again, I'm biased, obviously (https://github.com/nokia/CPU-Pooler)
at the very minimum just mention that "well, because of these requirements you prob need a special CPU Manager, up to you, just satisfy the requirements"

tomkivlin: So right now, we don't have either requirement, so the native CPU manager is fine.  I'm cool with that.  My question (which came from some specs that were written in an RI2 PR) was whether the first of your bullets was a requirement, that we hadn't documented yet.  If so, we need to add it to chapter02.md and as you rightly say, deal with it appropriately.  If not, then the RI2 PR needs an update as the CMK isn't required.

Levovar: > My understanding (please correct), the native CPU Manager doesn't really offer true isolation among user pods; exclusion is for the Kubernetes processes. CMK, on the other hand, provides isolation even among user pods.

Upstream CPU manager also can and does isolate exclusive cores from all others.
Only containers running on shared cores, well, share their cores. if your container is sensitive to context switches an argument can be made that you should be requesting exclusive cores in the first place, rather than trying to force the existence of "multiple shared pools" into the infrastructure. If your app is not sensitive to switches, they should be perfectly fine running on a shared cpuset, as there are enough administrative tools in pace to guarantee that you are getting the **amount** of CPU shares you require from the same shared pool.

Levovar: > So right now, we don't have either requirement, so the native CPU manager is fine. I'm cool with that. My question (which came from some specs that were written in an RI2 PR) was whether the first of your bullets was a requirement, that we hadn't documented yet. If so, we need to add it to chapter02.md and as you rightly say, deal with it appropriately. If not, then the RI2 PR needs an update as the CMK isn't required.

I have personally met some data plane applications which did require it. Basically DPDK-only use-case
An another issue if you want satisfy this via providing different exclusive pools on the same node, or optimizing all cores on dedicated nodes as explained above
if it is anyway planned to physically isolate these senstivie workloads then this special pooling might not be required.
on the other hand, there might not be enough nodes isolate these workloads on the far edge
one can argue either way. of course I'm personally well documented and recorded arguing that we need multiple pools, but it certainly has downsides as well (static pool config increase chances of resource fragmentation for example)

Levovar: to put things into real-life context :)
Openshift AFAIK uses upstream CPU Manager only, and solves the issue of data plane apps by 1: defining --reserve-cpus 2: applying the "sepcial config" for all cores (isolcpus basically)
products and projects I'm involved with support the concept of multiple pools, but as an optional, opt-in, per-host group configurable feature (i.e. enable it only on nodes it is required)
not sure about other products, but both approaches are definitely used on the field, so I'm not sure it is wise, or possible to rule out either. flexibility might be the key

michaelspedersen: My takeaway from the discussion is to keep the requirements as already listed (CPU pinning based on native CPU manager). Any additional pinning through e.g. CMK or CPU-Pooler is out of scope, but can be applied as needed. Does this sound right?

tomkivlin: Yep, I agree with that.

pgoyal01:
@tomkivlin the following specification would require changes to RM:

ra2.ch.011 | Virtual Node CPU Quantity | If using VMs, the Kubernetes Nodes must be equipped  with at least 16 vCPUs. Note, extra should be provisioned to cater for  any overhead required by the Operating System and any required OS  processes such as the container runtime, Kubernetes agents, etc. | e.cap.001
-- | -- | -- | --


RM [4.2.1.1 Predefined Compute Flavours](https://github.com/cntt-n/CNTT/blob/master/doc/ref_model/chapters/chapter04.md#4211-predefined-compute-flavours) Table 4-13 in the Notes specifies that the .2xlarge (8 vCPUs), 4xlarge (16 vCPUs) and ,8xlarge (32 vCPUs) "... are intended to be used for transitional purposes ..." 

@tomkivlin @walterkozlowski Given the need for larger number of vCPUs needed for certian compute intensive workloads should RM be modified to remove the "transition use" restriction?

tomkivlin:
> @tomkivlin the following specification would require changes to RM:
> 
> ra2.ch.011	Virtual Node CPU Quantity	If using VMs, the Kubernetes Nodes must be equipped with at least 16 vCPUs. Note, extra should be provisioned to cater for any overhead required by the Operating System and any required OS processes such as the container runtime, Kubernetes agents, etc.	e.cap.001
> RM [4.2.1.1 Predefined Compute Flavours](https://github.com/cntt-n/CNTT/blob/master/doc/ref_model/chapters/chapter04.md#4211-predefined-compute-flavours) Table 4-13 in the Notes specifies that the .2xlarge (8 vCPUs), 4xlarge (16 vCPUs) and ,8xlarge (32 vCPUs) "... are intended to be used for transitional purposes ..."
> 
> @tomkivlin @walterkozlowski Given the need for larger number of vCPUs needed for certian compute intensive workloads should RM be modified to remove the "transition use" restriction?

@pgoyal01 not sure I 100% understand.  Are you saying the misalignment is because of the "transition use" clause, or because 16vCPUs isn't as big as the maximum flavour?

tomkivlin:
@CsatariGergely could you review/approve please?

petorre (approved):

CsatariGergely (approved):

tomkivlin:
@scottsteinbrueck this is ready for your review/merge please.

Issue #1734, opened 2020-06-29
[RI2 Ch04] Kubernetes provisioning


michaelspedersen:
Add content to 4.3.2 following the work done in #1529

PR #1830, opened 2020-07-22
[RI2 Ch4.3.2] Kubernetes Provisioning

michaelspedersen:
Issue #1734 

Adds details to Chapter 4.3.2 about provisioning Kubernetes using BMRA

### DIFF
--- a/doc/ref_impl/cntt-ri2/chapters/chapter04.md
+++ b/doc/ref_impl/cntt-ri2/chapters/chapter04.md
@@ -55,7 +55,154 @@ After the hosts have been provisioned successfully, one can set up host networki
 <a name="4.3.2"></a>
 ### 4.3.2 Kubernetes Provisioning
 
-> Add detailed description of current K8s provisioning
+Early efforts to provision Kubernetes are based on existing tools such as [Bare Metal Reference Architecture (BMRA)](https://builders.intel.com/docs/networkbuilders/container-bare-metal-for-2nd-generation-intel-xeon-scalable-processor.pdf), using scripts available on [Github](https://github.com/intel/container-experience-kits), and the CNCF initiative [CNF Testbed](https://github.com/cncf/cnf-testbed). 
+
+Requirements are based on the [Kubernetes Based Reference Architecture (RA2)](../../../ref_arch/kubernetes), with initial focus on hardware and functionality related to CPU and networking. While this will not be sufficient to satisfy all the requirements listed in RA2, it will serve as a development platform for further feature integration, verification and ideally testing related to the [Kubernetes Based Reference Conformance (RC2)](../../../ref_cert/RC2).
+
+The Kubernetes provisioning follows the approach described in [Chapter 1](chapter01.md#13-reference-implementation-approach) of this RI. Steps for preparing the BMRA environment can be found here: [CNTT RI2 Confluence page](https://wiki.lfnetworking.org/display/LN/Kubernetes+Bare-Metal+Install+and+Configuration). 
+
+Make sure all servers are reachable from the jumphost through SSH. If necessary, create a key and push it to the master and worker nodes
+```
+$ ssh-keygen -t rsa -b 4096
+$ ssh-copy-id -i <path to public key> <user>@<host>
+```
+
+The following paragraphs will use v1.4.1 of the BMRA, and focus on configuration options that are relevant in terms of the RA2 and RI2.
+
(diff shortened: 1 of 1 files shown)


tomkivlin:
lgtm - great start!

doc/ref_impl/cntt-ri2/chapters/chapter04.md
```diff
@@ -55,7 +55,154 @@ After the hosts have been provisioned successfully, one can set up host networki
...
+  # Install CPU Manager for Kubernetes (CMK), which provides core affinity and isolation of workloads
+cmk_hosts_list: <csv of hosts where CMK should be deployed>
+  # Update the list if CMK is enabled
+cmk_shared_num_cores: <#>
+cmk_exclusive_num_cores: <#>
+  # Number of cores that will be added to the shared and exclusive pools of CMK
+```
+[SR-IOV Network device plugin for Kubernetes](https://github.com/intel/sriov-network-device-plugin):
+```
+sriov_net_dp_enabled: true
```
rihabbanday, line 88: Do we need to enable cpu manager & sr-iov plugins for every deployment?

michaelspedersen: Looking at the RA2, I would say CPU Manager is needed for CPU pinning support. One could argue that the Kubernetes CPU resource provides "pinning" through requests/limits, but from my experience this is not always sufficient (in particular when aiming for low latency).

SR-IOV on the other hand could be disabled. It is allowed based on an [exception in the RM](https://github.com/cntt-n/CNTT/blob/master/doc/ref_model/chapters/appendix-a.md#a3-exception-list), but as there are no requirements specific to performance (throughput/latency) it could be disabled in favor of kernel based networking using CNIs. There are options for using Userspace CNIs (with OVS and VPP), but I have not done much testing with those myself

rihabbanday: I see, atm I've set these features to false in bmra based Kuberef deployments, so was wondering if these are mandatory requirements. Thanks for the clarification.

tomkivlin: Re. CPU Manager - I'm not 100% clear on whether CMK is needed over the native CPU Manager, given the requirements we have.  i.e. can we achieve CPU pinning with native CPU manager, or do we need CMK?  I'm finding it hard to reconcile the documentation I've been reading with the requirements we have in RA2 chapter 2.

michaelspedersen: @rihabbanday Nothing in the configuration should be mandatory/required. Though there are some dependencies between parameters that might cause weird behavior or errors (e.g. enabling SR-IOV CNIs without enabling SR-IOV)

@tomkivlin I think it depends a lot on use-cases and scenarios. CMK adds additional functionality over the native manager, as it adds a list of assigned cores/threads to the container environment, which is very useful when running applications such as VPP, where you can specify exactly what cores you want your management and dataplane threads running on. It also adds a wrapper that can be used to `taskset` an application to the assigned cores. 

If we settle on what to include in a "default" configuration I can update the steps accordingly. I would however like to avoid having too many multiple-choice variables due to the dependencies between them.

tomkivlin: Re. the choices in this RI, I would suggest enabling anything that we need to test, even if it the RA2 spec is worded as a "if you need x, enable it by doing a, b and c".  A real deployment would only ever enable it if needed, but for the purposes of a platform for software companies to develop against etc, I think enabling by default makes sense, to me.

tomkivlin: > @tomkivlin I think it depends a lot on use-cases and scenarios. CMK adds additional functionality over the native manager, as it adds a list of assigned cores/threads to the container environment, which is very useful when running applications such as VPP, where you can specify exactly what cores you want your management and dataplane threads running on. It also adds a wrapper that can be used to `taskset` an application to the assigned cores.

I will pull you into the PR I'm drafting in RA2 - it sounds like we need more requirements than we have.

michaelspedersen: I've updated the configuration example to not include CMK by default. As mentioned in https://github.com/cntt-n/CNTT/pull/1805 I believe the native CPU Manager is sufficient for RA2/RI2, with the possibility of using CMK or CPU-Pooler for additional isolation if needed

PR #1872, opened 2020-08-07
[RA2 Ch02] 2.3 Additional Networking requirements

### LINKED ISSUE #1866: [RA2 Ch02] Network interfaces
Since RM Chapter 2 doesn't specify the support for multiple network interfaces, we need a set of requirements for the same,  for example:

Ref # | Category | Sub-category | Description
-- | -- | -- | --
req.inf.ntw.11 | Infrastructure | Network | The Architecture **must** support multiple network interfaces to be configured for a pod 
req.inf.ntw.12 | Infrastructure | Network | The networking solution to support  req.inf.ntw.11 **must** conform to the Kubernetes Networking CRD de-facto standard (K8S Network Plumbing Working Group) 
req.inf.ntw.13 | Infrastructure | Network | The networking solution to support  req.inf.ntw.11 **must** support any Kubernetes conformant CNI plugin to be incorporated as a networking solution
req.inf.ntw.14 | Infrastructure | Network | A tenant **must** specify at least one IP pool that is different from the control infrastructure IP pool or other tenant IP pools

pgoyal01:
Fixes #1866


doc/ref_arch/kubernetes/chapters/chapter02.md
```diff
@@ -304,6 +304,10 @@ The requirements in this section are to be delivered in addition to those in [se
 | `req.inf.ntw.08` | Infrastructure | Network | The Architecture **must**  provide a default network which implements the Kubernetes network model. | |
 | `req.inf.ntw.09` | Infrastructure | Network | The networking solution **must not** interfere with or cause interference to any interface or network it does not own. | |
 | `req.inf.ntw.10` | Infrastructure | Network | The Architecture **must** support cluster wide coordination of IP address assignment. | |
+| `req.inf.ntw.11` | Infrastructure 	Network 	The Architecture must support multiple network interfaces to be configured for a pod. | |
```
tomkivlin, line 307: Not sure this is needed.  In the RM we have the following requirements covering multiple network interfaces, which is pulled through in section 2.2.1 of this chapter.

> e.cap.004 | # Connection points | number | Max number of connection points that can be assigned to a single VM or Pod by the Cloud Infrastructure

doc/ref_arch/kubernetes/chapters/chapter02.md
```diff
@@ -304,6 +304,10 @@ The requirements in this section are to be delivered in addition to those in [se
...
+| `req.inf.ntw.12` | Infrastructure 	Network 	The networking solution to support req.inf.ntw.11 must conform to the Kubernetes Networking CRD de-facto standard (K8S Network Plumbing Working Group). | |
```
tomkivlin, line 308: Could you provide a link to the documentation of the standard please?

tomkivlin, line 308:
```suggestion
| `req.inf.ntw.12` | Infrastructure |	Network | The networking solution to support req.inf.ntw.11 must conform to the Kubernetes Networking CRD de-facto standard (K8S Network Plumbing Working Group). | |
```

doc/ref_arch/kubernetes/chapters/chapter02.md
```diff
@@ -304,6 +304,10 @@ The requirements in this section are to be delivered in addition to those in [se
...
+| `req.inf.ntw.13` | Infrastructure 	Network 	The networking solution to support req.inf.ntw.11 must support any Kubernetes conformant CNI plugin to be incorporated as a networking solution. | |
```
tomkivlin, line 309:
```suggestion
| `req.inf.ntw.13` | Infrastructure | Network | The networking solution to support req.inf.ntw.11 must support any Kubernetes conformant CNI plugin to be incorporated as a networking solution. | |
```

doc/ref_arch/kubernetes/chapters/chapter02.md
```diff
@@ -304,6 +304,10 @@ The requirements in this section are to be delivered in addition to those in [se
...
+| `req.inf.ntw.14` | Infrastructure 	Network 	A tenant must specify at least one IP pool that is different from the control infrastructure IP pool or other tenant IP pools. | |
```
tomkivlin, line 310:
```suggestion
| `req.inf.ntw.14` | Infrastructure | Network | A tenant must specify at least one IP pool that is different from the control infrastructure IP pool or other tenant IP pools. | |
```

doc/ref_arch/kubernetes/chapters/chapter02.md
pgoyal01: @tomkivlin Thanks -- missed that in my search. Deleted

doc/ref_arch/kubernetes/chapters/chapter02.md
pgoyal01: @tomkivlin Added. Thanks

doc/ref_arch/kubernetes/chapters/chapter02.md
```diff
@@ -304,6 +304,9 @@ The requirements in this section are to be delivered in addition to those in [se
 | `req.inf.ntw.08` | Infrastructure | Network | The Architecture **must**  provide a default network which implements the Kubernetes network model. | |
 | `req.inf.ntw.09` | Infrastructure | Network | The networking solution **must not** interfere with or cause interference to any interface or network it does not own. | |
 | `req.inf.ntw.10` | Infrastructure | Network | The Architecture **must** support cluster wide coordination of IP address assignment. | |
+| `req.inf.ntw.11` | Infrastructure |	Network | The networking solution to support multiple connection points (e.cap.004 - please see [Section 2.2.1](#2.2.1)) must conform to the Kubernetes Networking CRD de-facto standard ([K8S Network Plumbing Working Group](https://github.com/k8snetworkplumbingwg/community); [CRD](https://docs.google.com/document/d/1Ny03h6IDVy_e_vmElOqR7UdTPAG_RNydhVE1Kx54kFQ/edit)). | |
```
CsatariGergely, line 307: This spec unfortunately never made it to be a Kubernetes API, therefore it _not_ supported by Kubernetes so I suggest to remove the refernce to it.

pgoyal01: @CsatariGergely I am very confused. I thought the purpose of a CRD was to provide API extensibility and flexibility. Don't the CRDs extend the Kubernetes APIs?

CsatariGergely: Yes, CRD-s are to extend the Kubernetes API and this is exactly what my point is. The CRD-s specified in the linked document were never implemented as an in tree API in Kubernetes.

tomkivlin:
@pgoyal01 are you able to address @CsatariGergely comments?

doc/ref_arch/kubernetes/chapters/chapter02.md
pgoyal01: @CsatariGergely Not sure what you are trying to state here:
- as you know, a CRD doesn't affect the default Kubernetes installation so the default Kubernetes API tree will be unaffected
- when a CRD is incorporated in a custom installation (using kubectl apply), the API tree does get affected as it then includes the new API endpoints

tomkivlin: @pgoyal01 could you commit my suggestion above, and then we can see if there are any objections to that?

tomkivlin:
@CsatariGergely are you ok with the current wording?

walterkozlowski (approved):

doc/ref_arch/kubernetes/chapters/chapter02.md
CsatariGergely: My point was that the CRD-s defined by the network plumbing WG are not part of upstream Kubernetes there fore I would not say that it is a de-facto standard. The defined API does not provide an abstract way to represent network needs, but provides direct access to the underlying CNI-s configuration (You can imagine it as the SR-IOV or metaplugin CRD-s).
I'm okay with the wording to use CRD-s in general.

gkunz (approved):

doc/ref_arch/kubernetes/chapters/chapter02.md
pgoyal01: @CsatariGergely Isn't a CRD by definition not part of upstream kubernetes?

tomkivlin: OK here's another thought - does this requirement limit us in terms of the possible solutions available to us (e.g. NSM, which could provide the interface via the Device Plugin API)?  I don't think "metaplugin" is the only possible solution to multiple connection points?

petorre (approved):

doc/ref_arch/kubernetes/chapters/chapter02.md
tomkivlin: @CsatariGergely are you ok with the updated wording, or still a problem?

scottsteinbrueck (commented):
@tomkivlin @CsatariGergely @pgoyal01 Please resolve and merge, or remove Milestone=M3.  Thank you.

tomkivlin:
@CsatariGergely are you ok with the new wording, or not?  If so please remove the request for change.

doc/ref_arch/kubernetes/chapters/chapter02.md
CsatariGergely: @tomkivlin  yes, the current wording limits us to Multus as the only Implementation of the Network Plumbing WG-s spec.  I'm not ok with the current wording as it still requires compliancy with the Network Plumbing WG-s CRD spec what is designed in a way that the users can bypass Kubernetes security framework.
Without `req.inf.ntw.11` I'm okey with the changes.

tomkivlin: @pgoyal01 could you either address Gergely's concern by removing .011 so we can merge for Baraque, or remove the "Baraque" label and M3 milestone so we deal with it in the next release.

collivier (approved):

doc/ref_arch/kubernetes/chapters/chapter02.md
pgoyal01: @tomkivlin It is a CRD spec. The wording nowhere limits it to the MULTUS implementation.  Shouldn't the purpose of our specification be to rely on other specs. Maybe others should be asked to chime in on this particular requirement.

For now I am removing the Baraque and M3 label.

tomkivlin: First up, yes - we should rely on other specs but only where they are either widely adopted or a formal standard (in my opinion).  For me, the Kubernetes networking space is still settling and I'm not sure restricting the networking solutions that could be used is useful as it'll stifle innovation in a fast-moving area.

Re. the wording, the working group refers to Multus as "the" Reference Implementation of the CRD spec ([here](https://github.com/k8snetworkplumbingwg/community#working-items)).  So yes I agree it doesn't limit an operator to use Multus, but I suspect various valid solutions such as e.g. NSM would not be useable as a network solution, even if it met all other requirements.  As you say, it would be useful for others to chime in.

doc/ref_arch/kubernetes/chapters/chapter02.md
```diff
@@ -304,6 +304,9 @@ The requirements in this section are to be delivered in addition to those in [se
...
 | `req.inf.ntw.09` | Infrastructure | Network | The networking solution **must not** interfere with or cause interference to any interface or network it does not own. | |
 | `req.inf.ntw.10` | Infrastructure | Network | The Architecture **must** support cluster wide coordination of IP address assignment. | |
+| `req.inf.ntw.11` | Infrastructure |	Network | The networking solution to support multiple connection points (e.cap.004 - please see [Section 2.2.1](#2.2.1)) must conform to the [Multi-Network CRD Specification from the Kubernetes Network Plumbing Working Group](https://github.com/k8snetworkplumbingwg/multi-net-spec). | |
+| `req.inf.ntw.12` | Infrastructure | Network | The networking solution to support multiple connection points (e.cap.004 - please see [Section 2.2.1](#2.2.1)) must support any Kubernetes conformant CNI plugin to be incorporated as a networking solution. | |
+| `req.inf.ntw.13` | Infrastructure | Network | A tenant must specify at least one IP pool that is different from the control infrastructure IP pool or other tenant IP pools. | |
```
ulikleber, line 309: Not sure it makes sense to specify a requirement against a tenant here. It cannot be fulfilled by a platform.

pgoyal01: @ulikleber The multiple separate IP pools are supported by the Platform. Do you have another wording that possibly states this better.

ulikleber: > @ulikleber The multiple separate IP pools are supported by the Platform. Do you have another wording that possibly states this better.

What about: "The platform must allow specifying multiple separate IP pools."
The platform cannot make the tenant follow the guidelines that are in the wording.

pgoyal01: @ulikleber Thanks - please review. Changed to:

"The platform must allow specifying multiple separate IP pools. Tenants are required to select at least one IP pool that is different from the control infrastructure IP pool or other tenant IP pools."

ulikleber (approved):

doc/ref_arch/kubernetes/chapters/chapter02.md
tomkivlin: I would like to reach consensus on this.  For me, a spec from the working group set up to create Multus isn't a good enough source or standard.  If it were to come from SIG-Network then it's a different story.  As per the Tech Steering call just now, we know there is a gap when it comes to the standard management of multiple interfaces.  My suggestion would be to create an issue that gets us to document this gap in ch6, and we remove the CRD part of this change - the rest I'm fine with.  @pgoyal01 @CsatariGergely @walterkozlowski @gkunz @petorre @collivier @ulikleber  do you agree?

tomkivlin: @pgoyal01 correct, I already have an open issue for that: https://github.com/cntt-n/CNTT/issues/1911

pgoyal01: @tomkivlin Can you remove the hold and also let the associated PR #1912 get approved now so that the Baraque release is consistent.

pgoyal01: @tomkivlin  req.inf.ntw.11 commented out: "The networking solution to support multiple connection points (e.cap.004 - please see [Section 2.2.1](#2.2.1)) must conform to the [Multi-Network CRD Specification from the Kubernetes Network Plumbing Working Group](https://github.com/k8snetworkplumbingwg/multi-net-spec)."

gkunz: @tomkivlin yes, I kind of agree with this conclusion for the time being. I wasn't able to attend today's TSC, but the biggest issue for me is that each implementation (Multus, DANM, NSM) currently brings its own API (CRD) as all of those follow (slightly) different designs. Can we build a complex system without standardized APIs? No. Are the currently existing solutions good enough to solve the Kubernetes telco networking problem fully? Probably not. So, this is a good example of where CNTT's approach of building specs from existing solutions is challenged and we certainly need to step up our efforts of "influencing" the technical development of upstream communities.

CsatariGergely (approved):

Issue #1876, opened 2020-08-12
[RM] Add new requirement to address networking scale


sukhdevkapur:
Networking should scale across multiple clouds. There should be an ability to federate networks spanning across multiple deployments. Many cloud native deployments in 5G telecom space will be desegregated and span across multiple sites

tomkivlin:
Changing this to target RM changes.

wmk-admin:
@sukhdevkapur, can you raise this topic at a post-Elbrus RM meeting to clarify what in your view needs to be done?

walterkozlowski:
RM Meeting on 30 June decided to close Issue#1876 "address networking scale" and add this a subtopic in the Multi-cloud topic

PR #1877, opened 2020-08-12
[RA2 Ch2] Adding a new networking requirement to address scale

sukhdevkapur:
This addresses issue #1876

### DIFF
--- a/doc/ref_arch/kubernetes/chapters/chapter02.md
+++ b/doc/ref_arch/kubernetes/chapters/chapter02.md
@@ -304,6 +304,7 @@ The requirements in this section are to be delivered in addition to those in [se
 | `req.inf.ntw.08` | Infrastructure | Network | The Architecture **must**  provide a default network which implements the Kubernetes network model. | |
 | `req.inf.ntw.09` | Infrastructure | Network | The networking solution **must not** interfere with or cause interference to any interface or network it does not own. | |
 | `req.inf.ntw.10` | Infrastructure | Network | The Architecture **must** support cluster wide coordination of IP address assignment. | |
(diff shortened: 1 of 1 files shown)


tomkivlin:
@sukhdevkapur I fully support the intent of this PR.  Couple of comments:
1. Should this be at RM level (i.e. could it apply to all RAs)?
1. I think we need to be a bit more specific though, perhaps with multiple requirements.  For example, in the original issue you already have two (_Networking should scale across multiple clouds. There should be an ability to federate networks spanning across multiple deployments._) but more importantly what do we want to be able to scale?  Would a networking solution that has a single "visibility" plane but multiple disconnected control and data planes for each cloud count?  Or are we talking about being able to logically connect and control the networks from multiple clouds at e.g. IP layer from a single place?  Or something else?

I don't think we need loads more, but just a couple/few to explain what we want from multi-cloud networking would help determine the solutions.

pgoyal01:
@sukhdevkapur @tomkivlin asked a couple of questions that remain un-addressed. My vote would be to add these requirements to the RM chapter and then they can be incorporated in the RAs.

If you do not want to add in RM then the requirement should be added in [Section 2.3](https://github.com/cntt-n/CNTT/blob/sukhdevkapur-patch-5/doc/ref_arch/kubernetes/chapters/chapter02.md#2.3) as Section 2.2 is titled [Reference Model Requirements](https://github.com/cntt-n/CNTT/blob/sukhdevkapur-patch-5/doc/ref_arch/kubernetes/chapters/chapter02.md#2.2)

walterkozlowski:
@sukhdevkapur @tomkivlin @pgoyal01: I think that these are good candidates to be added to RM actaually

tomkivlin:
OK thanks.  I have renamed the issue (#1876) and moved to RM project.  Closing this PR.

PR #1793, opened 2020-07-14
[Glossary] Alphabetical order and editorial changes in the glossary

EditKoselak:
Closes #1792

PR #1908, opened 2020-08-24
[Glossary] Generalizing the definition of the Virtual Application

### LINKED ISSUE #1905: [Glossary] Generalise to make a VA run on a VM or container
As discussed in #1793, let's generalize the definition of Virtual Application (VA) in the glossary.

EditKoselak:
Closes #1905


doc/common/glossary.md
```diff
@@ -32,7 +32,7 @@ To help guide the reader, this glossary provides an introduction to the terminol
   - **VNF**s are built from one or more VNF Components (**VNFC**) and, in most cases,  the VNFC is hosted on a single VM or Container.
 - **Cloud Native Network Function (CNF)**: A cloud native network function (CNF) is a cloud native application that implements network functionality. A CNF consists of one or more microservices and has been developed using Cloud Native Principles including immutable infrastructure, declarative APIs, and a “repeatable deployment process”.
   >_*Note:*_ This definition is derived from the [Cloud Native Thinking for Telecommunications Whitepaper](https://github.com/cncf/telecom-user-group/blob/master/whitepaper/cloud_native_thinking_for_telecommunications.md#1.4), which also includes further detail and examples.
-- **Virtual Application (VA)**: A general term for software which can be loaded into a Virtual Machine.
+- **Virtual Application (VA)**: A general term for software which can be loaded into a Virtual Machine or a set of Containers.
```
pgoyal01, line 35: @CsatariGergely why not just containers as in other definitions:
```suggestion
- **Virtual Application (VA)**: A general term for software which can be loaded into  Virtual Machines or Containers.
```

ulikleber: Agree to @pgoyal01 : .. into Virtual Machines or Containers.
(VA may be multiple VMs, "set of" may confuse).

tomkivlin: As per my comments in the original PR (https://github.com/cntt-n/CNTT/pull/1793#discussion_r454384438) I disagree with this change.  

I know that by a dictionary definition, "virtualised" covers both VM-based and container-based however I think it completely misses the point of cloud native / containerisation, the benefits it brings and the changes needed at the software level.  "Virtualised" in my sense (meaning software that has been developed with VMs in mind) still relies on many traditional and non-cloud native concepts such as infra-level SLAs at the VM level (hence why vMotion and similar tech exists) and mutable infrastructure (i.e. changes made to the software in the VM after deployment).  The current definition of Virtual Application works for me as it covers that level of "maturity" or cloud nativeness (/nativity?).  

I'd prefer to see a definition for Cloud Native Application be added / used rather than fitting cloud native concepts into a previous generation of software architectures.  This was also the argument behind the change from NFVI to Cloud Infrastructure in this same page.

walterkozlowski: @tomkivlin: I agree with the need to define Cloud Native Application; it should be done.  But does it preclude us from defining Virtual Application as proposed? We probably need a generic term for an application that can be run in a VM or a container.

tomkivlin: > @tomkivlin: I agree with the need to define Cloud Native Application; it should be done. But does it preclude us from defining Virtual Application as proposed? We probably need a generic term for an application that can be run in a VM or a container.

Personally I think it should preclude us, yes.  As we chose to do for NFVI -> Cloud Infrastructure instead of including CaaS in NFVI, I think it's important to keep the Virtual/Virtualised/Virtualisation terms for things that are about VMs, hypervisors, etc. and for the Cloud/Containerised/etc. things to be covered by separate definitions, given the complete different in software architecture.  

Perhaps the term "Application" could cover both?

walterkozlowski: Yes, perhaps we should just speak about Applications, if the term Virtual creates such VM like connotations. 
Personally, I have much broader connotations with the term "virtual", e.g. virtual reality.  But I am not religious about things like that.

pgoyal01: Do we use Virtual Application anywhere in the document? If yes, can we change to "workload"?
Later we can delete this definition.

FYI: "Virtual Application" definition was originally taken from an ETSI document and by changing it we have already moved away from a standard.

walterkozlowski: @pgoyal01: ETSI have changed their definitions since then. 
However, I agree that using "workload" is the most neutral.

EditKoselak: We already have a definition for "workload", so I assume we can just remove this definition. The term "Virtual Application" is not used in other parts of the documentation.

tomkivlin (approved):

ulikleber (approved):

doc/common/glossary.md
EditKoselak: @walterkozlowski, I think this conversation can be resolved as we agreed on a solution. Can we merge this pull request? :)

petorre (approved):

scottsteinbrueck (approved):

scottsteinbrueck:
@EditKoselak please merge and bring to the master.

EditKoselak:
> @EditKoselak please merge and bring to the master.

Thanks, done:)

Issue #1874, opened 2020-08-10
[Tech Steering] To what level do we go to ensure "consistency"


tomkivlin:
**Question for Tech Steering**: we need a decision as to whether it is ONLY the container images (i.e. the software) that must be abstracted from the infrastructure, or whether the deployment artefacts, which are often developed alongside the software by the software vendor, are also included in the scope of this abstraction.

Background
---
Within RA2, because of how tightly coupled the application lifecycle management artefacts (e.g. Helm) is to the development of the software by vendors (i.e. they release both together), a question has arisen over time as to the level to which we decide to want to avoid "options" in RA2 Chapter04 (Architecture Specification).  In particular, it comes down to:
1. Whether the container images differ (i.e. whether different container images are required for the specification provided)
1. Whether the deployment artefacts differ (i.e. whether different YAML configurations - not just variables but included options - are required for the specification provided)

I think everyone is agreed that 1. above is definitely what the RA2 specification is looking to avoid.  I'm not sure there is full agreement/acceptance of 2. above and hence the question to steering group.

e.g.

For multi-NIC networking solution, we could go through all the requirements and decide that the only metaplugin that meets them all is DANM.  We specify that in the RA2 spec, meaning all implementations (the RI, any VIs, etc.) must use DANM, and all CNFs that require multiple connection points must use DANM.  The benefits of this are that there is consistency and everyone knows what they're building for - in particular the container images _and_ deployment artefacts are the same across all implementations.  The downsides are that this then removes all the other potential solutions to providing multiple connections points - whether that be another metaplugin (e.g. Multus) or another approach entirely (e.g. CRDs/NSM), which in turn could reduce the overall benefit seen within the community (i.e. people stop using RA2 as a specification and we're back to square one).

[...]

PR #1912, opened 2020-08-25
[ra2 ch4] CNI multiplexer wording

### LINKED ISSUE #1911: [RA2 Ch4] Alternative wording for CNI multiplexer

tomkivlin:
Closes #1911


doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -156,7 +156,7 @@ In order for the networking solution(s) to be conformant with the Reference Arch
...
 |---|---|---|---|
-|`ra2.ntw.001`|CNI multiplexer/metaplugin|As the selected CNI multiplexer/metapulgin MUST support other CNI plugins (`req.inf.ntw.06`) and SHOULD provide an API based solution to administer the networks from a central point (`req.inf.ntw.03`) the selected CNI multiplexer/metapulgin may be [DANM](https://github.com/nokia/danm). For more detailed feature comparision of CNI multiplexers/metaplugins see Table 4-5| [req.inf.ntw.06](./chapter02.md#23-kubernetes-architecture-requirements), [req.inf.ntw.03](./chapter02.md#23-kubernetes-architecture-requirements) |
+|`ra2.ntw.001`|CNI multiplexer/metaplugin|The selected CNI multiplexer/metapulgin MUST be conformant with the CNI spec, MUST support other CNI plugins (req.inf.ntw.06), and SHOULD provide an API based solution to administer the networks from a central point (req.inf.ntw.03). Example CNI implementations which meet these requirements include:<br> - DANM<br> - Multus.<br><br>For more detailed feature comparision of CNI multiplexers/metaplugins see Table 4-5| [req.inf.ntw.06](./chapter02.md#23-kubernetes-architecture-requirements), [req.inf.ntw.03](./chapter02.md#23-kubernetes-architecture-requirements) |
```
CsatariGergely, line 159: Let's wait until #1874 concluded.

CsatariGergely (changes requested):

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: @CsatariGergely are you ok with this, given:
- the discussion we had at TSC
- I will raise an issue I will raise to document the gap in abstracted management of multi-NIC solutions
- we remove the CRD requirement from #1872

pgoyal01 (approved):

doc/ref_arch/kubernetes/chapters/chapter04.md
CsatariGergely: Yes if we somehow enable deferring from [principle 8](https://github.com/cntt-n/CNTT/blob/master/doc/common/chapter00.md#overall-principles) and we list the gaps of Multus compared to the CNTT requirements (no central IPAM, non abstract API (DANM has an abstract API))

pgoyal01: @CsatariGergely  I am sure the IPAM requirements can be satisfied by a stand-alone CNI IPAM plugin. W.r.t. the meta plugin again it may not be necessary to create another set of API but just conform to the CNI APIs.

collivier (approved):

tomkivlin:
@CsatariGergely can you check if you're happy with this please?  Just mentioned in the RA2 call that I'll start the voting process if we can't gain consensus through this PR.
> https://github.com/cntt-n/CNTT/blob/master/doc/gov/chapters/chapter05.md

tomkivlin:
FWIT, I agree with Pankaj in that I don't believe all the networking requirements need to be delivered by the same single tool/solution.  I think it's fine for e.g. IPAM to be done by something different than e.g. multiple connection points to the pods.

doc/ref_arch/kubernetes/chapters/chapter04.md
CsatariGergely: IPAM issue: The problem with Muliplexers is that they manage multiple CNI-s, therefore without a central IPAM they are not capable to guarantee that there are no overlaps in the IP allocations. At the moment there is no mention about a separate central IPAM in the component list, so with Multus RA2 is incomplete.
API issue: CNI API is not something what you can manage centrally while req.inf.ntw.03 requires central administration and configuration. 

We should list all of these defects of the Multus based solution.

CsatariGergely:
> 
> 
> @CsatariGergely can you check if you're happy with this please? Just mentioned in the RA2 call that I'll start the voting process if we can't gain consensus through this PR.
> 
> > https://github.com/cntt-n/CNTT/blob/master/doc/gov/chapters/chapter05.md

If we list the missing features of the Multus based solution I'm okay.

CsatariGergely:
> 
> 
> FWIW, I agree with Pankaj in that I don't believe all the networking requirements need to be delivered by the same single tool/solution. I think it's fine for e.g. IPAM to be done by something different than e.g. multiple connection points to the pods.

I agree if that different thing can guarantee, that there are no overlaps between the IP allocations to the muliple underlying CNI-s. Also we need to describe that for the Multus alternative the infrastructure should also contain this different thing.

doc/ref_arch/kubernetes/chapters/chapter04.md
pgoyal01: @CsatariGergely The point is that there are requirements for IPAM, Security Policies, Multiplexers, etc. and these are all separate components needed for an overall networking solution. They are not defects in Multus or for that matter any other solution.

CsatariGergely:
> 
> 
> > FWIW, I agree with Pankaj in that I don't believe all the networking requirements need to be delivered by the same single tool/solution. I think it's fine for e.g. IPAM to be done by something different than e.g. multiple connection points to the pods.
> 
> I agree if that different thing can guarantee, that there are no overlaps between the IP allocations to the muliple underlying CNI-s. Also we need to describe that for the Multus alternative the infrastructure should also contain this different thing.

I've clarified these in #2104  and gapped them in #2105

CsatariGergely (approved):

petorre (approved):

PR #1922, opened 2020-08-27
[RM Appendix Part 3] Removing tech specific exceptions

walterkozlowski:

PR #1917, opened 2020-08-27
[Chapter 00] Principles update - Acceleration alignment

walterkozlowski:
This PR is meant to provide clarity and technology independence for the principles around dependencies between virtual and physical infrastructure, and network workloads in alignment with the evolved Reference Model and advances in technology. It moves technology dependent policies, exceptions and requirements into the technology dependent Reference Architectures specifications and the downstream implementation specs. It also changes the verbiage form the commandments to the principles.

PR #1925, opened 2020-08-27
[RM Ch04] Table 4-17 and forward compatibility updates

walterkozlowski:
Updated Table 4-17 and some contents in Sec 4.2.7.3 on forward compatibility removing technology specific aspects as per the alignment with Pr #1917 and #1922 around hardware accelerations

### DIFF
--- a/doc/ref_model/chapters/chapter04.md
+++ b/doc/ref_model/chapters/chapter04.md
@@ -586,9 +586,9 @@ Flavours are unique only when combined with a profile. For example, CNTT release
 `B_GenN.small`
 
 <a name="4.2.7.3"></a>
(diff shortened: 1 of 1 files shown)


tomkivlin (approved):

tomvanpelt (commented):
I am not finding the way to do proper suggestions, but "exceptions" is missing at the start of the second sentence. It could probably start "Technology specific **exceptions** are..." (i.e. without the "The").

Next to that there's some historical typos that should maybe be corrected as part of this update. The reference should be to section 9.4.3 instead of 9.2.3 and the section 4.2.7.3 title isn't marked properly ("####" is missing).

tomvanpelt (approved):

PR #1969, opened 2020-09-07
[RM Ch3] Add Section 3.6 Storage

### LINKED ISSUE #1943: [RM Ch3] Write section 3.x Storage
(As per discussions on RM calls) Write section 3.x Storage similar like there is [3.5 Network](https://github.com/cntt-n/CNTT/blob/master/doc/ref_model/chapters/chapter03.md#3.5) .

petorre:
(As per discussions on RM calls) This is new section 3.6 Storage similar like there is 3.5 Network .
Fixes #1943 .

### DIFF
--- a/doc/ref_model/chapters/chapter03.md
+++ b/doc/ref_model/chapters/chapter03.md
@@ -13,6 +13,8 @@
   * [3.3.2 Hardware Infrastructure Manager](#3.3.2)
 * [3.4 Hardware Infrastructure Resources](#3.4)
 * [3.5 Network](#3.5)
+* [3.6 Storage](#3.6)
+* [3.7 Sample reference model realization](#3.7)
 
 It is necessary to clearly define the infrastructure resources and their capabilities a shared cloud infrastructure (network function virtualisation infrastructure, NFVI) will provide for hosting workloads including virtual network functions (VNFs) and/or cloud-native network functions (CNFs). The lack of a common understanding of which resources and corresponding capabilities a suitable cloud infrastructure should provide may lead to several issues which could negatively impact the time and the cost for on-boarding and maintaining these solutions on top of a virtualised infrastructure. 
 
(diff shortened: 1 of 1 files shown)


walterkozlowski (approved):

ASawwaf (approved):

pgoyal01:
@petorre Should 3.5 Network and 3.6 Storage be made subsection of the 3.4 Hardware Infrastructure Resources as was the case for Virtual resources?

doc/ref_model/chapters/chapter03.md
```diff
@@ -263,7 +265,20 @@ Principles that should be followed during the development and definition of the
 * Future proof: Network model is extendible to support known and emerging technology trends including SmartNICs, FPGAs and Programmable Switches, integrated for multi-clouds, and Edge related technologies.
 
 <a name="3.6"></a>
-## 3.6 Sample reference model realization
+## 3.6 Storage
```
pgoyal01, line 268: @petorre do some of the capabilities specified in Table 3.3 also need to be covered here. For example, data availability, performance and enhanced features?

sukhdevkapur (approved):

petorre:
> 
> 
> @petorre Should 3.5 Network and 3.6 Storage be made subsection of the 3.4 Hardware Infrastructure Resources as was the case for Virtual resources?
> @pgoyal01: No, these talk about Storage and Network in general, not only from the h/w point of view. It is good as it is, in my view (and that how we decided some time ago). Perhaps in the future we will rewrite the whole Chapter, to align with the industry evolution, but that's for the future.

I was thinking about similar from visual flow, but reading 3.5 Network didn't get the feeling that it was written with that in mind.
Agree it is for future. I'd be happy to look at all this again later.

doc/ref_model/chapters/chapter03.md
petorre: I was thinking about that also, but to keep it short only wrote "_various storage types and performance requirements (more in Virtual Resources 3.2.3 Storage)_" with link to text above on the same page.

PR #1972, opened 2020-09-07
[Glossary] Clarification of Kubernetes Cluster

### LINKED ISSUE #1970: [Glossary] Clarification of Kubernetes Cluster in the Glossary
As discussed in #1793, the definition of the Kubernetes Cluster should be clarified.

EditKoselak:
Closes #1970


tomkivlin:
I'm not sure this PR is necessary, but being pedantic, I recall the original change was because there was no definition of "machine" in the glossary.  I note there is also no definition of "server" - what's the need for the change?

EditKoselak:
> I'm not sure this PR is necessary, but being pedantic, I recall the original change was because there was no definition of "machine" in the glossary. I note there is also no definition of "server" - what's the need for the change?

I resolved the conflicts just in case... I hope @pgoyal01 would take a look at your question:)

walterkozlowski:
already deployed

Issue #1966, opened 2020-09-07
[RM Ch2.2] Issues with Use Case section


walterkozlowski:
-	There’s 2 sections numbered 2.2 
-	For the figures included in the first section 2.2 a different style of including is used than for others in the document which will results in them not coming across automatically in the conversion to GSMA format. 
@ASawwaf, since you added these use cases to Chapter 2, could you please fix this?  Thanks.

PR #1971, opened 2020-09-07
[RM CH02] Renumbering CH02 RM

ASawwaf:

PR #1973, opened 2020-09-07
[RM CH02 ] Use cases renumbering 

ASawwaf:
https://github.com/cntt-n/CNTT/issues/1966

### DIFF
--- a/doc/ref_model/chapters/chapter02.md
+++ b/doc/ref_model/chapters/chapter02.md
@@ -226,7 +226,7 @@ The list of, most likely to be virtualised,  Network Functions below, covering a
     
 
 <a name="2.2"></a>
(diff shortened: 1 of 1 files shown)


pgoyal01:
@ASawwaf The Table of Contents needs to be corrected : Use Cases needs to be added and the numbering changed

TFredberg (approved):
I just requested a change on PR #1971 before I could approve it, since it only changed the ToC and now this PR is the other way around.
It is very confusing to have one PR that changes the ToC but not the sub chapters numbering and another PR that changes the sub chapters numbering, but not the ToC!
With that comment I Approve this.

petorre (approved):

walterkozlowski (approved):

PR #2018, opened 2020-09-15
[RA2] remove ref to rm appendix and move exceptions to chapter01.md

### LINKED ISSUE #2017: [RA2 Appendix A] wrong reference to RM
Appendix A - Guidance For application Vendors in RA-2 points in section A.1 Exception List to RM:
"The exception list is collected in Appendix A - Guidelines For Application Vendors of the Reference Model.", which is not the case any more.  Exceptions are now treated within each RA because of the technology specific nature of the exceptions (see e.g. Issue  #1924). Consequently the subsection on exceptions in the RM  Appendix A was removed.
@tomkivlin: pleas remove the link to RM Appendix A from RA-2 https://github.com/cntt-n/CNTT/blob/master/doc/ref_arch/kubernetes/chapters/appendix-a.md

tomkivlin:
Fixes #2017

### DIFF
--- a/doc/ref_arch/kubernetes/chapters/appendix-a.md
+++ b/doc/ref_arch/kubernetes/chapters/appendix-a.md
@@ -4,21 +4,15 @@
 
 ## Table of Contents
 
-* [A.1 Exception List](#A.1)
-* [A.2 Guidance for VNF-to-CNF Transition](#A.2)
-* [A.3 TBC](#A.3)
-* [A.4 TBC](#A.4)
+* [A.1 Guidance for VNF-to-CNF Transition](#A.2)
+* [A.2 TBC](#A.3)
+* [A.3 TBC](#A.4)
 
 <a name="A.1"></a>
-## A.1 Exception List
-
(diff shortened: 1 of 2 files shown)


pgoyal01:
@tomkivlin I am approving but not sure if you also want "version exceptions". In k8s that could include allowing use of technologies and services in, say, alpha or beta phases.

pgoyal01 (approved):

tomkivlin:
> @tomkivlin I am approving but not sure if you also want "version exceptions". In k8s that could include allowing use of technologies and services in, say, alpha or beta phases.

I did wonder that - I took it out for now as I think that's best controlled in the spec (ch4).  i.e. the "version" of Kubernetes will be dictated by the k8s version control policy (n-2) however the status of a specific feature within that version isn't really a version exception and is more a choice we make in the spec - i.e. should a certain feature gate be used, despite it being in alpha state.

Happy to add in if people feel strongly, or at a later stage if needed.

wmk-admin (approved):

walterkozlowski (approved):

PR #2024, opened 2020-09-16
[RM Ch07] Proofreading

karinesevilla:
Typos 
Missing punctuation
References in section 7.9.8


zwarico (commented):
I agree with all the changes. Thank you for catching all of these small things.

wmk-admin (approved):

PR #2069, opened 2020-10-15
[RTD]: Using all sphinx extensions from pip

CsatariGergely:
All the tree Sphinx modules implemented are moved out to separate projects and used from pip.


walterkozlowski (approved):

tomkivlin (approved):

pgoyal01 (approved):

gkunz (approved):

ulikleber (approved):

collivier (changes requested):
Would it be possible to run pep8 and pylint on the py scripts (I quickly saw a couple of issues)?
It don't like so much defining concrete dependencies in requirements (I would prefer a PBR like design leveraging in upper-constraints to pin software version).

I would prefer not to merge with any pep8 and pylint issues. The upper-constraints part could be postponed if it helps.
Let me know if you need any help here.

PR #2080, opened 2020-10-22
[RM Ch5] Removing process

### LINKED ISSUE #2079: [RM Ch5] Removing process

EditKoselak:
Closes #2079


walterkozlowski (approved):

ulikleber (approved):
CPU pinning however could also be relevant on other layers. 
Here we talk infra requirements, it is correct to to mention only vCPU "being pinned".

petorre (approved):

PR #2093, opened 2020-11-09
[RM Ch2] Update list of network functions 

### LINKED ISSUE #2066: [RM Ch2.1] Review network applications used in the analysis
From the vF2F in October 2020:

> I note the presence of OCS in 3G/4G but no reference to the CHF/CCS in 5G.  I understand that many Edge based scenarios are expecting deployment of user plane functions but there are also discussions that control plane functions (which includes CHF - Charging Function) might make sense to bring to the edge.

In addition, the "Gathering and Validating CNF Requirements..." session agreed that this section needs a review to ensure any further analysis takes an up-to-date list of network applications as input.

Reference: https://etherpad.opnfv.org/p/Gathering_and_Validating_CNF_Requirements

Smitholi67:
Fixes #2066 
Added CHF/CCS to list of 5G Core NFs (Control Plane).

### DIFF
--- a/doc/ref_model/chapters/chapter02.md
+++ b/doc/ref_model/chapters/chapter02.md
@@ -75,6 +75,7 @@ The list of, most likely to be virtualised,  Network Functions below, covering a
     - UDR: Unified Data Repository
     - NRF:  Network Repository Function
     - NEF:  Network Exposure Function
+    - CHF/CCS:  Charging Function / Converged Charging System
     > _**Note:** for Service-based Architecture (SBA) all Network Functions are stateless (store all sessions/ state on unified data repository UDR)_
 - **IP Multimedia Subsystem (IMS)**
   - Data Plane
--- a/doc/ref_model/chapters/chapter02.md
+++ b/doc/ref_model/chapters/chapter02.md
@@ -75,7 +75,8 @@ The list of, most likely to be virtualised,  Network Functions below, covering a
     - UDR: Unified Data Repository
     - NRF:  Network Repository Function
     - NEF:  Network Exposure Function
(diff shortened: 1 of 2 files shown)


doc/ref_model/chapters/chapter02.md
```diff
@@ -75,6 +75,7 @@ The list of, most likely to be virtualised,  Network Functions below, covering a
     - UDR: Unified Data Repository
     - NRF:  Network Repository Function
     - NEF:  Network Exposure Function
+    - CHF/CCS:  Charging Function / Converged Charging System
```
pgoyal01, line 78: please split CHF and CCS into their own separate rows.

Smitholi67: I can do that, but CHF is a part of the CCS,  CHF interacts on the SBA, the CCS does not.  To me it makes more sense to keep together, or just drop CCS.

pgoyal01: Agree that they are related -- but we have other related functions listed separately.  

It is your decision to make. 

BTW you will also need to add these to the abbreviations list

walterkozlowski (approved):

karinesevilla (commented):
@Smitholi67 @pgoyal01 
CCS is not a function, but a set of functions. 
If CHF is the only function relevant in our scope, I prefer to keep CCS with CHF and not split them.
May be a wording like: ' CHF: charging Function part of the converged charging system'

pgoyal01 (approved):

karinesevilla (approved):
@Smitholi67 Thanks for the modifications

PR #2104, opened 2020-11-19
[RA2 Ch4]: Clarifications of differences between CNI multiplexers

### LINKED ISSUE #1911: [RA2 Ch4] Alternative wording for CNI multiplexer

CsatariGergely:
Partially closes #1911 
As I propose in #1912 the differences between Multus and DANM should be more explained if we would like to allow the usage of both Multiplexers. 
I also think that we should list the gaps of Multus in the gaps chapter.

### DIFF
--- a/doc/ref_arch/kubernetes/chapters/chapter04.md
+++ b/doc/ref_arch/kubernetes/chapters/chapter04.md
@@ -162,19 +162,19 @@ Architecture they must be implemented as per the following specifications:
 | Requirement | Support in Multus | Support in DANM |
 |-------------|-------------------|-----------------|
 | The overlay network encapsulation protocol needs to enable ECMP in the underlay (`infra.net.cfg.002`) | Supported via another CNI plugin | Supported via another CNI plugin |
-| NAT (`infra.net.cfg.003`) | Supported via another CNI plugin | Supported |
+| NAT less connectivity (`infra.net.cfg.003`) | Supported via another CNI plugin | Supported |
 | Security Groups (`infra.net.cfg.004`) | Not supported | Not supported <sub>1)<sub> |
 | SFC support (`infra.net.cfg.005`) | Not relevant | Not relevant |
 | Traffic patterns symmetry (`infra.net.cfg.006`) | Not relevant | Not relevant |
 | Network resiliency (`req.inf.ntw.01`) | Supported | Supported |
 | Centrally administrated and configured (`req.inf.ntw.03`) | Not supported | Partially suported |
-| Dual stack IPv4 and IPv6 for Kubernetes workloads (`req.inf.ntw.04`) | Supported via another CNI plugin | Suported |
+| Dual stack IPv4 and IPv6 for Kubernetes workloads (`req.inf.ntw.04`) | Automatic management of the IP allocation of IPv4 and IPv6 addresses requires additional components | Suported |
 | Integrating SDN controllers (`req.inf.ntw.05`) | Supported via another CNI plugin | Supported via another CNI plugin |
 | More than one networking solution (`req.inf.ntw.06`) | Supported | Supported |
 | Choose whether or not to deploy more than one networking solution (`req.inf.ntw.07`) | Supported | Supported |
 | Kubernetes network model (`req.inf.ntw.08`) | Supported via another CNI plugin | Supported via another CNI plugin |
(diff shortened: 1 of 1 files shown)


pgoyal01:
I think the requirements should be for the multiplexer alone not features of available products. For example, since if multiple interfaces were not required, we would still need the requirements for NAT, IP management, etc., these should not be requirements for a multiplexer -- a multiplexer solution may still implement these, but IMO tehse are not inherent multiplexer requirements.

CsatariGergely:
> 
> 
> I think the requirements should be for the multiplexer alone not features of available products. For example, since if multiple interfaces were not required, we would still need the requirements for NAT, IP management, etc., these should not be requirements for a multiplexer -- a multiplexer solution may still implement these, but IMO these are not inherent multiplexer requirements.

Yes, I agree that providing NAT less connectivity and IP address management are not requirements for a multiplexer. The table does not say that these are mandatory requirements of a multiplexer, it only maps the different requirements to the allowed multiplexer solutions. 

We need to show in the RM, that in case of Multus there are some additional components needed to cover the requirements. As I'm not familiar with Multus based solutions I would like to ask Multus users to extend the table with references to concrete components and solutions.

tomkivlin:
I am planning another PR for after Kubecon.  The gist of it is that the multi-NIC thing is just one requirement of the whole networking solution.  As with any solution, it can be comprised of multiple components - it (the networking solution) doesn't have to be delivered by a single "thing".  However, our requirements rightly state that it must be managed centrally via the k8s API.  We also have to remember that not all network applications require multiple connections and so the multiplexer, as with other parts of the solution, is something that needs to be in place for certain applications and not in place for others.

walterkozlowski:
I support @tomkivlin and will look for the new PR.  We really must stop thinking in terms of the existing products (Multus or DAMN), but about the minimum requirements needed for the CNI multiplexers.  Appealing to @CsatariGergely for understanding....

sukhdevkapur:
> I support @tomkivlin and will look for the new PR. We really must stop thinking in terms of the existing products (Multus or DAMN), but about the minimum requirements needed for the CNI multiplexers. Appealing to @CsatariGergely for understanding....

sukhdevkapur:
I agree with @tomkivlin somewhat. If you are gong to create a new PR, my suggestion will be to stay away from specifically calling out multiplexers or even multiple CNIs. The requirement should be specified in a "cloud native" way - i.e. the framework must provide ability for the workloads to connect to multiple networks - without any regards to whether  a single CNI or multi CNIs are used to achieve this. As an example, in a 5G core deployment,  a workload providing AMF function will need to connect to gNB on N2, SMF on N11 and to PCF on N15. 

To achieve this one could use multiple CNIs or single CNI. For instance some operators are using Multux with multiple CNIs to achieve this connectivity where as if they are using Tungsten Fabric CNI, this can be achieved using single CNI. 

We, the CNTT body must not be going into the specifics of saying use Multux, DANM or anything else. We should simply specify the requirements.

tomkivlin (approved):

pgoyal01:
Based on @tomkivlin, @walterkozlowski and @sukhdevkapur comments, may I suggest that we 
1. modify the introduction to Table 4.4 and make clear that these requirements can be met by a single or multiple networking solutions.
2. Delete Table 4.5 "Comparision of CNI multiplexers/metaplugins"

rgstori:
to have a like-for-like comparison, would it make sense to propose multi-component solutions?
e.g. a multus+whereabouts solution may have (or not) an ipam feature comparable to danm?

karinesevilla (approved):
The chapter 4 should list solutions to meet the requirements listed in chapter 2.
It makes sense to describe the components

rgstori:
@CsatariGergely can you rebase this after https://github.com/cntt-n/CNTT/pull/2121 gets merged please?

pgoyal01:
Why are we comparing multiplexers but not say solutions for other CNIs?

tomkivlin:
Given the age I am going to force merge.

pgoyal01:
Still not sure why we have this section as there are no other comparisons. Also it suffers from the fact that it is comparing two products where they have made different choices on what to include beyond what a pure multiplexer would need based on the principle of "do one thing".

tomkivlin:
This is lagging behind #2121 which adds two more examples.  Doesn't really matter which gets merged first but this is just a small change to the wording that can be applied to either.

PR #2105, opened 2020-11-19
[RA2 Ch6]: Adding Multus gaps

### LINKED ISSUE #1911: [RA2 Ch4] Alternative wording for CNI multiplexer

CsatariGergely:
As Multus is added as an option for CNI multiplexer to RA2 its gaps have to be addressed.

Partially closes #1911

### DIFF
--- a/doc/ref_arch/kubernetes/chapters/chapter06.md
+++ b/doc/ref_arch/kubernetes/chapters/chapter06.md
@@ -109,15 +109,6 @@ Note: with an underlying IaaS this is possible, but then it introduces (undesira
 > **Gap description:** Allocation of hugepages from the same NUMA node as other resources of a Pod. To support this [cAdvisor needed a change to support NUMA](https://github.com/google/cadvisor/pull/2304). Changes in Kubernetes are planned to be implemented in the [Node Topology Manager](https://github.com/kubernetes/enhancements/issues/693).
 
 <a name="6.2.9"></a>
-### 6.2.9 Missing central IPAM solution in Multus
-
-> **Related requirements:** `req.inf.ntw.04` and `req.inf.ntw.10`
-
-> **Baseline project:** Multus
-
-> **Gap description:** If Multus is selected as a CNI multiplexer an additional component have to be added and configured to support the centralized management of IP addresses. This is cruical to avoid the overlap of allocated IP address ranges.
-
-<a name="6.2.9"></a>
 ### 6.2.9 Lack of management API in Multus
 
 > **Related requirements:** `req.inf.ntw.03`
@@ -127,7 +118,7 @@ Note: with an underlying IaaS this is possible, but then it introduces (undesira
 > **Gap description:** If Multus is selected as a CNI multiplexer no network management API is provided. Multus does not provide a management API, therefore its API to consume the networks needs to contain administrative data, like CIDRs.
 
 
-<a name="6.2.9"></a>
+<a name="6.2.10"></a>
 ### 6.2.9 Missing IPVLAN CNI in case of Multus is used as a CNI multiplexer 
 
 > **Related requirements:** `infra.net.cfg.003`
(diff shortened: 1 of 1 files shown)


ASawwaf (approved):

rgstori:
In my opinion gaps should specify kubernetes as the baseline, as they are essential features of the platform regardless of the usage of a multiplexer

CsatariGergely:
@rgstori 

- #2121  defines the IPAM as an additional component and I agree with it. All what we need to do is to document that if the CNI Multiplexer is not DANM an additional component is needed to provide this functionality. I've created #2124 to cover this and removed the gap from Ch6. 
- The network API part is a bit more tricky as we already have a gap for Kubernetes in 6.2.5 and this gap is for the lack of the workaround in Multus. I'm open to a similar solution than to the IPAM issue, to add a description to Ch3 that in case of Multus a separate component is needed to provide the API and to add the component level details to Ch4, but here my problem is that I'm not aware of any open source solutions what we could name there. In case of DANM this is solved, therefore it is a gap in the Multus solution.
- The IPVLAN CNI case is similar to the network API one in a sense, that if we would have an open source, production ready IPVLAN CNI we could add it to Ch3 and Ch4, but I'm not aware of any solutions what would automatically provision the VLAN-s and configure the network fabric. In case of DANM this is also solved, therefore it is a gap in the Multus solution.

Sorry I could not explain these in the meeting yesterday due to my audio issues.

ulikleber:
Structure of this addition should be fixed. Maybe one title "gaps related to Multus" and then two parts, instead of two titles 6.2.9.

pgoyal01:
@CsatariGergely @tomkivlin There are many ways to skin a cat. IMO, we should list the CNIs that are required and those that may be optional depending upon the choices that the operator needs to make, for example, IPvlan, MACvlan, BGP, ..... 

No CNI multiplexer will meet all of the needs -- nor should they. A combination of solutions may provide the best possible solution to what an operator needs. For example, MULTUS and Calico (includes IPAM, BGP, etc.) can meet the necessary requirements including of enforcing security policy.

Again, we need a list of what CNIs are needed and not gaps in multiplexers. A CNI multiplexer's task is to support multiple CNIs - period. DANM may choose to implement other features but that doesn't make a gap for another multiplexer.

CsatariGergely:
> @CsatariGergely @tomkivlin There are many ways to skin a cat. IMO, we should list the CNIs that are required and those that may be optional depending upon the choices that the operator needs to make, for example, IPvlan, MACvlan, BGP, .....
> 
> No CNI multiplexer will meet all of the needs -- nor should they. A combination of solutions may provide the best possible solution to what an operator needs. For example, MULTUS and Calico (includes IPAM, BGP, etc.) can meet the necessary requirements including of enforcing security policy.
> 
> Again, we need a list of what CNIs are needed and not gaps in multiplexers. A CNI multiplexer's task is to support multiple CNIs - period. DANM may choose to implement other features but that doesn't make a gap for another multiplexer.

I agree, we should describe both the Multus and the DANM alternatives in Ch 4. on component level. 
What my struggle is that these gaps exist only in case of the Multus alternative, so they are not gaps when DANM is selected as a CNI multiplexer. Let me try a variant when these are generic and I describe that they are valid only in the Multus case.

CsatariGergely:
> 
> 
> Structure of this addition should be fixed. Maybe one title "gaps related to Multus" and then two parts, instead of two titles 6.2.9.

The two gaps should be in two separate chapter, I just had problems getting the chapter numbering right. Hopefully it is good now.

tomkivlin:
> I agree, we should describe both the Multus and the DANM alternatives in Ch 4. on component level.

I would agree, if we thought they are the only two options for providing multiple connection points to Pods.  But they're not, so we need to provide the specifications to allow appropriate solution components to be chosen.

tomkivlin:
This is no longer needed - will be covered in #2134 as per meeting decision on 17th December 2020: https://wiki.lfnetworking.org/display/LN/2020-12-17+-+%5BCNTT+-+RA2%5D+-+Meeting+Agenda+and+Minutes

PR #2114, opened 2020-12-01
[RM Ch9.1] Update to LCM Introduction

walterkozlowski:
Changed from VNF to workload, added comments about ITIL and added a Note below the table 1.

### DIFF
--- a/doc/ref_model/chapters/chapter09.md
+++ b/doc/ref_model/chapters/chapter09.md
@@ -12,46 +12,24 @@
 <a name="9.1"></a>
 ## 9.1 Introduction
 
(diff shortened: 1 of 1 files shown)


ASawwaf (approved):

pgoyal01:
@walterkozlowski I think we need to define the phases of LCM. For example, the phases may include:

- Capacity Build
    - Procure
    - Build and Configure (or simply configure)
- Operate and Manage
    - Operate (Customer management, accounting, etc.)
    - Monitor, Log and Alert (Faults, Performance)
    - Maintain (preventive and reactive) - including configuration changes

karinesevilla (approved):

walterkozlowski:
@pgoyal01, I am not really sure that we should distinguish between "Build" and "Operate" phases. For instance, in the spirit of CI/CD, Configuration is an activity for both (as it is actually defined in Sec 9.2. I would also assume that  Security Automation, Assurance Automation, Capacity Optimization are built in into the design, and hence apply to all phases, even before "Build" (Day -1).

sukhdevkapur (approved):

PR #2121, opened 2020-12-08
[RA2] Networking solution changes

### LINKED ISSUE #2119: [RA2 Ch2] Review NAT requirement(s)
There is currently only one requirement relating to NAT, which specifies that NAT must be supported ([infra.net.cfg.003](https://github.com/cntt-n/CNTT/blob/master/doc/ref_model/chapters/chapter05.md#523-virtual-networking)).

Do we need a requirement that states NAT-less connectivity is required?  If so, for which workloads?

Refer to previous version of network spec in RA2 (since hidden, as no requirement trace):
|`ra2.ntw.005`|NAT less connectivity|An IPVLAN CNI plugin or the [MACVLAN CNI](https://github.com/containernetworking/plugins/tree/master/plugins/main/macvlan) may be used||

What is the use-case for that requirement?  @CsatariGergely ?

tomkivlin:
As per multiple discussions on the topic, I propose a change whereby the architectural components / building blocks are described in chapter 3, and this is where I think the comparison table best sits.  Then the specifications in chapter 4 are cleaner and more discrete.

Resolves #2119


pgoyal01:
@tomkivlin Why have a comparison table? It is limited only to 2 out of many available solutions.

tomkivlin:
> @tomkivlin Why have a comparison table? It is limited only to 2 out of many available solutions.

I think it's worth showing the different approaches that can be taken.  Happy to add words to that effect.

rgstori (approved):
approved

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -209,70 +209,99 @@ infrastructure itself from the workloads compute resources.
...
-and the implicit requirement in [4.2.2 Virtual Network Interface
-Specifications](../../../ref_model/chapters/chapter04.md#422-virtual-network-interface-specifications)
-that documents the requirement to have the capability to attach several network
-interfaces to the pods, the CNTT architecture must support a CNI metaplugin/CNI
-multiplexer.
+Kubernetes networking is considered an "extension" to the core functionality,
+and is managed through the use of plugins [Network
+Plugins](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/).
+For hardware resources that are needed by Kubernetes applications, [Device
+Plugins](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/)
```
CsatariGergely, line 216: The link is not correct.

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -209,70 +209,99 @@ infrastructure itself from the workloads compute resources.
...
+by the Kubernetes applications. This allows resources such as "GPUs,
+high-performance NICs, FPGAs, InfiniBand adapters, and other similar computing
+resources that may require vendor specific initialization and setup" to be
+managed and consumed via standard interfaces.
+
+Figure 3-2 below shows the main building blocks of a Kubernetes networking solution:
+- **Kubernetes Control Plane**: this is the core of a Kubernetes cluster - the
+apiserver, etcd cluster, kube-scheduler and the various controller-managers. The
+control plane (in particular the apiserver) provide a centralised point by which
+the networking solution is managed using an abstract management API.
```
CsatariGergely, line 227: I do not see an abstract API in Kubernetes to manage networking solutions.

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -209,70 +209,99 @@ infrastructure itself from the workloads compute resources.
...
-A CNI metaplugin/CNI multiplexer has the capability to attach several
-interfaces, using different other CNI plugins, to a pod. Note that the different
-network characteristics of the interfaces might require different networking
-technologies, which would potentially require different CNI plugins.
+<p align="center"><img src="../figures/ch03_networking.png" alt="Kubernetes Networking Architecture" Title="Kubernetes Networking Architecture" width="100%"/></p>
+<p align="center"><b>Figure 3-2:</b> Kubernetes Networking Architecture</p>
 
-To comply with `req.inf.ntw.08`, inter node communication must be served by a
-CNI plugin which complies with the default [Kubernetes networking assumptions](https://kubernetes.io/docs/concepts/cluster-administration/networking/#the-kubernetes-network-model).
+<!--The above diagram is maintained here: https://wiki.lfnetworking.org/display/LN/CNTT+RA2+-+Kubernetes+-+Diagrams+-+Networking-->
```
CsatariGergely, line 267: "CNI Plugin Additional" should connect to a non eth0 interface.

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -209,70 +209,99 @@ infrastructure itself from the workloads compute resources.
...
+| Centrally administrated and configured (`req.inf.ntw.03`) | Supported via Kubernetes API | Supported via Kubernetes API |
+| Dual stack IPv4 and IPv6 for Kubernetes workloads (`req.inf.ntw.04`) | Supported via the additional CNI plugin | Supported |
+| Integrating SDN controllers (`req.inf.ntw.05`) | Supported via the additional CNI plugin | Supported via the additional CNI plugin |
+| More than one networking solution (`req.inf.ntw.06`) | Supported | Supported |
+| Choose whether or not to deploy more than one networking solution (`req.inf.ntw.07`) | Supported | Supported |
+| Kubernetes network model (`req.inf.ntw.08`) | Supported via the additional CNI plugin | Supported via the additional CNI plugin |
+| Do not interfere with or cause interference to any interface or network it does not own (`req.inf.ntw.09`) | Supported | Supported |
+| Cluster wide coordination of IP address assignment (`req.inf.ntw.10`) | Supported via the additional CNI plugin | Supported |
 
-The container based architecture must support telecom equipment networking where
```
CsatariGergely, line 248: I think we should keep this.

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -209,70 +209,99 @@ infrastructure itself from the workloads compute resources.
...
-abstract management APIs to manage the network connectivity of the CNF pods
-themselves.
+<p align="center"><b>Table 3-1:</b> Comparison of CNI multiplexers/meta-plugins within the networking solution</p>
 
-The API must support multiple tenants and must require elevated access rights to
-manipulate infrastructure related API objects as these operations generally
-require reconfiguration of the physical network infrastructure.
+(1): Under implementation in the current release.  
 
-To fulfill the requirements of `e.cap.016` the architecture should optionally
```
CsatariGergely, line 258: I think this should not be removed.

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -209,70 +209,99 @@ infrastructure itself from the workloads compute resources.
...
-support the use of device plugins via the Device Plugin API and the alignment of
-the devices, CPU topology and Huge Pages must be supported using the [Topology
-Manager](https://kubernetes.io/docs/tasks/administer-cluster/topology-manager/).
-
-The architecture must support IPv4, IPv6 and dual stack interfaces of the
```
CsatariGergely, line 263: I think this should not be removed.

CsatariGergely (changes requested):

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -209,70 +209,99 @@ infrastructure itself from the workloads compute resources.
...
+and advertisement of vendor hardware devices. In particular, devices such as
+FPGA, SR-IOV NICs, SmartNICs, etc. can be made available to Pods by using Device Plugins.
+Note that alignment of these devices, CPU topology and Huge Pages will need the use
+of the [Topology Manager](https://kubernetes.io/docs/tasks/administer-cluster/topology-manager/).
+- **External / Application Load Balancing**: As Kubernetes Ingress, Egress and Services have no
+support for all the protocols needed in telecommunication environments
+(Diameter, SIP, LDAP, etc) and their capacity is limited, the architecture includes
+the use of alternative load balancers, including external or ones built into the
+application. Management of external load balancers must be possible via Kubernetes API objects.
+- **Other Features**: these additional features that are required by the
```
CsatariGergely, line 254: I think we should list the "Other features" here and for the IPAM it should be noted, that the same IPAM should be responsible for the IP address assignment of all CNI plugins or other networking solutions.

CsatariGergely (changes requested):

pgoyal01:
@tomkivlin Does the chapter need the types of networking -- host, overlay, underlay , ...? Depending upon the type of networking chosen would determine the CNIs.

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: Thanks - updated in https://github.com/cntt-n/CNTT/pull/2121/commits/d15eeddaa160a783ccf6b5813d409dae7dcef582

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: ok maybe we are using the term differently then - for me, the Kubernetes APIs, with their extensibility through CRDs or AA are an abstract API to manage the networking solutions within Kubernetes (and beyond - e.g. for external load balancers).  What do you see an abstract API "doing"?

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: It does?  On my version it is connected to either `net0` or `net1`.

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: I'm not sure the requirement "requires" that.  

req.inf.ntw.10 states "The Architecture must support cluster wide coordination of IP address assignment." - I would argue that this doesn't necessarily mean a single IPAM component, as long as anything "doing" IPAM is aware of and communicates, somehow, with anything else doing IPAM - i.e. it is all co-ordinated.

tomkivlin:
> @tomkivlin Does the chapter need the types of networking -- host, overlay, underlay , ...? Depending upon the type of networking chosen would determine the CNIs.

Perhaps - I'm keen to avoid repeating Kubernetes documentation for the sake of it, so if we can link it back to the requirements then yes.

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: OK - but for which requirement?

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: Ah good spot - I meant to add this to chapter 4 instead.  Have added in https://github.com/cntt-n/CNTT/commit/21e0cf1f09c7a7ef73daf4a8768b38dd3705549f

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: Another mistake, sorry - have added to chapter 4 in https://github.com/cntt-n/CNTT/commit/719956b66256e0e10c06b64e5097c5a0dbab22bf

tomkivlin:
@fkautz I don't know enough about NSM to know whether the specs as presented in chapter04.md prevent it's use?  e.g. one that concerns me is `ra2.ntw.003` as I'm not sure whether NSM uses the CNI to integrate with the k8s control plane - my understanding is that it uses Device Plugin API and the kube-apiserver - is that right?

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -154,7 +156,7 @@ Architecture they must be implemented as per the following specifications:
...
 |`ra2.ntw.005`|SR-IOV Device Plugin for Network Intensive|When hosting workloads matching the Network Intensive profile, a Device Plugin for SR-IOV must be used.|[`e.cap.013`](chapter02.md#221-cloud-infrastructure-software-profile-capabilities)|
 |`ra2.ntw.006`|SR-IOV Network Plugin for Network Intensive|When hosting workloads matching the Network Intensive profile, a CNI-conformant Network Plugin for SR-IOV must be used.|[`e.cap.013`](chapter02.md#221-cloud-infrastructure-software-profile-capabilities)|
-|`ra2.ntw.007`|User plane networking|When hosting workloads matching the Network Intensive profile, a CNI network plugin that supports the use of OVS-DPDK or VPP vSwitches must be deployed as part of the networking solution.|[`infra.net.acc.cfg.001`](chapter02.md#223-cloud-infrastructure-software-profile-requirements)|
+|`ra2.ntw.007`|User plane overlay networking|When hosting workloads matching the Network Intensive profile, a CNI network plugin that supports the use of OVS-DPDK or VPP vSwitches should be deployed as part of the networking solution.|[`infra.net.acc.cfg.001`](chapter02.md#223-cloud-infrastructure-software-profile-requirements)|
```
tomkivlin, line 159: @rgs the original requirement states this is a must for DPDK:
image

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -157,7 +157,7 @@ Architecture they must be implemented as per the following specifications:
 |`ra2.ntw.005`|SR-IOV Device Plugin for Network Intensive|When hosting workloads matching the Network Intensive profile, a Device Plugin for SR-IOV must be used.|[`e.cap.013`](chapter02.md#221-cloud-infrastructure-software-profile-capabilities)|
 |`ra2.ntw.006`|SR-IOV Network Plugin for Network Intensive|When hosting workloads matching the Network Intensive profile, a CNI-conformant Network Plugin for SR-IOV must be used.|[`e.cap.013`](chapter02.md#221-cloud-infrastructure-software-profile-capabilities)|
 |`ra2.ntw.007`|User plane overlay networking|When hosting workloads matching the Network Intensive profile, a CNI network plugin that supports the use of OVS-DPDK or VPP vSwitches should be deployed as part of the networking solution.|[`infra.net.acc.cfg.001`](chapter02.md#223-cloud-infrastructure-software-profile-requirements)|
-|`ra2.ntw.008`||||
+|`ra2.ntw.008`|NATless connectivity|When hosting workloads that require source and destination IP addresses to be preserved in the traffic headers, a CNI plugin that exposes the pod IP directly to the external networks (e.g. MACVLAN or IPVLAN) is required.|*tbd*|
```
tomkivlin: Just wondering if we should add Calico to the list of e.g. plugins?  Given it is often a default.

pgoyal01: @tomkivlin BTW, as you know, if we use Calico we do not need a separate IPAM.

doc/ref_arch/kubernetes/chapters/chapter04.md
rgstori: i just noticed that note (3) under infra.net.acc.cfg.001 states:

> this feature is not applicable for Kubernetes based infrastructures due to lack of vSwitch however workloads need access to user space networking solutions.

I'm not sure about it, many CNIs (and NSM) have a dataplane component with a vswitch running.

Realistically, an user of a network intensive CNF would use either SRIOV or a vswitch-based CNI with some sort of acceleration (eg DPDK), but probably not both at the same time. we can reword this as a "must", and the SR-IOV one as a "may" or "should", or viceversa.

doc/ref_arch/kubernetes/chapters/chapter04.md
rgstori: yes that's fine for me

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: yeah makes sense - we should have an issue to update both the RM and the RA2 ch2 (this `infra.net.acc.cfg.001` is sourced in the RM chapter 5).

rgstori:
> @tomkivlin Does the chapter need the types of networking -- host, overlay, underlay , ...? Depending upon the type of networking chosen would determine the CNIs.

@pgoyal01 what about something like https://github.com/cntt-n/CNTT/commit/6ff56797bd67e03c8bd8f367ac406b9a2ecf39c0 ? Then we can categorise the types of traffic a CNF can have, and assign a CNI type to each one.

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: @CsatariGergely what are your thoughts?   I think either we need to change the requirement in chapter 2 to specify that a single IPAM component is required to manage cluster wide IP address assignment, or we accept that co-ordination between multiple IPAM components is acceptable.

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: Is this ok to resolve @CsatariGergely?

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: Is this ok to resolve @CsatariGergely?

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: Is this ok to resolve @CsatariGergely?

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: Is this ok to resolve @CsatariGergely?

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: OK, updated in https://github.com/cntt-n/CNTT/pull/2121/commits/9ff051c5f71eb6fe2c5366807a6fe23cda0d1670

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -209,70 +209,99 @@ infrastructure itself from the workloads compute resources.
...
-> Editors note: The possibility to SR-IOV for DPDK is under discussion.
-
-The integration of SDN solutions required by `req.inf.ntw.05` should be enabled
-via their respective CNI integration.
 
-> Note: An SDN solution can manage the pod networks via the Kubernetes API or
-can have a CNI integration.
+| Requirement | Networking Solution with Multus | Networking Solution with DANM |
+|---|---|---|
+| The overlay network encapsulation protocol needs to enable ECMP in the underlay (`infra.net.cfg.002`) | Supported via the additional CNI plugin | Supported via the additional CNI plugin |
```
pgoyal01, line 289: 1. \<broken record> I don't think this table is needed as we don't compare other CNI plugins for the same solution.
2. Tungsten Fabric should be included even though it is not a multiplexer but rather a default CNI that doesn't require a multiplexer to support multiple network connections -- @sukhdevkapur had stated that in a couple of meetings
3.  The support for DPDK and SR-IOV should be added as requirements to this Table: [infra.net.acc.cfg.001](https://github.com/cntt-n/CNTT/blob/tomkivlin-ra2-networking/doc/ref_arch/kubernetes/chapters/chapter02.md#223-cloud-infrastructure-software-profile-requirements)

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -209,70 +209,99 @@ infrastructure itself from the workloads compute resources.
...
+| NAT (`infra.net.cfg.003`) | Supported via the additional CNI plugin | Supported |
+| Security Groups (`infra.net.cfg.004`) | Not supported | Not supported <sub>(1)<sub> |
```
pgoyal01, line 291: Do we need Security Groups (an AWS term) or can the same be accomplished through Security Policies? May need to change RM and other places in RA2

doc/ref_arch/kubernetes/chapters/chapter04.md
pgoyal01: @rgstori You may have to word it so that they are exclusive -- both can be "must" but the exclusivity would ensure support for only one. Should this be a host (node) level?

doc/ref_arch/kubernetes/chapters/chapter03.md
CsatariGergely: For me an abstract API is independent from its implementation. I think it is the subject of the API operations where our misunderstanding is. We should have 3 API-s: 

1. An API to manage the LCM of different Kubernetes extensions including ones related to networking. 
2. An API to administer network related configurations. Tibycally for things done my network engineering, like IP range allocations.
3. An API to consume the networks. 
In Kubernetes 2) and 3) are collapsed into one API what enables the usage of implementation specific annotations, this is not an abstract API.

doc/ref_arch/kubernetes/chapters/chapter03.md
CsatariGergely: Correct, the 0-s confused me.

doc/ref_arch/kubernetes/chapters/chapter03.md
CsatariGergely: It is related to `req.gen.cnt.05`

doc/ref_arch/kubernetes/chapters/chapter03.md
CsatariGergely: @pgoyal01 
1. At the moment we define two options for CNI multiplexers DANM and Multus and as there are differences between them we need to describe these differences somewhere. I do not see why is this a returning problem. 
2. Please create a pr if you think that Tungsten Fabric should be specifically mentioned somewhere in the document.
3. Agreed

doc/ref_arch/kubernetes/chapters/chapter03.md
CsatariGergely: Yes, I think we mean Kubernetes Network Policies here.

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: I'm not sure that's the intention of that requirement and even if it is, the scope of "the Architecture" doesn't include other telecom equipment.  I think if we want to keep this statement in place, we need a corresponding requirement that asks for this.

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: Updated in https://github.com/cntt-n/CNTT/pull/2121/commits/0ef603d4c08dd9c0b3b86034353ebd9de31d826a.  @pgoyal01 ok to resolve?

doc/ref_arch/kubernetes/chapters/chapter03.md
pgoyal01: @CsatariGergely 
1. We should stop treating one or the other of our example multiplexers as the reference. Both Multus and DANM are CNI multiplexers.
2. Just because DANM has chosen to be more than a multiplexer doesn't mean that Multus as a multiplexer has gaps.
3. We have mentioned Multus and DANM as examples. Tomorrow if we add CNI-Genie as an example, would we create DANM gaps because CNI-Genie has incorporated some other functions.
4. We cannot add Tungsten Fabric (TF) as a multiplexer. When TF is used as the default CNI it does away with the need for a multiplexer.
5. In many CNTT documents we use example solutions but do not undertake to compare them.

doc/ref_arch/kubernetes/chapters/chapter03.md
pgoyal01: Yes -- it is resolved

doc/ref_arch/kubernetes/chapters/chapter03.md
pgoyal01: This will be covered by Issue #2134

doc/ref_arch/kubernetes/chapters/chapter04.md
rgstori: agreed to have a single "MUST" requirement, that has all 3 options: sriov, dpdk, vpp etc.

fkautz:
@tomkivlin The wording looks like it does actively block NSM from consideration. NSM is not a CNI solution and by extension is not implemented/accessed with a multiplexing CNI.

Would you like help rewording this?

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -148,37 +150,34 @@ Architecture they must be implemented as per the following specifications:
...
-
+|`ra2.ntw.001`|Centralised network administration|The networking solution deployed within the implementation must be administered through the Kubernetes API using native Kubernetes API resources and objects, or Custom Resources.|[`req.inf.ntw.03`](chapter02.md#223-cloud-infrastructure-software-profile-requirements)|
+|`ra2.ntw.002`|Container Network Interface|The networking solution deployed within the implementation must use a CNI-conformant Network Plugin for the Default Pod Network, as the alternative (kubenet) does not support cross-node networking or Network Policies.|[`req.inf.ntw.08`](chapter02.md#23-kubernetes-architecture-requirements)<br><br>[`infra.net.cfg.004`](chapter02.md#223-cloud-infrastructure-software-profile-requirements)|
+|`ra2.ntw.003`|Multiplexer/meta-plugin CNI Conformance|The networking solution deployed within the implementation must support the use a multiplexer/meta-plugin that conforms with the CNI specification.|[`req.inf.ntw.06`](chapter02.md#23-kubernetes-architecture-requirements)<br><br>[`req.inf.ntw.07`](chapter02.md#23-kubernetes-architecture-requirements)|
```
fkautz, line 154: This looks like it actively blocks NSM which is not implemented through a CNI approach.

We should consider rewording it.

doc/ref_arch/kubernetes/chapters/chapter02.md
```diff
@@ -306,8 +306,6 @@ The requirements in this section are to be delivered in addition to those in [se
 | `req.inf.ntw.08` | Infrastructure | Network | The Architecture **must**  provide a default network which implements the Kubernetes network model. | |
 | `req.inf.ntw.09` | Infrastructure | Network | The networking solution **must not** interfere with or cause interference to any interface or network it does not own. | |
 | `req.inf.ntw.10` | Infrastructure | Network | The Architecture **must** support cluster wide coordination of IP address assignment. | |
-| `req.inf.ntw.12` | Infrastructure | Network | The networking solution to support multiple connection points (e.cap.004 - please see [Section 2.2.1](#2.2.1)) must support any Kubernetes conformant CNI plugin to be incorporated as a networking solution. | |
```
CsatariGergely, line 308: What is the reason to remove this?

CsatariGergely (changes requested):

doc/ref_arch/kubernetes/chapters/chapter02.md
tomkivlin: I don't think this is correct (stops the use of NSM, for example) and even if we agree it is, I think it is a specification not an input requirement.

tomkivlin: I've added `ra2.ntw.007` in https://github.com/cntt-n/CNTT/pull/2121/commits/82ba2060adaadb5b54aa8c395c02b4d25b74c1ed to capture this specification.  My feeling is CNI is part of the spec, not an input requirement.

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: I have re-worded all the requirements in https://github.com/cntt-n/CNTT/pull/2121/commits/82ba2060adaadb5b54aa8c395c02b4d25b74c1ed.  Can you let me know if that fits better please?

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: I have added a TF based solution and NSM based solution to the table - will need some help to fill in some parts / check they're correct (@pgoyal01, @sukhdevkapur, @fkautz).

tomkivlin: Sorry @pgoyal01 I didn't see that issue before I did the latest commit.

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -209,71 +210,134 @@ infrastructure itself from the workloads compute resources.
...
+networks require NAT-less communication documented in `infra.net.cfg.003` and will
 need to be served by a CNI plugin with IPVLAN or MACVLAN support. On the other
 hand, the low latency, high throughput networks used for handling the user plane
 traffic require the capability to use a user space networking technology.
 
+| Requirement | Networking Solution with Multus | Networking Solution with DANM | Networking Solution with Tungsten Fabric | Networking Solution with NSM |
+|---|---|---|---|---|
+| The overlay network encapsulation protocol needs to enable ECMP in the underlay (`infra.net.cfg.002`) | Supported via the additional CNI plugin | Supported via the additional CNI plugin | Supported | TBC |
+| NAT (`infra.net.cfg.003`) | Supported via the additional CNI plugin | Supported | Supported | TBC |
+| Network Policies (Security Groups) (`infra.net.cfg.004`) | Not supported | Not supported <sub>(1)<sub> | Supported | Supported via the default CNI plugin |
```
CsatariGergely, line 319: We should add a note, to the default CNI pluign part that several CNI plugins do not support Network Policies.

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -209,71 +210,134 @@ infrastructure itself from the workloads compute resources.
...
+| SFC support (`infra.net.cfg.005`) | Not relevant | Not relevant | Not relevant | Not relevant |
+| Traffic patterns symmetry (`infra.net.cfg.006`) | Not relevant | Not relevant | Not relevant | Not relevant |
+| Network resiliency (`req.inf.ntw.01`) | Supported | Supported | Supported | Supported |
+| Centrally administrated and configured (`req.inf.ntw.03`) | Supported via Kubernetes API | Supported via Kubernetes API | Supported via Kubernetes API | Supported via Kubernetes API |
```
CsatariGergely, line 323: I think `req.inf.ntw.03` is about how the network resources (like IP address ranges) are configured and in this case Multus in itself does not support this. Others I guess support it via different CRD-s so I think it would be more precise to write "Suported via CRD-s".

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -148,38 +149,30 @@ Architecture they must be implemented as per the following specifications:
...
-
+|`ra2.ntw.001`|Centralised network administration|The networking solution deployed within the implementation must be administered through the Kubernetes API using native Kubernetes API resources and objects, or Custom Resources.|[`req.inf.ntw.03`](chapter02.md#223-cloud-infrastructure-software-profile-requirements)|
+|`ra2.ntw.002`|Default Pod Network - CNI|The networking solution deployed within the implementation must use a CNI-conformant Network Plugin for the Default Pod Network, as the alternative (kubenet) does not support cross-node networking or Network Policies.|[`req.inf.ntw.08`](chapter02.md#23-kubernetes-architecture-requirements)<br><br>[`infra.net.cfg.004`](chapter02.md#223-cloud-infrastructure-software-profile-requirements)|
+|`ra2.ntw.003`|Multiple connection points|The networking solution deployed within the implementation must support the capability to connect between ONE and FIVE connection points to each Pod, which are additional to the default connection point managed by the default Pod network CNI plugin.|[`e.cap.004`](chapter02.md#221-cloud-infrastructure-software-profile-capabilities)|
```
CsatariGergely, line 154: Is there a reason for maxing in 5 connection points?

CsatariGergely (changes requested):

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: The requirement in RM states up to 6 connection points per Pod, so this is default + 5.

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: yes good point - have clarified in https://github.com/cntt-n/CNTT/pull/2121/commits/f31fff76dfac38718ce808e2781d1e6074a16b80

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: OK. I think to avoid misunderstanding the requirement should be clarified.  For example, in a Multus-based solution, I would manage all components through the k8s API - I would deploy and configure Multus via k8s API, configure physical devices using DPAPI, etc.  I have updated in https://github.com/cntt-n/CNTT/pull/2121/commits/1e963f31762bb213339d58ff765acaf87e6bd896 to say "Kubernetes API Server" as it is an API hosted on that control plane service.  Is that ok?

fkautz: Would a CNI that doesn't use CRDs be considered non-compliant? Most CNIs don't use CRDs to determine their IP, subnet, or routes. These are often configured in a config file. 

Furthermore, in NSM, these addresses may sometimes be assigned by dynamically with parameters chosen by the network service endpoint such as in the case of a VPN.

doc/ref_arch/kubernetes/chapters/chapter04.md
fkautz: Can we change this to "at least 5" instead of artificially restricting the number of connection points?

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: @fkautz could you review the updated wording please?

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: @CsatariGergely are you able to resolve this please?

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: @fkautz good points.  This seems related to @CsatariGergely comments here: https://github.com/cntt-n/CNTT/pull/2121#discussion_r545129024.

Would you both agree to the following:
- Default cluster network (and any other extensions): can be managed by config files, k8s API server (e.g. CRDs), or a combination.
- Additional network management plane: can be managed by config files, k8s API server (e.g. CRDs), or a combination.
- Additional network connections to Pods: must be managed by k8s API server (e.g. CRDs, DPAPI, etc.)
- Configuration of the additional network connections to Pods: can be k8s API server (e.g. CRDs) or an external management plane (e.g. dynamic address assignment)

Does that sound feasible?

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: Let's consolidate this conversation here: https://github.com/cntt-n/CNTT/pull/2121#discussion_r552078730

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: I don't necessarily disagree, but I don't see why all IPAM in a cluster needs to be centralised.  Does anyone have a concrete reason why this requirement is needed?  I am happy to restrict the scope to the default cluster network (as I don't see the problem in something else managing the IPAM for additional networking needs).

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: @rgstori @pgoyal01 are either of you able to prepare a change to the PR to accommodate this?

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: Done in https://github.com/cntt-n/CNTT/commit/521f47178e3499b65cb9320b1d25dc6a9fe387b8

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: @CsatariGergely updated in https://github.com/cntt-n/CNTT/pull/2121/commits/0bca1dbf3c64d3164a057ac4beff6909596fb01e - are you ok to resolve the conversation?

doc/ref_arch/kubernetes/chapters/chapter04.md
rgstori: edited in https://github.com/cntt-n/CNTT/commit/5c28f9fd19328d3452e1a9618be181416f1b8cc1 . the requirement is ra2.ntw.010 now

rgstori: ra2.ntw.008 also mentions sr-iov for network intensive applications, have changed its verb to 'may'

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -209,71 +210,139 @@ infrastructure itself from the workloads compute resources.
...
+networks require NAT-less communication documented in `infra.net.cfg.003` and will
 need to be served by a CNI plugin with IPVLAN or MACVLAN support. On the other
 hand, the low latency, high throughput networks used for handling the user plane
 traffic require the capability to use a user space networking technology.
 
+| Requirement | Networking Solution with Multus | Networking Solution with DANM | Networking Solution with Tungsten Fabric | Networking Solution with NSM |
+|---|---|---|---|---|
+| The overlay network encapsulation protocol needs to enable ECMP in the underlay (`infra.net.cfg.002`) | Supported via the additional CNI plugin | Supported via the additional CNI plugin | Supported | TBC |
+| NAT (`infra.net.cfg.003`) | Supported via the additional CNI plugin | Supported | Supported | TBC |
+| Network Policies (Security Groups) (`infra.net.cfg.004`) | Not supported | Not supported <sub>(1)<sub> | Supported | Supported via a CNI Network Plugin that supports Network Policies |
```
pgoyal01, line 324: "Supported via a CNI Network Plugin that supports Network Policies" -- that would be the case for all solutions. For example, MULTUS and Calico, where Calico supports the policies.

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -148,38 +149,30 @@ Architecture they must be implemented as per the following specifications:
...
+|`ra2.ntw.005`|Multiplexer/meta-plugin|The networking solution deployed within the implementation may support the use a multiplexer/meta-plugin.|[`req.inf.ntw.06`](chapter02.md#23-kubernetes-architecture-requirements)<br><br>[`req.inf.ntw.07`](chapter02.md#23-kubernetes-architecture-requirements)|
+|`ra2.ntw.006`|Multiplexer/meta-plugin CNI Conformance|If used, the selected multiplexer/meta-plugin must integrate with the Kubernetes control plane via CNI.|[`req.inf.ntw.06`](chapter02.md#23-kubernetes-architecture-requirements)<br><br>[`req.inf.ntw.07`](chapter02.md#23-kubernetes-architecture-requirements)|
+|`ra2.ntw.007`|Multiplexer/meta-plugin CNI Plugins|If used, the selected multiplexer/meta-plugin must support the use of multiple CNI-conformant Network Plugins.|[`req.inf.ntw.06`](chapter02.md#23-kubernetes-architecture-requirements)|
+|`ra2.ntw.008`|SR-IOV Device Plugin for Network Intensive|When hosting workloads matching the Network Intensive profile, a Device Plugin for SR-IOV may be used to configure the SR-IOV devices and advertise them to the `kubelet`.|[`e.cap.013`](chapter02.md#221-cloud-infrastructure-software-profile-capabilities)|
```
tomkivlin, line 159: @rgstori rather than changing the verb here, can we add something to the statement.  Right now, SR-IOV could be used in a Kubernetes cluster, the DPAPI not used, but it would still pass a test for this requirement.  How about:
```suggestion
|`ra2.ntw.008`|SR-IOV Device Plugin for Network Intensive|When hosting workloads that match the Network Intensive profile and require SR-IOV acceleration, a Device Plugin for SR-IOV must be used to configure the SR-IOV devices and advertise them to the `kubelet`.|[`e.cap.013`](chapter02.md#221-cloud-infrastructure-software-profile-capabilities)|
```

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -209,71 +210,139 @@ infrastructure itself from the workloads compute resources.
...
+| SFC support (`infra.net.cfg.005`) | Not relevant | Not relevant | Not relevant | Not relevant |
```
pgoyal01, line 325: Why include, if "not relevant"? 
Is there anything in a solution that may be more beneficial or may be negatively impacting to SFCs

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -209,71 +210,139 @@ infrastructure itself from the workloads compute resources.
...
+| Traffic patterns symmetry (`infra.net.cfg.006`) | Not relevant | Not relevant | Not relevant | Not relevant |
```
pgoyal01, line 326: Why include, if "not relevant"?

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -209,71 +210,139 @@ infrastructure itself from the workloads compute resources.
...
+| Network resiliency (`req.inf.ntw.01`) | Supported | Supported | Supported | Supported |
```
pgoyal01, line 327: Is network resiliency supported by the solution or is supported by kubernetes

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -209,71 +210,139 @@ infrastructure itself from the workloads compute resources.
...
+| Dual stack IPv4 and IPv6 for Kubernetes workloads (`req.inf.ntw.04`) | Supported via the additional CNI plugin | Supported | Supported | Supported |
+| Integrating SDN controllers (`req.inf.ntw.05`) | Supported via the additional CNI plugin | Supported via the additional CNI plugin | TF is an SDN controller | TBC |
+| More than one networking solution (`req.inf.ntw.06`) | Supported | Supported | Supported | Supported |
+| Choose whether or not to deploy more than one networking solution (`req.inf.ntw.07`) | Supported | Supported | Supported | Supported |
+| Kubernetes network model (`req.inf.ntw.08`) | Supported via the additional CNI plugin | Supported via the additional CNI plugin | Supported | Supported via the default CNI plugin |
+| Do not interfere with or cause interference to any interface or network it does not own (`req.inf.ntw.09`) | Supported | Supported | Supported | Supported |
+| Cluster wide coordination of IP address assignment (`req.inf.ntw.10`) | Supported via the additional CNI plugin | Supported | Supported | Supported via IPAM CNI plugin |
```
pgoyal01, line 335: It should be same for Multus and Networking Solution with NSM

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -209,71 +210,139 @@ infrastructure itself from the workloads compute resources.
...
+
```
pgoyal01, line 336: Need an additional requirement that these solutions should be able to work with any CNI plugins including CNIs that implement functionality provided by that solution. For example, the use of an operator-preferred IPAM  CNI.

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -148,38 +149,30 @@ Architecture they must be implemented as per the following specifications:
...
+|`ra2.ntw.009`|Multiple connection points with multiplexer/meta-plugin|When a multiplexer/meta-plugin is used, the additional non-default connection points must be managed by a CNI-conformant Network Plugin.|[`req.inf.ntw.08`](chapter02.md#23-kubernetes-architecture-requirements)<br><br>[`infra.net.cfg.004`](chapter02.md#223-cloud-infrastructure-software-profile-requirements)|
+|`ra2.ntw.010`|User plane networking|When hosting workloads matching the Network Intensive profile, a CNI network plugin that supports the use of DPDK, VPP, or SR-IOV must be deployed as part of the networking solution.|[`infra.net.acc.cfg.001`](chapter02.md#223-cloud-infrastructure-software-profile-requirements)|
+|`ra2.ntw.011`|NATless connectivity|When hosting workloads that require source and destination IP addresses to be preserved in the traffic headers, a CNI plugin that exposes the pod IP directly to the external networks (e.g. Calico, MACVLAN or IPVLAN) is required.|[`req.inf.ntw.14`](chapter02.md#23-kubernetes-architecture-requirements)|
```
pgoyal01, line 162: Calico is a solution whereas MCVLAN and IPVLAN are technology choices

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -148,38 +149,30 @@ Architecture they must be implemented as per the following specifications:
...
+|`ra2.ntw.005`|Multiplexer/meta-plugin|The networking solution deployed within the implementation may support the use a multiplexer/meta-plugin.|[`req.inf.ntw.06`](chapter02.md#23-kubernetes-architecture-requirements)<br><br>[`req.inf.ntw.07`](chapter02.md#23-kubernetes-architecture-requirements)|
+|`ra2.ntw.006`|Multiplexer/meta-plugin CNI Conformance|If used, the selected multiplexer/meta-plugin must integrate with the Kubernetes control plane via CNI.|[`req.inf.ntw.06`](chapter02.md#23-kubernetes-architecture-requirements)<br><br>[`req.inf.ntw.07`](chapter02.md#23-kubernetes-architecture-requirements)|
+|`ra2.ntw.007`|Multiplexer/meta-plugin CNI Plugins|If used, the selected multiplexer/meta-plugin must support the use of multiple CNI-conformant Network Plugins.|[`req.inf.ntw.06`](chapter02.md#23-kubernetes-architecture-requirements)|
```
pgoyal01, line 158:
```suggestion
|`ra2.ntw.007`|Multiplexer/meta-plugin CNI Plugins|If used, the selected multiplexer/meta-plugin must support the use of any CNI-conformant Network Plugins.|[`req.inf.ntw.06`](chapter02.md#23-kubernetes-architecture-requirements)|
```

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -148,38 +149,30 @@ Architecture they must be implemented as per the following specifications:
...
+|`ra2.ntw.003`|Multiple connection points|The networking solution deployed within the implementation must support the capability to connect at least FIVE connection points to each Pod, which are additional to the default connection point managed by the default Pod network CNI plugin.|[`e.cap.004`](chapter02.md#221-cloud-infrastructure-software-profile-capabilities)|
+|`ra2.ntw.004`|Multiple connection points presentation|The networking solution deployed within the implementation must ensure that all additional non-default connection points are requested by Pods using standard Kubernetes resource scheduling mechanisms such as annotations or container resource requests and limits.|[`e.cap.013`](chapter02.md#221-cloud-infrastructure-software-profile-capabilities)|
+|`ra2.ntw.005`|Multiplexer/meta-plugin|The networking solution deployed within the implementation may support the use a multiplexer/meta-plugin.|[`req.inf.ntw.06`](chapter02.md#23-kubernetes-architecture-requirements)<br><br>[`req.inf.ntw.07`](chapter02.md#23-kubernetes-architecture-requirements)|
```
pgoyal01, line 156:
```suggestion
|`ra2.ntw.005`|Multiplexer/meta-plugin|The networking solution deployed within the implementation may utilise a CNI-conformant multiplexer/meta-plugin.|[`req.inf.ntw.06`](chapter02.md#23-kubernetes-architecture-requirements)<br><br>[`req.inf.ntw.07`](chapter02.md#23-kubernetes-architecture-requirements)|
```

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -148,38 +149,30 @@ Architecture they must be implemented as per the following specifications:
...
+|`ra2.ntw.008`|SR-IOV Device Plugin for Network Intensive|When hosting workloads matching the Network Intensive profile, a Device Plugin for SR-IOV may be used to configure the SR-IOV devices and advertise them to the `kubelet`.|[`e.cap.013`](chapter02.md#221-cloud-infrastructure-software-profile-capabilities)|
+|`ra2.ntw.009`|Multiple connection points with multiplexer/meta-plugin|When a multiplexer/meta-plugin is used, the additional non-default connection points must be managed by a CNI-conformant Network Plugin.|[`req.inf.ntw.08`](chapter02.md#23-kubernetes-architecture-requirements)<br><br>[`infra.net.cfg.004`](chapter02.md#223-cloud-infrastructure-software-profile-requirements)|
+|`ra2.ntw.010`|User plane networking|When hosting workloads matching the Network Intensive profile, a CNI network plugin that supports the use of DPDK, VPP, or SR-IOV must be deployed as part of the networking solution.|[`infra.net.acc.cfg.001`](chapter02.md#223-cloud-infrastructure-software-profile-requirements)|
```
pgoyal01, line 161:
```suggestion
|`ra2.ntw.010`|User plane networking|When hosting workloads matching the Network Intensive profile, CNI network plugins that support the use of DPDK, VPP, or SR-IOV must be deployed as part of the networking solution.|[`infra.net.acc.cfg.001`](chapter02.md#223-cloud-infrastructure-software-profile-requirements)|
```

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: What's the basis for this (i.e. a requirement in ch2, or other)?  I'm thinking it's essentially unachievable for any multiplexer to support all CNI plugins.  Surely it only needs to support the CNI plugins that are needed for the other requirements (e.g. IPAM, MACVLAN, SR-IOV, DPDK, etc.)?

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: Deliberately written as-was - next requirement (`ra2.ntw.006`) deals with the CNI conformance.

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: I'd argue they are well known enough as reference CNI plugins to warrant inclusion (e.g. https://www.cni.dev/plugins/main/macvlan/).

doc/ref_arch/kubernetes/chapters/chapter04.md
pgoyal01: It should be any CNI plugin and not support only, say, the multiplexer solution approved CNIs

doc/ref_arch/kubernetes/chapters/chapter04.md
pgoyal01: we can still simplify and just use "utilise"

doc/ref_arch/kubernetes/chapters/chapter04.md
pgoyal01: My issue is with listing Calico

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: Good point - corrected in https://github.com/cntt-n/CNTT/commit/1a9ee8b0d70479a8d95d510e48032b29f7c9048f

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: @fkautz @CsatariGergely are you ok with the wording I've added here: https://github.com/cntt-n/CNTT/commit/25235f3ffd99af431f0e94c8d8843d842bb7ab2b

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: Not my area of expertise, but I had always thought the SFC thing was managed at the orchestration layer, rather than within a Kubernetes cluster? I don't know of any "gotchas" regarding SFCs - i.e. if a CNF can run, can it be used in a chain? @rgstori what do you think?

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: Good point. Having looked back at the original requirement, I think it is relevant, but the answer to all should be "depends on CNI plugin used" - would you agree?

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: It's not a great requirement tbh "The Architecture must support network resiliency." - could mean anything. Resilient DC fabric? Multiple NICs to each worker node? What sort of resiliency, and to mitigate which risks?  I think unless we have better requirements, I think the answer needs to remain fairly high level.

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: Updated in https://github.com/cntt-n/CNTT/commit/62eef1a8aaf0743f1b9895ffb8472a34237f4f27

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: OK - if we need additional requirements they should be managed in chapter02 (not in this PR).

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: Why? They are all open source CNI plugins that can do what the description says.

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: I think it is more appropriate to list the CNI plugins that are needed. Then we can more clearly describe the gaps in ch6. Otherwise we'll need to document a gap for every possible CNI plugin?

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: Updated in https://github.com/cntt-n/CNTT/commit/e049b4d6aac1f9ca1e2a1bf58763903eee41fbaa

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: @CsatariGergely removed the term "abstracted" from this section in https://github.com/cntt-n/CNTT/commit/f16f1ca30506e50c2840cb7f4ed797982f93aaf5.  OK to resolve?

doc/ref_arch/kubernetes/chapters/chapter03.md
```diff
@@ -209,71 +210,158 @@ infrastructure itself from the workloads compute resources.
...
+| Integrating SDN controllers (`req.inf.ntw.05`) | Supported via the additional CNI plugin | Supported via the additional CNI plugin | TF is an SDN controller | TBC |
+| More than one networking solution (`req.inf.ntw.06`) | Supported | Supported | Supported | Supported |
+| Choose whether or not to deploy more than one networking solution (`req.inf.ntw.07`) | Supported | Supported | Supported | Supported |
+| Kubernetes network model (`req.inf.ntw.08`) | Supported via the additional CNI plugin | Supported via the additional CNI plugin | Supported | Supported via the default CNI plugin |
+| Do not interfere with or cause interference to any interface or network it does not own (`req.inf.ntw.09`) | Supported | Supported | Supported | Supported |
+| Cluster wide coordination of IP address assignment (`req.inf.ntw.10`) | Supported via IPAM CNI plugin | Supported | Supported | Supported via IPAM CNI plugin |
+
+<p align="center"><b>Table 3-1:</b> Comparison of example networking solutions</p>
+
+(1): Under implementation in the current release.
```
CsatariGergely, line 358: Add at least a note that the different networking solutions require different descriptors from the CNF-s, therefore the used networking solution should be agreed between the CNF vendors and the CNF operators.

doc/ref_arch/kubernetes/chapters/chapter04.md
CsatariGergely: How about `(e.g. Calico, MACVLAN or IPVLAN CNI plugins)`?

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -148,38 +149,30 @@ Architecture they must be implemented as per the following specifications:
...
+|`ra2.ntw.011`|NATless connectivity|When hosting workloads that require source and destination IP addresses to be preserved in the traffic headers, a CNI plugin that exposes the pod IP directly to the external networks (e.g. Calico, MACVLAN or IPVLAN) is required.|[`req.inf.ntw.14`](chapter02.md#23-kubernetes-architecture-requirements)|
+|`ra2.ntw.012`|Optional Device Plugins|When hosting workloads matching the Network Intensive profile that require the use of FPGA or other Acceleration Hardware, a Device Plugin for that FPGA or Acceleration Hardware may be used.|[`e.cap.016`](chapter02.md#221-cloud-infrastructure-software-profile-capabilities)|
+|`ra2.ntw.013`|Dual stack CNI|The networking solution deployed within the implementation must use a CNI-conformant Network Plugin that is able to support dual-stack IPv4/IPv6 networking.|[`req.inf.ntw.04`](chapter02.md#23-kubernetes-architecture-requirements)|
+|`ra2.ntw.012`||||
```
CsatariGergely, line 165: Delete the empty requirements.

CsatariGergely (changes requested):

doc/ref_arch/kubernetes/chapters/chapter03.md
pgoyal01: @tomkivlin Totally. Thanks

doc/ref_arch/kubernetes/chapters/chapter03.md
pgoyal01: @tomkivlin Should this requirement be dropped given that it raises more questions and the "supported" doesn't provide any insights.

doc/ref_arch/kubernetes/chapters/chapter04.md
pgoyal01: @CsatariGergely Yes that works. Thanks

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: @pgoyal01 I think dropping requirements at this late stage isn't ideal.  I'd rather leave it as-is and deal with the removal or improvement of the requirement and any subsequent wording in ch3/ch4 post-Elbrus.

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: OK- updated in https://github.com/cntt-n/CNTT/pull/2121/commits/1afa44b08815326f666da831151e0d7c9e46f957.

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: Done in https://github.com/cntt-n/CNTT/pull/2121/commits/9340d33ece604a3c39bb59dad4d63f18b6c2cb74

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: Agreed - have added something here: https://github.com/cntt-n/CNTT/pull/2121/commits/dba179249d9fa8611b34234e8d145ae55266af11

doc/ref_arch/kubernetes/chapters/chapter03.md
pgoyal01: @tomkivlin OK with a backlog issue

tomkivlin: #2187 created

pgoyal01: @tomkivlin Thanks for creating the issue. Can this entry be deleted from this table?

tomkivlin: Yes good idea, have done in https://github.com/cntt-n/CNTT/pull/2121/commits/fca7128df5da841af8a8366fe9be94b09c79a203

rgstori (approved):

doc/ref_arch/kubernetes/chapters/chapter03.md
rgstori: agree, SFC is more suited to 3GPP NSMF/NSSMF and ETSI MANO components. although the implementation of these _can be done e.g. with k8s operators_, it's not the point of CNTT RA2-compliant platform to "support SFC" (eg ONAP being a better place for modelling/maintaining/operating SFCs). I'm ok with removing the line entirely, or leaving it as is if it's needed as explicit memento of how SFC doesn't strictly belong here

rgstori (approved):

CsatariGergely (approved):

tomkivlin:
@pgoyal01 @fkautz @cntt-n/ref-a2 we need just one more approval on this monster please.  Many thanks for inputs so far!

doc/ref_arch/kubernetes/chapters/chapter03.md
tomkivlin: Thanks - I have removed the line in this table, and have created an issue to review/remove the requirement #2189

doc/ref_arch/kubernetes/chapters/chapter04.md
```diff
@@ -88,6 +88,7 @@ In order for the Kubernetes components to be conformant with the Reference Archi
...
+|`ra2.k8s.011`|Anuket profile labels|To clearly identify which worker nodes are compliant with the different profiles defined by Anuket the worker nodes must be labelled according to the following pattern: an `anuket.io/profile/basic` label must be set to `true` on the worker node if it can fullfill the requirements of the [basic profile](../../../ref_model/chapters/chapter04.md#4.2.5) and an `anuket.io/profile/network-intensive` label must be set to `true` on the worker node if it can fullfill the requirements of the [network intensive profile](../../../ref_model/chapters/chapter04.md#4.2.5).|||
```
pgoyal01, line 91: @tomkivlin Should the references be to Table 2-3 and 2-4 in Sections "[2.2.3 Cloud Infrastructure Software Profile Requirements](https://github.com/cntt-n/CNTT/blob/master/doc/ref_arch/kubernetes/chapters/chapter02.md#2.2.3)" and "[2.2.4 Cloud Infrastructure Hardware Profile Requirements](https://github.com/cntt-n/CNTT/blob/master/doc/ref_arch/kubernetes/chapters/chapter02.md#2.2.4)" as they have RA2 specific information?

pgoyal01 (approved):
Approved but there is one suggestion about a reference

doc/ref_arch/kubernetes/chapters/chapter04.md
tomkivlin: Good point - have updated in https://github.com/cntt-n/CNTT/pull/2121/commits/39433224ce604e354a24ef3afaacca6e8df114ae to point at RA2 Chapter 2.

Issue #2125, opened 2020-12-11
[RI2 Ch3] Update purpose of RI2 wrt. RC2


michaelspedersen:
RI2 currently states that RC2 is developed and tested using RI2, which is incorrect.

RI2, using Kuberef, should be verified by RC2, while also providing feedback to RC2 based on the results.
Details on the feedback process is still TBD

PR #2128, opened 2020-12-14
[RI2 Ch04] Add information about validation

rihabbanday:
Added few lines about RI2 validation. As mentioned in the issue #2125, the feedback loop between RI2 and RC2 is still TBD. More details on that will be added later.


michaelspedersen (approved):
Lgtm. As you mention, there's still a TODO to update the purpose between RI2 and RC2

gkunz (approved):

tomkivlin (approved):

Issue #2085, opened 2020-11-06
[RM Chxx] Primary & secondary networking


walterkozlowski:
General guidance and requirements on abstract level

wmk-admin:
Closed as PR#2155 has been approved and merged

PR #2155, opened 2020-12-21
[RM Ch04] Primary and secondary network textual motivation.

TFredberg:
Added verbal description for primary and secondary networks in the RM Networking model. 
Textual part of solution of Issue #2085.


walterkozlowski (approved):

petorre (approved):

doc/ref_model/chapters/chapter04.md
```diff
@@ -629,6 +629,8 @@ When having multiple Virtualization Layer instances on a shared hardware infrast
 
 The Overlay Networking separation is often done through encapsulation of Tenants traffic using overlay protocols e.g. through VxLAN or EVPN on the Underlay Networks e.g. based on L2 (VLAN) or L3 (IP) networks.
 
+Each Virtualization Infrastructure instances Overlay Network must support a basic primary communication channel in between the individual virtual compute instances in the tenants virtualized cluster. Due to the nature of Telecom applications handling of Networks and their related Network Functions they often need access to external non-translated traffic flows and have multiple separated or secondary traffic channels with abilities for different traffic treatments.
```
tomkivlin, line 632: Minor wording juggle to make it clearer (for me, anyway).  Also, should we be making this more agnostic than just virtualised? e.g.
```suggestion
The Overlay Network for each Cloud Infrastructure instance must support a basic primary communication channel between the individual compute instances in the tenant's cluster. Due to the way that Telecom applications handle Networks and their related Network Functions they often need access to external non-translated traffic flows and have multiple separated or secondary traffic channels with abilities for different traffic treatments.
```

tomkivlin (changes requested):

doc/ref_model/chapters/chapter04.md
TFredberg: Thanks @tomkivlin, I agree to all but one reformulation, where your suggested reformulation of:
"primary communication channel in between the individual virtual compute instances in the tenants virtualized cluster"
  into 
"primary communication channel between the individual compute instances in the tenant's cluster" 
  changes the scope I was trying to establish.
The intent was to express that it is the OpenStack or K8s instance that is responsible to provide a primary communication channel in between the tenants VMs or Containers withing their virtualization instance.
The intent was not to state that the OpenStack or K8s instance is responsible to provide the networking in between the compute instances i.e. the server hosts.

Do you think it works to express it as:
"primary communication channel between the individual virtual compute instances in the tenant's cluster" ?

tomkivlin: @TFredberg ah ok, understood.  How does the following work then to avoid over-use of "instance" and align with the [Glossary](https://github.com/cntt-n/CNTT/blob/master/doc/common/glossary.md#1.3)?

```suggestion
The Overlay Network for each Cloud Infrastructure deployment must support a basic primary communication channel between the compute Instances within each Tenant. Due to the nature of Telecom applications handling of Networks and their related Network Functions they often need access to external non-translated traffic flows and have multiple separated or secondary traffic channels with abilities for different traffic treatments.
```

TFredberg: @tomkivlin I don´t think we should use a term "Instance" here, as I agree it will be very overloaded, but rather the English word "instance" and ensure it always has a described context in the text e.g. "Virtualization Layer compute instance" or maybe shorter (and maybe better) "virtual compute instance" which both could represent a single Container or VM.

tomkivlin: @TFredberg that's what the term is defined as in the Glossary (https://github.com/cntt-n/CNTT/blob/master/doc/common/glossary.md#1.1) which is where terms should be defined, not in the RM text:

**Instance**: is a virtual compute resource, in a known state such as running or suspended, that can be used like a physical server.

>Note: Can be used to specify VM Instance or Container Instance.

TFredberg: @tomkivlin Sorry, I didn´t know about "Instance" that definition in the Glossary. 
Although in makes a lot of sense to have a collective term for a Container and a VM virtual compute instance, I don´t think it is good to use a single commonly used work of English as a Term for a local definition to a single layer in the stack. The resolution of our discussion is though likely not in a fast change of the Glossary :-) 

The Glossary now states: 
- Instance: is a virtual compute resource, in a known state such as running or suspended, that can be used like a physical server.
    - Note: Can be used to specify VM Instance or Container Instance.
- Tenant (Internal) Networks: Virtual networks that are internal to Tenant Instances.
The Glossary also have a number of terms that is explained using the English word "instances", so it is OK to use the word when not capitalized.

I don´t think "compute Instance" is correct since that should then carry the meaning: 
"compute virtual compute resource, in a known state"

One way would be to use the network term "Tenant Network" (not in plural) already defined in the Glossary as meaning a single" Virtual network internal to Tenant Instances" i.e. what is really meant here:
"The Overlay Network for each Cloud Infrastructure deployment must support a basic primary Tenant Network between the Instances within each Tenant."
Will this be fully comprehended by people that do not speak "the Glossary" as a native language?

tomkivlin: I think that works well, yes.  Also happy if we use "instances" instead of "Instances" to allow use of english word if you think that helps.

TFredberg: @tomkivlin I updated the text in this branch as per my last suggested first sentence and your other proposed changes.

tomkivlin (approved):

ulikleber (approved):

Issue #2195, opened 2021-01-15
[RM Ch4.2.1.1] Footnote on table 4-13 (Local Disk)


joaoarodrigues:
The disc is local to the VM/Pod-s point of view, otherwise this can be misunderstood

PR #2214, opened 2021-01-22
[RM Chapter04] Footnote added on Table 4-13

joaoarodrigues:
Highlight that Local Storage is inherit to VM/POD
and could be linked to an external storage

Closes: #2195


doc/ref_model/chapters/chapter04.md
```diff
@@ -331,7 +331,8 @@ The intent of the following Flavours list is to be comprehensive and yet effecti
 <p align="center"><b>Table 4-13:</b> Predefined Compute Flavours.</p>
 
 **1)** These compute Flavours are intended to be used for transitional purposes and workload vendors are expected to consume smaller Flavours and adopt microservices-based designs for their workloads.<br>
-**2)** In Kubernetes based environments these are the resource requests of the containers in the pods. To get guaranteed resources the resource requests should be set to the same values as the resource limits, to get burstable resources the resource limits should be higher than the resource requests while to get best effort resources none of resource requests of resource limits should be set.
+**2)** In Kubernetes based environments these are the resource requests of the containers in the pods. To get guaranteed resources the resource requests should be set to the same values as the resource limits, to get burstable resources the resource limits should be higher than the resource requests while to get best effort resources none of resource requests of resource limits should be set.<br>
+**3)** The disc is local to the VM/POD, which can be either local or external physical storage
```
pgoyal01, line 335: The sentence is a little confusing with the use of "local" twice. Also, typo - typical to use "disc" in Music while "disk"  in IT. Suggestion:
```suggestion
**3)** The disk is local to the VM/POD, which can be either installed on the local server or and external physical storage
```

doc/ref_model/chapters/chapter04.md
```diff
@@ -331,7 +331,8 @@ The intent of the following Flavours list is to be comprehensive and yet effecti
...
 
 **1)** These compute Flavours are intended to be used for transitional purposes and workload vendors are expected to consume smaller Flavours and adopt microservices-based designs for their workloads.<br>
-**2)** In Kubernetes based environments these are the resource requests of the containers in the pods. To get guaranteed resources the resource requests should be set to the same values as the resource limits, to get burstable resources the resource limits should be higher than the resource requests while to get best effort resources none of resource requests of resource limits should be set.
+**2)** In Kubernetes based environments these are the resource requests of the containers in the pods. To get guaranteed resources the resource requests should be set to the same values as the resource limits, to get burstable resources the resource limits should be higher than the resource requests while to get best effort resources none of resource requests of resource limits should be set.<br>
+**3)** The disk is local to the VM/POD, which can be either installed on the local server or and external physical storage
```
ulikleber, line 335: I don't understand how a disk can be on an external physical storage but local to the VM/POD. 
Do you mean with "local" storage that it can be accessed only by that VM/POD? Then some re-wording is needed.

tomkivlin: I too am slightly confused by the need for the clarification.  I was always under the impression that the "local disk" was the root disk/instance storage on which the OS is installed/containers are unpacked and is ephemeral.  So I guess could be from disks in the same physical server, or an external storage device (e.g. NFS/LUN from a SAN) mounted as the root disk.  Would the following make sense?
```suggestion
**3)** The "local disk" is an ephemeral disk that contains the operating system for a VM, or is used for unpacking and running Pod container images. This "local disk" can either be provided by storage devices housed within the physical server on which the VM/Pod is running, or from an external storage device such as SAN or NFS.
```

pgoyal01: @tomkivlin If you look at the table in RM [4.2.1.1 Predefined Compute Flavours](https://github.com/nokia/CNTT/blob/2195-rmch4-footnote/doc/ref_model/chapters/chapter04.md#4211-predefined-compute-flavours), you will see that the "local disk" sizes can be pretty large.  

In some clouds, the "root disk" is separate from the "local disk" -- it used to be the case in AWS where General Purpose Flavors came with "local disks" and the "root disk" wasn't specified (it was sufficient for the image, etc.). AWS has now changed its terminology.

As you know, Google cloud differentiates between local and persistent disks, where you can attach a maximum of 24 local SSD partitions each 375 GB in size for a total of 9 TB per instance.

To avoid confusion maybe we can instead adapt the term "Instance (VM/POD) Storage" that exists for the life of the instance. This can be created on disks on a server or attached externally.

tomkivlin: Agreed @pgoyal01 and the wording needs to reflect the purpose - I have forgotten but is the intention for the local disk in this table to be "instance storage and so same lifecycle as VM/Pod" or "local to the VM/Pod for performance reasons"?

pgoyal01: @tomkivlin I think was done to imitate AWS at the time. But we have the opportunity to rationalise the content. 

IMHO, it would be too much for general purpose computes to specify localised storage for performance. Thoughts?

Replace "local disk" in the Table and text with "Instance Storage for the life of the instance (VM/POD)"

tomkivlin: Yeah makes sense - updated my suggestion:
```suggestion
**3)** The "local disk" is an ephemeral disk that provides storage for the life of a VM/Pod, and can either be provided by storage devices housed within the physical server on which the VM/Pod is running, or from an external storage device such as SAN or NFS.
```

ulikleber: Fine for me ;-)

ulikleber (approved):

doc/ref_model/chapters/chapter04.md
walterkozlowski: Change not made according to @ulikleber suggestion, please do it

walterkozlowski (approved):
I am approving on the condition that @joaoarodrigues resolves conversation and suggestions from @tomkivlin regarding point 3) and also resolves the conflicting files

wmk-admin:
@joaoarodrigues, please resolve the conflict, and I will merge the change.

joaoarodrigues:
@wmk-admin Please check if it's ok now.

PR #2245, opened 2021-01-28
Add CNTT RC1 Elbrus release in History

collivier:


tomkivlin (approved):

walterkozlowski (approved):

pgoyal01 (commented):
RA1 4.0 added Edge and changes to certain properties necessitated by Edge. Does RC1 have tests associated with those changes?

collivier:
@pgoyal01 I interpret the bundle version as it conforms to RA1 in the current testing scope. RC1 doesn't fully cover all requirement from RA1. As far as I understand, there is no API changes in Chapter 5 for Edge which is driving the current testing. There is now possible mismatches between Chapter 4 and 5 (4 could ask for features in 5 as optional or the opposite).
True that Edge could ask for a new set of mandatory API (RA1) and then new test cases (RC1) based on the same test framework (basically Functest tunning as tempest_neutron and tempest_neutron_cntt).

karinesevilla (approved):

PR #2247, opened 2021-01-28
Update latest release to CNTT Elbrus

collivier:
Bundle: 5

### DIFF
--- a/doc/ref_arch/kubernetes/README.md
+++ b/doc/ref_arch/kubernetes/README.md
@@ -5,11 +5,11 @@
 This is Kubernetes based Reference Architecture (RA-2)
 
 ## Release Information
-**Bundle: _4_**
+**Bundle: _5_**
 
 **Version: _0_**
 
-**Release Date: _25th September 2020_**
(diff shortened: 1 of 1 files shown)


walterkozlowski (approved):

pgoyal01 (approved):

pgoyal01:
@collivier @tomkivlin RA2 as version 5.0 implies that it is aligned with RM 5.0 - is that the case? If not this could be a 4.x where x is not '0'

tomkivlin:
> @collivier @tomkivlin RA2 as version 5.0 implies that it is aligned with RM 5.0 - is that the case? If not this could be a 4.x where x is not '0'

I'm not aware of any misalignment - certainly nothing in the RA2 backlog.  I thought the intention was for RM/RA/RC to align each time with RI lagging.

ulikleber (approved):

collivier:
@tomkivlin @pgoyal01 I'm fine to update this PR and the one about RC2 to 4.1

PR #2249, opened 2021-02-01
Update Ansible community general to 2.0

collivier:
https://github.com/ansible-collections/community.general/blob/stable-2/CHANGELOG.rst

### DIFF
--- a/doc/ref_cert/RC1/chapters/chapter04.md
+++ b/doc/ref_cert/RC1/chapters/chapter04.md
@@ -206,7 +206,7 @@ virtualenv functest
 . functest/bin/activate
 pip install ansible
 ansible-galaxy install collivier.xtesting
-ansible-galaxy collection install ansible.posix community.general community.grafana community.kubernetes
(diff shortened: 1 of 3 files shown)


tomkivlin (approved):
No problem with the change, but the title doesn't match the changes made? Or, I'm not sure how adding the `community.docker` and `community.postgresql` collections is updating `community.general` to 2.0?

collivier:
@tomkivlin 
It is but I may create one shortcut or two:
See the counter part: https://github.com/collivier/ansible-role-xtesting/commit/b556155567b3517dc1ba8f46fd1ad6f8dc473b33

+ the following extract from the changelog printed in git commit msg:
If you use Ansible 2.9 and the docker plugins or modules from this collections, community.general 2.0.0 results in errors when trying to use the docker content by FQCN, like community.general.docker_container. Since Ansible 2.9 is not able to use redirections, you will have to adjust your playbooks and roles manually to use the new FQCNs (community.docker.docker_container for the previous example) and to make sure that you have community.docker installed.

rgstori (approved):

Issue #2376, opened 2021-04-27
[RM Ch03] Storage Telemetry: Capacity


acmacm:
Table 3-3 gives attributes of virtual storage that can be requested to support a given workload.

For the Performance attribute: how is "capacity" different from the attribute: size ??

IOW, it helps to have definitions of the attributes available to the readers.  See https://github.com/cntt-n/CNTT/issues/2375

Issue #2375, opened 2021-04-27
[RM Ch03] Storage Telemetry: Disk Operation


acmacm:
Table 3-3 gives attributes of virtual storage that can be requested to support a given workload.
When required to assure that the performance attributes are being delivered, then some of these can be measured passively with [Disk Operation metrics](https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.2/html/metrics_store_user_guide/chap-metrics#Disks):

- Latency (R/W): The average amount of time it took to do a read/write operation, in milliseconds.  
- IOPS: The time spent doing I/Os in milliseconds.

Issue #2374, opened 2021-04-27
[RM Ch03] Storage Telemetry: Utilization in Bytes


acmacm:
Table 3-3 gives attributes of virtual storage that can be requested to support a given workload.
When [monitoring the virtual storage](https://access.redhat.com/documentation/en-us/red_hat_virtualization/4.2/html/metrics_store_user_guide/chap-metrics#DF ) is needed, then the amount of free, used, and reserved disk space, in bytes, on the assigned file system is a useful set of metrics.

PR #2377, opened 2021-04-28
[RM Ch03] Storage Telemetry Definitions

acmacm:
See Issues https://github.com/cntt-n/CNTT/issues/2374 https://github.com/cntt-n/CNTT/issues/2375 and https://github.com/cntt-n/CNTT/issues/2376

### DIFF
--- a/doc/ref_model/chapters/chapter03.md
+++ b/doc/ref_model/chapters/chapter03.md
@@ -130,7 +130,9 @@ Persistent storage outlives the compute instance whereas ephemeral storage is li
 There are multiple storage performance attributes, such as latency, IO/s (Input/Output Operations per second, a.k.a. IOPS), and throughput. For example, a workload may require one of its storage devices to provide low latency, high IO/s and very large/huge storage size (terabytes of data).
 Low Latency storage is for workloads which have strong constraints on the time to access the storage.
 High IO/s oriented storage is for workloads requiring lots of read/write actions.
-Large size storage is for workloads that need lots of volume without strong perfomance constraints.
+Large size storage is for workloads that need lots of volume without strong performance constraints.
+Note that approximate numeric ranges for the qualitative values used above are given in the 
+"["Storage Extensions"]"(https://cntt.readthedocs.io/en/latest/ref_model/chapters/chapter04.html#4.2.3) section.
 
 Storage resources have the following attributes, with metric definitions that support verification through passive measurements (telemetry) where appropriate:
 
--- a/doc/ref_model/chapters/chapter03.md
+++ b/doc/ref_model/chapters/chapter03.md
@@ -138,8 +138,8 @@ Storage resources have the following attributes, with metric definitions that su
 |---------------------|------------------------------------------------------------------------------------------------|
 | `name`              | name of storage resources                                                                      |
 | `data availibilty`  | persistent or ephemeral                                                                        |
-|                     | Read and Write Latency, The average amount of time to perform a R/W operation, in milliseconds |
-| `performance`       | Read and Write IO/s, The average rate of performing R/W in IO operations per second            |
+| `performance`       | Read and Write Latency, The average amount of time to perform a R/W operation, in milliseconds |
+|                     | Read and Write IO/s, The average rate of performing R/W in IO operations per second            |
 |                     | Read and Write Throughput, The average rate of performing R/W operations in Bytes per second   |
 | `enhanced features` | replication, encryption                                                                        |
 | `type`              | block, object or file                                                                          |
--- a/doc/ref_model/chapters/chapter03.md
+++ b/doc/ref_model/chapters/chapter03.md
@@ -124,24 +124,26 @@ _**Example**: a virtual compute descriptor as defined in TOSCA Simple Profile fo
 <a name="3.2.1.3"></a>
 #### 3.2.1.3 Virtual Storage
 
-A workload can request different types of storage based on data longevity: persistent or ephemeral storage.
+A workload can request different sizes and types of storage based on data longevity: persistent or ephemeral storage.
 Persistent storage outlives the compute instance whereas ephemeral storage is linked to compute instance lifecycle.
 
-There are multiple storage performance requirements such as latency, IOPS and capacity. For example, a workload may require one of its storage device to provide low latency, high IOPS and very large/huge storage capacity (terabytes of data).
+There are multiple storage performance attributes, such as latency, IO/s (Input/Output Operations per second, a.k.a. IOPS), and throughput. For example, a workload may require one of its storage devices to provide low latency, high IO/s and very large/huge storage size (terabytes of data).
 Low Latency storage is for workloads which have strong constraints on the time to access the storage.
-High IOPS oriented storage is for workloads requiring lots of read/write actions.
-Capacity oriented storage is for workloads that need lots of volumetry without strong perfomance constraints.
+High IO/s oriented storage is for workloads requiring lots of read/write actions.
+Large size storage is for workloads that need lots of volume without strong perfomance constraints.
 
-Storage resources have the following attributes:
+Storage resources have the following attributes, with metric definitions that support verification through passive measurements (telemetry) where appropriate:
 
-| Attribute           | Description                                                              |
-|---------------------|--------------------------------------------------------------------------|
-| `name`              | name of storage resources                                                |
-| `data availibilty`  | persistent or ephemeral                                                  |  
-| `performance`       | latency, IOPS, capacity                            |
-| `enhanced features` | replication, encryption                                                  |
-| `type`              | block, object or file                                                    |
-| `size`              | size in GB                                                       |
+| Attribute           | Description                                                                                    |
+|---------------------|------------------------------------------------------------------------------------------------|
+| `name`              | name of storage resources                                                                      |
+| `data availibilty`  | persistent or ephemeral                                                                        |
+|                     | Read and Write Latency, The average amount of time to perform a R/W operation, in milliseconds |
+| `performance`       | Read and Write IO/s, The average rate of performing R/W in IO operations per second            |
+|                     | Read and Write Throughput, The average rate of performing R/W operations in Bytes per second   |
+| `enhanced features` | replication, encryption                                                                        |
+| `type`              | block, object or file                                                                          |
+| `size`              | size in GB, telemetery includes the amount of free, used, and reserved disk space, in bytes    |
 
 <p align="center"><b>Table 3-3:</b> Attributes of storage resources</p>
 
--- a/doc/ref_model/chapters/chapter03.md
+++ b/doc/ref_model/chapters/chapter03.md
@@ -132,7 +132,7 @@ Low Latency storage is for workloads which have strong constraints on the time t
 High IO/s oriented storage is for workloads requiring lots of read/write actions.
 Large size storage is for workloads that need lots of volume without strong performance constraints.
 Note that approximate numeric ranges for the qualitative values used above are given in the 
-"["Storage Extensions"]"(https://cntt.readthedocs.io/en/latest/ref_model/chapters/chapter04.html#4.2.3) section.
+"["Storage Extensions"]"(chapter04.html#4.2.3) section.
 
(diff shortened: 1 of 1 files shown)


doc/ref_model/chapters/chapter03.md
```diff
@@ -124,24 +124,26 @@ _**Example**: a virtual compute descriptor as defined in TOSCA Simple Profile fo
 <a name="3.2.1.3"></a>
 #### 3.2.1.3 Virtual Storage
 
-A workload can request different types of storage based on data longevity: persistent or ephemeral storage.
+A workload can request different sizes and types of storage based on data longevity: persistent or ephemeral storage.
```
pgoyal01, line 127: typo: 'size' instead of 'sizes'

Is the 'size' determined by 'data longevity'?

```suggestion
A workload can request storage based on data longevity (persistent or ephemeral storage), different types of storage (HDD, SSD, etc.) and storage size.
```

doc/ref_model/chapters/chapter03.md
```diff
@@ -124,24 +124,26 @@ _**Example**: a virtual compute descriptor as defined in TOSCA Simple Profile fo
...
 Persistent storage outlives the compute instance whereas ephemeral storage is linked to compute instance lifecycle.
 
-There are multiple storage performance requirements such as latency, IOPS and capacity. For example, a workload may require one of its storage device to provide low latency, high IOPS and very large/huge storage capacity (terabytes of data).
+There are multiple storage performance attributes, such as latency, IO/s (Input/Output Operations per second, a.k.a. IOPS), and throughput. For example, a workload may require one of its storage devices to provide low latency, high IO/s and very large/huge storage size (terabytes of data).
```
pgoyal01, line 130: are we defining the terms: **high** IO/s, **very large/huge** storage size

doc/ref_model/chapters/chapter03.md
```diff
@@ -124,24 +124,26 @@ _**Example**: a virtual compute descriptor as defined in TOSCA Simple Profile fo
...
-
-| Attribute           | Description                                                              |
-|---------------------|--------------------------------------------------------------------------|
-| `name`              | name of storage resources                                                |
-| `data availibilty`  | persistent or ephemeral                                                  |  
-| `performance`       | latency, IOPS, capacity                            |
-| `enhanced features` | replication, encryption                                                  |
-| `type`              | block, object or file                                                    |
-| `size`              | size in GB                                                       |
+High IO/s oriented storage is for workloads requiring lots of read/write actions.
```
pgoyal01, line 132: instead of "lots" can we have a quantitative value as in 4.2.3

doc/ref_model/chapters/chapter03.md
```diff
@@ -124,24 +124,26 @@ _**Example**: a virtual compute descriptor as defined in TOSCA Simple Profile fo
...
+Large size storage is for workloads that need lots of volume without strong perfomance constraints.
```
pgoyal01, line 133: Can we specify a value or range for "Large"

doc/ref_model/chapters/chapter03.md
```diff
@@ -124,24 +124,26 @@ _**Example**: a virtual compute descriptor as defined in TOSCA Simple Profile fo
...
+
+Storage resources have the following attributes, with metric definitions that support verification through passive measurements (telemetry) where appropriate:
+
+| Attribute           | Description                                                                                    |
+|---------------------|------------------------------------------------------------------------------------------------|
+| `name`              | name of storage resources                                                                      |
+| `data availibilty`  | persistent or ephemeral                                                                        |
+|                     | Read and Write Latency, The average amount of time to perform a R/W operation, in milliseconds |
+| `performance`       | Read and Write IO/s, The average rate of performing R/W in IO operations per second            |
```
pgoyal01, line 142: Can you move 'performance' to the row above as in the rendered table it appears Latency is part of 'data availability'

or use html tables

doc/ref_model/chapters/chapter03.md
acmacm: LGTM, thanks

doc/ref_model/chapters/chapter03.md
acmacm: It appears that the RM provides quantitative definitions for some of these performance attributes in https://cntt.readthedocs.io/en/latest/ref_model/chapters/chapter04.html#4.2.3  (as you pointed out today). The sentence that begins "For example, ..." is only modified to use consistent terms.

doc/ref_model/chapters/chapter03.md
acmacm: I suggest we reference https://cntt.readthedocs.io/en/latest/ref_model/chapters/chapter04.html#4.2.3 for the range of values. I don't know the syntax to do a forward reference though...

doc/ref_model/chapters/chapter03.md
acmacm: I suggest we reference https://cntt.readthedocs.io/en/latest/ref_model/chapters/chapter04.html#4.2.3 for the range of values. I don't know the syntax to do a forward reference though...

doc/ref_model/chapters/chapter03.md
acmacm, line 142:
```suggestion
| `performance`       | Read and Write Latency, The average amount of time to perform a R/W operation, in milliseconds |
|                     | Read and Write IO/s, The average rate of performing R/W in IO operations per second            |
```

acmacm: I think I proposed the change below...

acmacm: Done below

doc/ref_model/chapters/chapter03.md
pgoyal01: @acm If I were to use the url and attach it to the word 'reference" I would use the following syntax with the quotes:
       "["reference"]"(https://cntt.readthedocs.io/en/latest/ref_model/chapters/chapter04.html#4.2.3)

From Chapter 2 a reference to chapter 4 the relative reference would be "./chapter04.html#4.2.3"

acmacm: Thanks, I'll try a summary sentence to resolve your comments on qualitative terms. 
The last sentence of the paragraph is below, so I'll add it there.

doc/ref_model/chapters/chapter03.md
acmacm:
```suggestion
Large size storage is for workloads that need lots of volume without strong performance constraints.
Note that approximate numeric ranges for the qualitative values used above are given in the 
"["Storage Extensions"]"(https://cntt.readthedocs.io/en/latest/ref_model/chapters/chapter04.html#4.2.3) section.
```

doc/ref_model/chapters/chapter03.md
```diff
@@ -124,24 +124,28 @@ _**Example**: a virtual compute descriptor as defined in TOSCA Simple Profile fo
...
-| `name`              | name of storage resources                                                |
-| `data availibilty`  | persistent or ephemeral                                                  |  
-| `performance`       | latency, IOPS, capacity                            |
-| `enhanced features` | replication, encryption                                                  |
-| `type`              | block, object or file                                                    |
-| `size`              | size in GB                                                       |
+High IO/s oriented storage is for workloads requiring lots of read/write actions.
+Large size storage is for workloads that need lots of volume without strong performance constraints.
+Note that approximate numeric ranges for the qualitative values used above are given in the 
+"["Storage Extensions"]"(https://cntt.readthedocs.io/en/latest/ref_model/chapters/chapter04.html#4.2.3) section.
```
CsatariGergely, line 135:
```suggestion
"["Storage Extensions"]"(chapter04.html#4.2.3) section.
```
Please use relative links to keep the links valid in multiple release branches.

CsatariGergely (changes requested):

doc/ref_model/chapters/chapter03.md
acmacm: Thanks Greg0, better to reference within than external!

doc/ref_model/chapters/chapter03.md
```diff
@@ -124,24 +124,28 @@ _**Example**: a virtual compute descriptor as defined in TOSCA Simple Profile fo
...
-| `name`              | name of storage resources                                                |
-| `data availibilty`  | persistent or ephemeral                                                  |  
-| `performance`       | latency, IOPS, capacity                            |
-| `enhanced features` | replication, encryption                                                  |
-| `type`              | block, object or file                                                    |
-| `size`              | size in GB                                                       |
+High IO/s oriented storage is for workloads requiring lots of read/write actions.
+Large size storage is for workloads that need lots of volume without strong performance constraints.
+Note that approximate numeric ranges for the qualitative values used above are given in the 
+"["Storage Extensions"]"(chapter04.html#4.2.3) section.
```
pgoyal01, line 135:
```suggestion
[Storage Extensions](./chapter04.html#4.2.3) section.
```

acmacm: Thanks, WFM

petorre:
This is great improvement and clarification of existing text.
Only "IOPS" should not be changed to "IO/s". Search engines like [https://duckduckgo.com/?q=**input+output+operations+per+second**&ia=web](https://duckduckgo.com/?q=input+output+operations+per+second&ia=web) all show "IOPS", same with storage vendor webs sites.

markbeierl:
> This is great improvement and clarification of existing text.
> Only "IOPS" should not be changed to "IO/s". Search engines like https://duckduckgo.com/?q=input+output+operations+per+second&ia=web all show "IOPS", same with storage vendor webs sites.

Absolutely correct.  IOPS is very much an industry standard term and must be used in this context, thanks.

doc/ref_model/chapters/chapter03.md
```diff
@@ -124,24 +124,28 @@ _**Example**: a virtual compute descriptor as defined in TOSCA Simple Profile fo
 <a name="3.2.1.3"></a>
 #### 3.2.1.3 Virtual Storage
 
-A workload can request different types of storage based on data longevity: persistent or ephemeral storage.
+A workload can request storage based on data availability (persistent or ephemeral storage), different types of storage (HDD, SSD, etc.) and storage size.
```
markbeierl, line 127: Availability is more about the resource being possible.  I think durability could be a better word for length of storage persistance.

markbeierl: Sorry, also the types of storage are not typically requested by workload.  Given something like openstack, you would request the storage from a specific pool by name.  While the pool name might be HDD, the actual workload can never know that is really is HDD and is definitely not allowed to know that level of detail about the underlying infrastructure.

acmacm:
> > This is great improvement and clarification of existing text.
> > Only "IOPS" should not be changed to "IO/s". Search engines like https://duckduckgo.com/?q=input+output+operations+per+second&ia=web all show "IOPS", same with storage vendor webs sites.
> 
> Absolutely correct. IOPS is very much an industry standard term and must be used in this context, thanks.

I agree, although I was asked to consider  table 4-15 in section 4.2.3 for consistency, where they used the term IO/s.
I'll revert to IOPS in a new commit.

doc/ref_model/chapters/chapter03.md
acmacm: I agree we could replace **availability** with another term, and I think "retaining policy" may be approximately what we want to say.  IOW, data in ephemeral storage could last a very long time (such as the life of a VM), while data in persistent storage could be overwritten frequently - if its main purpose  is to recover state following a crash or other event. So, associating time with persistent and ephemeral might be right in many circumstances, but not all. 

Also, we already use *type* for  block, object or file storage.  
So, we won't mention HDD or SDD as attributes that workload can request.

CsatariGergely (approved):

petorre (approved):

markbeierl (approved):

walterkozlowski (approved):
Good effort to improve the contents

acmacm:
Thanks Walter, and everyone who commented to improve this PR!

PR #2413, opened 2021-05-11
[RI2 Ch03] Update Lab Requirements

### LINKED ISSUE #2208: [RI2 ch04] Add physical server specs
e.g. number/speed of CPU, number of SR-IOV capable NICs, RAM, Storage, etc.

Needed to conform with `ra2.ch.various`.

### LINKED ISSUE #2397: [RI2 Ch03] Update Lab requirements
The infrastructure requirements in Ch 03 seem to be outdated and needs to be updated

rihabbanday:
Fixes #2397, Fixes #2208


doc/ref_impl/cntt-ri2/chapters/chapter03.md
```diff
@@ -72,22 +72,23 @@ The network fabric shall comprise redundant layer 2 connectivity between hosts.
...
+1. Five (5) physical or virtual servers, serving as either master (controller) or compute/worker nodes
 
 The jump / test host allows to install and manage the operation of the RI2. For example, the host can be used to execute the installer software to deploy and configure the RI2 on the other physical or virtual servers. Once installed, the host may be used to run test tools or it may be used to generate test traffic intended for performance measurements. To support the latter, the server provided for the jump / test host shall meet similar hardware requirements as the master (controller) / compute hosts. If the jump host is used only for deployment or execution of functional tests after deployment, a VM with the resources as outlined below is sufficient.
 
 Each server shall meet the following minimum specifications:
 
 * CPU
-  *  2x x86_64 CPU sockets, providing 20 cores each, 40 simultaneous multi-threads (SMT), at 2.2 GHz
+  *  2x x86_64 CPU sockets, providing 20 cores each, 40 simultaneous multi-threads (SMT), at 2.2 GHz. If using VM's, the servers must be equipped with at least 16 vCPUs.
```
pgoyal01, line 82: I realise that these are issues that exist and not new but should we take this opportunity to clean-up? Thanks
- Are we asking for 2-sockets? But how many sockets populated with CPUs? 
- Is the 20 cores "each" for the server or each of the CPUs?
- the 40 SMT threads is as a result of the SMT enablement -- so does that and the 16 vCPUs part of server specifications or server configurations?

doc/ref_impl/cntt-ri2/chapters/chapter03.md
```diff
@@ -72,22 +72,23 @@ The network fabric shall comprise redundant layer 2 connectivity between hosts.
...
-  * 256 GB RAM
+  * 32 GB RAM
 * Storage
-  * 1.0 TB SSD via SATA 6 Gbps
+  * 320 GB
 * Network Interfaces
   * 4x 10 Gbps Ethernet Ports
   * Out-of-band Management Port (physical server only)
 
+Note that the minimum specifications listed above map to the requirements for the Basic Profile. For the Network Intensive Profile, please refer to table 2.2.1 in [RA2 Chapter 2](../../../ref_cert/RA2/chapters/chapter02.md).
```
pgoyal01, line 91:
```suggestion
Note that the minimum specifications listed above map to the requirements for the Basic Profile. For the Network Intensive Profile, please refer to table 2.2.1 in [RA2 Chapter 2](../../../ref_arch/kubernetes/chapters/chapter02.md#221-cloud-infrastructure-software-profile-capabilities).
```

doc/ref_impl/cntt-ri2/chapters/chapter03.md
```diff
@@ -72,22 +72,23 @@ The network fabric shall comprise redundant layer 2 connectivity between hosts.
...
 The jump / test host allows to install and manage the operation of the RI2. For example, the host can be used to execute the installer software to deploy and configure the RI2 on the other physical or virtual servers. Once installed, the host may be used to run test tools or it may be used to generate test traffic intended for performance measurements. To support the latter, the server provided for the jump / test host shall meet similar hardware requirements as the master (controller) / compute hosts. If the jump host is used only for deployment or execution of functional tests after deployment, a VM with the resources as outlined below is sufficient.
 
 Each server shall meet the following minimum specifications:
 
 * CPU
-  *  2x x86_64 CPU sockets, providing 20 cores each, 40 simultaneous multi-threads (SMT), at 2.2 GHz
+  *  2x x86_64 CPU sockets, providing 20 cores each, 40 simultaneous multi-threads (SMT), at 2.2 GHz. If using VM's, the servers must be equipped with at least 16 vCPUs.
 * Memory
-  * 256 GB RAM
+  * 32 GB RAM
```
pgoyal01, line 84: The 32 GB RAM, 320 GB disk etc are the maximum assigned to a Pod. So is the intent to make the max pod ask to be a minimum for the Server?

rihabbanday: The [ RA2 documentation](https://github.com/cntt-n/CNTT/blob/master/doc/ref_arch/kubernetes/chapters/chapter04.md) specifies "at least" 32GB RAM/320 GB disk for the worker nodes so I assumed that it refers to the minimum requirement of the worker nodes (assuming that node=server in our 5 server k8s setup). Is there any formal definition of Nodes/Pods/Servers within the Anuket documentation?

jgu17: I assume that the 32GB/320GB doesn't count for the workloads or additional components in a specific RI deployment such as logging and monitoring? 

@pgoyal01 I am not sure I follow what do you mean by "assigned to a pod", do you mean all the k8s controller pods together should be assignedto at least 32GB?

pgoyal01: @jgu17  What I am trying to state, not very clearly as usual, is that the physical server has to be sized larger than the specification below (and it is stated in the specs also).  BTW if the server only supports 32 GB then that is the total available for OS, monitoring, tools, utilities, nodes, etc.

From RA2 [4.2 Kubernetes Node](https://github.com/cntt-n/CNTT/blob/master/doc/ref_arch/kubernetes/chapters/chapter04.md#42-kubernetes-node) (emphasis mine):

Ref | Specification | Details | Requirement Trace | Reference Implementation Trace
-- | -- | -- | -- | --
ra2.ch.010 | Local Filesystem Storage Quantity | The Kubernetes Nodes must be equipped with local filesystem capacity  of **at least 320GB** for unpacking and executing containers. Note, _extra  should be provisioned_ to cater for any overhead required by the  Operating System and any required OS processes such as the container  runtime, Kubernetes agents, etc. | e.cap.003 |  
ra2.ch.012 | Kubernetes Node RAM Quantity | The Kubernetes Nodes must be equipped with **at least 32GB of RAM**.  Note, _extra should be provisioned_ to cater for any overhead required by  the Operating System and any required OS processes such as the container  runtime, Kubernetes agents, etc. | e.cap.002 |

trevgc:
The purpose of specifying infrastructure requirements in chapter 2 i.e. "Requirements for Labs" is to agree on a "standard / common" environment for 1) community collaboration on development and testing of RI-2 and 2) allow anybody to setup their own environment to easily deploy RI-2 (i.e. purpose is not requirements for a production environment). IMO we should try to make the requirements as flexible as possible for current and anticipated needs but also very practical to minimize cost and support overhead. To this end should we specify a 1) minimum configuration 2) ideal configuration? Are there different deployment options that would impact the infrastructure requirements and should these be captured here? 3.2.2 Connectivity specifies 5 physical networks (from the original Pharos spec.) ... is this still a requirement for RA-2? Is out-of-band hardware management network being used for RI-2 (I am not sure that lab users have access and may only be used by lab admins). Does RI-2 require physically separate Admin and Public networks for dev/testing?

doc/ref_impl/cntt-ri2/chapters/chapter03.md
trevgc: 32 GB RAM, 320 GB disk seems really low for a lab machine. I think there should be distinction between minimum called for by the spec vs what is practical for a lab environment to provide flexibility needed for dev/testing environment

acmacm: Agree with Trevor and Lincoln on these specifications for RAM and Storage. Fitting-out a Lab installation with minimums that are so low (32/320) means replacing that system when the minimums rise in a short amount of time, and there are many such systems in the lab. Lincoln's (256/1024)G suggestion has some legs for the future.

acmacm:
Trevor, I would like to see a configuration that remains useful in the long-term. Also, I would like to see the networking specs: it's time to include 25G to 40G links to the switch, if this topic is in-scope.

michaelspedersen:
Looking at [RA2 2.2.1](https://github.com/cntt-n/CNTT/blob/master/doc/ref_arch/kubernetes/chapters/chapter02.md#221-cloud-infrastructure-software-profile-capabilities) these values (mem, storage) are the minimum per-pod requirement. For the node/server amount of resources must be quite a bit higher, both to support multiple pods per node but also for the overhead of the OS and K8s deployment.

I am thinking something along the lines of:
- 2x (min.) 20 core x86_64 CPUs, for a total of 40 (or more) cores (80+ threads with SMT)
- 256+GB memory (DDR4)
- 1+TB storage (not sure if this should be raised to 1.5TB minimum)
  - Should be "solid" storage (NVMe and SSD).
  - Additional bulk storage can be added as HDD if needed (I have not looked at details for storage requirements)
- 25Gbps+ network solution
  - Ideally at least 2 ports per socket (NUMA)
  - Additional NICs/ports (10Gbps+) might be useful as well
- (to be considered) Additional PCI devices (e.g. FPGA, GPU, QAT) based on the mentions of device plugins
  - These could potentially be added later on
 
I think the above also follows some of the previous comments, but any additional input is appreciated.

rihabbanday:
@trevgc @lylavoie @acmacm I agree to the points listed above - We had discussed the referenced issue #2397 in last week's RI2 call and concluded that it would be best to derive these minimum requirements from RA2 Basic Profile. But the minimum requirements in RA2 have clearly raised new questions which I believe should be clarified, best in a separate issue. I can't recall if there was a discussion last year when these minimum specifications, physical networks, etc. were originally documented in this RI2. I'll probably refactor the entire chapter, and will stick to the ideal specifications, something similar to what @michaelspedersen has proposed and what we are currently using in Kuberef. Min vs ideal requirements proposed by @trevgc sounds like a good proposal, but the question remains still open - how do we define the minimum requirements?

lylavoie:
@rihabbanday I agree the minimum requirements for the architecture are for the RA2 documentation.  But for the lab setup, I think those requirements would 1) likely be above the minimum (I'm thinking minimum is a true minimum), and 2) might need to address differing scenarios between development or trials, etc.

@acmacm For 25G networking, that would bring things inline with the requirements for the RI1 lab, which already lists 25G, so I support adding that into the documentation.

jgu17:
I agree with what was suggested above to have the "desired" or recommended lab spec is more important than to have the true minimum for the RI. The dev/trial vs reference lab In my mind is about the number of servers, even though it might not hurt to state that the dev configuration can have less than ram/disk etc. But to have some consistency across the lab pods (such as Intel lab), to use the same default ram/cpu/disk spec even for dev, it might have the benefits here.

The other question I have, but do not want to digress from what Rihab wants to focus here, is the coming Airship RI-2 has its own RI-2 lab requirement, most of which is similar to the desired/ideal spec Michael raised up. A quick question here is what is everyone' thought whether we want to list lab requirements separately for Airship and Kuberef, and find the common spec?

gkunz:
Hi all,
to @jgu17's question: I would like to go for (wish for) a common specification valid for Kuberef and Airship. Separate requirement sets are just confusing and do not add value. Moreover, as mentioned above, the lab requirements should be a solid configuration above the minimum specs which should remain valid for some time. I assume we should be able to find such a configuration which suits both Kuberef and Airship.

pgoyal01:
Want to second a number of similar comments. Can we also take 'workload" requirements into account such as open-source 5G core, Clearwater IMS, etc..

michaelspedersen:
> Want to second a number of similar comments. Can we also take 'workload" requirements into account such as open-source 5G core, Clearwater IMS, etc..

I agree it might be good to take some of these into considerations. I would however assume that most/all of them should work on "generic" hardware configurations, otherwise (at least in my opinion) it defeats the purpose of cramming it into a K8s cluster to begin with.

rihabbanday:
As agreed in the RI2 meeting today - start with ideal requirements for RI2 that would be valid for both Kuberef and Airship.

lylavoie:
I think this looks good now.

pgoyal01 (approved):

gkunz (approved):

pgoyal01:
@walterkozlowski This is ready for merge

rihabbanday:
@walterkozlowski @wmk-admin could you please merge/approve this PR? It has 3 approvals, but I am not able to merge it

PR #2478, opened 2021-06-02
Add Storage Model diagram

jbhartley:
Storage Model provide simple Cost vs Performance with Consumption Model overlay classification.


jbhartley:
Will resubmit following naming convention

PR #2514, opened 2021-06-28
[RM] nit: Add black line before Table 9-1 ref

collivier:
Else it's falsy considered as part of the tab


pgoyal01:
This is merged and should be closed

PR #2572, opened 2021-08-25
[RM Chap 3.5.7] Added short version of Kubernetes Networking Semantics

qabpea:
Moved RM Network Semantics into dedicated branch


walterkozlowski (changes requested):
@qabpea, this is all true but I am not sure that our RM is a place for a critique of K8s, at least not without any proposal how to mitigate this state of affairs.  Can't we say something less tied up with Kubernetes; the latter has its place in RA-2?  Perhaps we need only to say that there is a gap in addressing a problem of multi networking in currently leading container management systems used in telecommunications (like Kubernetes), and there is a possible way of addressing it via Networking Semantics, and then refer to RA-2 for the K8s case.

qabpea:
> @qabpea, this is all true but I am not sure that our RM is a place for a critique of K8s, at least not without any proposal how to mitigate this state of affairs. Can't we say something less tied up with Kubernetes; the latter has its place in RA-2? Perhaps we need only to say that there is a gap in addressing a problem of multi networking in currently leading container management systems used in telecommunications (like Kubernetes), and there is a possible way of addressing it via Networking Semantics, and then refer to RA-2 for the K8s case.

I do not see it as a critique of K8s, it is stating a well know fact. I can soft it up a bit, but the reality is that there is very little there. Perhaps best to remove this part and only keep what is in the ref architecture?

pgoyal01:
@qabpea @walterkozlowski 
May I suggest a rewrite to cover both OpenStack and Kubernetes networking as below rather than as a contrast.  The rewrite, however, suffers from that it doesn't address "semantics" -- suggestions?

--------------------------------------

3.5.7 Networking Semantics

OpenStack Neutron provides “network connectivity as a service” between interface devices; it implements the OpenStack Networking API. It manages virtual networking and access to physical networks. Neutron provides actual layer 2 connectivity to compute instances and supports subnets. A subnet is attached to a network and describes an IP network; a subnet has a CIDR and a gateway IP address. Neutron supports dedicated static IP addresses or DHCP,  and Floating IP addresses for dynamic traffic rerouting. These and other capabilities, including support for SDN, allow, for example, the instantiation of network services such as L3aaS and LBaaS and their interconnection. Networking APIs and other networking capabilities can be added through the use of the ML2 plugin.
  
Kubernetes networking can be divided into two parts, built-in network functionality available through the pod's mandatory primary interface and network functionality available through the pod's optional secondary interfaces. Kubernetes currently only allows for one network, the *cluster* network, and one network attachment for each pod. All pods and containers have an *eth0* interface, this interface is created by Kubernetes at pod creation and attached to the cluster network. All communication to and from the pod is done through this interface. To only allow for one interface in a pod removes the need for traditional networking tools such as *VRFs* and additional routes and routing tables inside the pod network namespace. The support for traditional network orchestration is non-existent in Kubernetes, there is no network orchestration API and there is no way to create L2 networks. Multi networks, pod multi-network attachments, or network orchestration is supported by using a [*Container Network Interface*](https://github.com/containernetworking/cni) multiplexer such as [Multus](https://github.com/k8snetworkplumbingwg/multus-cni). The [*Network Plumbing Working Group*](https://github.com/k8snetworkplumbingwg/community) has produced the [Kubernetes Network Custom Resource Definition De-facto Standard](https://docs.google.com/document/d/1Ny03h6IDVy_e_vmElOqR7UdTPAG_RNydhVE1Kx54kFQ/edit). This document describes how secondary networks can be defined and attached to pods.

qabpea:
I updated the section according to @pgoyal01 suggestion

walterkozlowski (approved):

ulikleber (approved):

PR #2583, opened 2021-08-27
[RC2] Remove DNS configMap nameserver testing

collivier:
It stops and recreate coredns pods which is incompatible with testing
live env.

An there is no single test in sig_network_serial, it can safely
removed from manatory test cases.


rihabbanday (approved):

collivier:
Agreed in RA2 meeting

PR #2594, opened 2021-09-02
[RM Ch 7 &8] Update references.md with 

walterkozlowski:
Updates References with Security references required for Multi-Cloud Security (Issue#2314)


karinesevilla (approved):
@walterkozlowski  Section 7.10 of security chapter contains also references. It is a sub set of the references gathered in the common/references document. It is better to centralise all the references in one unique place. For me, section 7.10 is now becoming worthless and can be removed

pgoyal01 (approved):

pgoyal01:
@karinesevilla "For me, section 7.10 is now becoming worthless and can be removed" -- agree but before we do that, we need to make sure that all of the references in Section 7 have been captured in the common references.

PR #2615, opened 2021-09-10
[RA2 Ch4] add spec for ipam cni

### LINKED ISSUE #2275: [RA2 Ch04] Add requirement on IPAM capability
CH04 should include a requirement for a K8S plugin capable of IP address management

rgstori:
closes #2275
I've added a spec for an IPAM CNI for secondary network interfaces, highlighting it's for all nodes of the cluster (otherwise static or host-local would be ok)
I don't think we need another one for primary interfaces as they're managed by k8s natively.
what about an external dhcp? would that be ok?


CsatariGergely (approved):

pgoyal01:
@rgstori "what about an external dhcp? would that be ok?"
I think yes

pgoyal01 (approved):

karinesevilla (approved):

Issue #2618, opened 2021-09-17
[RA1 All] Correct typos'/language aligned with GSMA doc


pgoyal01:
While creating the RA-1 documentation for GSMA a number of typographical and language issues were identified and corrected. A PR will be created for each Chapter to make the necessary corrections.

pgoyal01:
Now that the related PRs have been merged, closing this issue.

PR #2625, opened 2021-09-21
[RA-1 Ch05] Typo's, spelling, language corrections

pgoyal01:
Made corrections discovered during creation of GSMA doc

Partially Fixes Issue #2618

### DIFF
--- a/doc/ref_arch/openstack/chapters/chapter05.md
+++ b/doc/ref_arch/openstack/chapters/chapter05.md
@@ -18,14 +18,14 @@ implementations that **get certified by RC** can be considered as Anuket RA Conf
 The Chapter presents the APIs for the core OpenStack services defined in Chapter 3 and a
 consolidated view of these and other APIs that are of interest.
 
-OpenStack is a multi-project framework composed of services evolving independently. It is not enough to rely only on the
+OpenStack is a multi-project framework composed of independently evolving services. It is not enough to rely only on the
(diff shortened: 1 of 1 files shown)


pgoyal01:
@collivier w.r.t. the maximal and minimal microversions, I was wondering if statements in line with the following need to be added:

As stated above, maximal microversions may be specified for a service in a given OpenStack release; the service supports all features from all of the mciroversions prior to and including the maximal microversion. This set of features for the service is the feature list of interest to this RA.  Utilising a predecessor of the maximal microversion will result in missing features.  Thus, in this RA we specify a minimal microversion - the version that has the complete set of features of interest to this RA viz., the maximal microversion specified for an OpenStack release.  

An implementation may choose a successor microversion of the RA minimal microversion so long as it is understood that some of the newer features in the selected microversion may not work with the microversions specified in the RA for other services.  Because of the OpenStack backward compatibility rules, the feature set of the minimal microversion are supported in the successor microversions.

ildikov (approved):
Thanks for the fixes!

karinesevilla (commented):
@pgoyal01 Changes from Cédric on API microversion or template version have been merged on Lakelse. I think it is the only chapter with modifications merged. These changes must not to apply to Kali release. We must be careful about it.

karinesevilla (approved):

walterkozlowski (approved):

PR #2650, opened 2021-10-12
[rst]: WIP: First try of tools

CsatariGergely:
To do the conversion we will need:
  - A pre processing script to remove ToC-s, heading numbers, raw html and maybe more
  - A script to run pandoc with all the correct parameters
  - Maybe a post processing script to fix tings in the rst files what we could not do in pre-processing
    - Maybe generating the index.rst-s


CsatariGergely:
> 
> 
> Frankly I don't think we need an additional py scrit. pandoc seems enough if we fix the orig files (see RC1 and RC2)

What this script tries to do is to do the fixing of the orig files for us to reduce the amount of manual work and human errors.

gkunz:
@collivier, @CsatariGergely: I am not happy with the tables in rst: rst supports either grid tables of simple tables. The former is the default output of pandoc in case of text wrapping per cell, BUT these cells are super tedious to manage since everything needs to align. Any thoughts / recommendations from your side?

CsatariGergely:
> 
> 
> I also saw that I think there is no option out of manual processing. But I would rather consider that the benefits are much bigger than the cost here.

I also try to achieve some consistency in how do we convert the files by codifying what do we do.

PR #2653, opened 2021-10-14
[RC1] Fix markdown lists in chapter2

collivier:
It's asked by pandoc when converting to RST.


pgoyal01:
The bogometer sizes are large -- is there a way to scale them?

gkunz:
> 
> 
> Will the chapters have numbers? Else how do we reference them?

One goal of the conversion to rst is to leverage the automatic numbering provided by the build tool (Sphinx). References can still be made to the section headers, but without including numbers in the references themselves. This helps a lot in case of inserting new sections etc.

collivier:
I'm converting in RST via this PR. You may see a couple of additional commits in the next hours.

PR #2658, opened 2021-10-17
[RM Ch 7.9.7] Monitoring for prod and non-prod

walterkozlowski:
Updating table in Ch 7.9.7 differentiating between Monitoring/Auditing requirements between prod and non-prod environments.  Partially addresses Issue#2541


walterkozlowski:
@pgoyal01, @ulikleber, @karinesevilla: I have made changes as we discussed in the RM meeting:
1. I have added explanation before the table. Most of the requirements now are for both Prod and NonProd.  Whenever a requirement mentions only Prod, it means that this requirement is optional for NonProd, hence I removed the requirements for NonProd only (formerly all requirements from req 22 onwards)
2. I have split up req 1 into two
3. I have changed req 12 to cover both Prod and NonProd
4. I have re-numbered requirements to cover for the split of req 1.
Since things are getting messy, I suggest we accept this version and merge it, and then we can raise some word-smithing PRs if needed.  So please approve if you can.

PR #2723, opened 2021-11-13
[RM Ch 8] Chapter 8 proofreading

walterkozlowski:


doc/ref_model/chapters/chapter08.md
```diff
@@ -22,11 +22,11 @@
...
+It is unrealistic to expect that a private cloud can cost effectively and meet the need of all loads, including peak loads and disaster recovery. It is for that reason that enterprises will implement a hybrid cloud.  In a hybrid cloud deployment, at least two or more distinct cloud infrastructures are inter-connected together.  In a multi-cloud the distinct cloud infrastructures of the hybrid cloud may be implemented using one or more technologies.  The hybrid multi-cloud infrastructure has differences requiring different abstractions. These hybrid multi-clouds can be considered to be federated.
```
pgoyal01, line 25: @walterkozlowski I think the "and" (i) completely changes the intent and (ii) makes the sentence awkward (e.g., "cost effectively" what?).  The intent of the existing sentence is that the private clouds would ve very expensive if they were to host all workloads, in particular, those workloads whose load characteristics vary significantly (very high standard deviation).

```suggestion
It is unrealistic to expect that a private cloud can cost-effectively meet the needs of all workloads when the private cloud must also meet the needs for peak loads and disaster recovery. For this reason alone, enterprises will need to implement a hybrid cloud.  In a hybrid cloud deployment, at least two or more distinct cloud infrastructures are interconnected together.  In a multi-cloud, the distinct cloud infrastructures of the hybrid cloud may be implemented using one or more technologies.  The hybrid multi-cloud infrastructure has differences requiring different abstractions. These hybrid multi-clouds can be considered to be federated.
```

walterkozlowski:
@pgoyal01, with "cost-effectively" instead of "cost effectively" - it makes sense, and wanted to suggest as well.  Previously I read "cost" in "cost effectively" as a verb.  I accepted the change to "cost-effectively".

pgoyal01 (approved):

gkunz (approved):

doc/ref_model/chapters/chapter08.md
```diff
@@ -22,11 +22,11 @@
...
+It is unrealistic to expect that a private cloud can cost-effectively meet the needs of all workloads when the private cloud must also meet the needs for peak loads and disaster recovery. For this reason alone, enterprises will need to implement a hybrid cloud.  In a hybrid cloud deployment, at least two or more distinct cloud infrastructures are interconnected together.  In a multi-cloud, the distinct cloud infrastructures of the hybrid cloud may be implemented using one or more technologies.  The hybrid multi-cloud infrastructure has differences requiring different abstractions. These hybrid multi-clouds can be considered to be federated.
```
ildikov, line 25: I think this should be either "connected together" or "interconnected."

ildikov (approved):
I added a nit in line, but it's not a blocking one so will approve.

Thanks,
Ildikó

PR #2656, opened 2021-10-15
[RI2] Converting docs from markdown to rst

gkunz:
This PR implements cleanup of the documentation and conversion from markdown to rst.

Implements #2649

PR #2740, opened 2021-12-07
[RI2] Dos2unix chapter 1, 3, 4 and 5

collivier:
It basically fails the RST conversion.


doc/ref_impl/cntt-ri2/chapters/chapter01.md
```diff
@@ -1,4 +1,4 @@
-﻿# 1. Overview
+# 1. Overview
 
 ![State](../figures/bogo_dfp.png) <!-- width="35" -->
```
pgoyal01, line 3: I thought we agreed to delete the bogometers

gkunz: @pgoyal01 we did. They have been removed in PR #2656 which implements the rst conversion for RI2.

collivier:
Fully agree to remove it in a second change.
Here the proposal is to remove the dos hidden chars which fails the conversion.

pgoyal01 (approved):

gkunz:
@collivier there is a PR #2656 which implements the rst conversion. I'll check this PR for hidden characters, but those files have been processed by pandoc, so I don't expect any.

collivier:
@gkunz The pandoc conversions fails here (and atom doesn't interpret this lines as title) - double checking
I would encourage to fix the md first (agree that pandoc should remove them automatically).

gkunz:
@collivier I checked the files in the PR mentioned above, but pandoc has, as we assumed, unified the file encoding. At least running dos2unix does not change the files. I also checked the resulting output, but couldn't spot issues with the headings. Maybe you can quickly take a look and check if the headings are correct in my PR since you have encountered problems during conversion. Thanks.

collivier:
@gkunz Could we merge it? trivial but crucial.

gkunz:
@collivier we can merge it, but it won't matter. See #2656. I'd rather like to merge #2656.

collivier:
@gkunz yes it will be removed by your next change - but the point is to apply the change as we don't know when rst will be effective.

gkunz:
The problem is that I cannot merge it. @rihabbanday?

rihabbanday (approved):

PR #2742, opened 2021-12-07
[nit] fix falsy links

collivier:
detected by sphing linkcheck when converting to rst.

### DIFF
--- a/doc/ref_cert/RC2/chapters/chapter04.md
+++ b/doc/ref_cert/RC2/chapters/chapter04.md
@@ -20,7 +20,7 @@ The following is a Requirements Traceability Matrix (RTM) mapping Test Case, and
 
 ### Test Case Traceability to RA2 Requirements
 
-This section focuses on the test cases covering the requirements in [RA-2 Chapter 4.9](../../../ref_arch/kubernetes/chapters/chapter04.md#49-kubernetes-workloads) for Kubernetes workloads.
+This section focuses on the test cases covering the requirements in [RA-2 Chapter 4.9](../../../ref_arch/kubernetes/chapters/chapter04.md) for Kubernetes workloads.
 
 | RM/RA Ref | High-level test definition | Test name and project | Priority |
 |---|---|---|---|
--- a/doc/ref_cert/RC2/chapters/chapter04.md
+++ b/doc/ref_cert/RC2/chapters/chapter04.md
@@ -20,7 +20,9 @@ The following is a Requirements Traceability Matrix (RTM) mapping Test Case, and
 
 ### Test Case Traceability to RA2 Requirements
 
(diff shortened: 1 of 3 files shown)


doc/ref_cert/RC2/chapters/chapter04.md
```diff
@@ -20,7 +20,7 @@ The following is a Requirements Traceability Matrix (RTM) mapping Test Case, and
 
 ### Test Case Traceability to RA2 Requirements
 
-This section focuses on the test cases covering the requirements in [RA-2 Chapter 4.9](../../../ref_arch/kubernetes/chapters/chapter04.md#49-kubernetes-workloads) for Kubernetes workloads.
+This section focuses on the test cases covering the requirements in [RA-2 Chapter 4.9](../../../ref_arch/kubernetes/chapters/chapter04.md) for Kubernetes workloads.
```
pgoyal01, line 23: The problem is that now the reference is to the main chapter and someone has to search to go to a specific section. Until RA2 is converted, the reference should remain. To make it easy to change, maybe the references can be named and listed sperartely, and we will only need to change in one easily visible place.

gkunz: I agree with Pankaj here

collivier: As far as I understand I cannot be more precise with markdown. Let me double check why linkcheck is considering this link as false.
Yes it's simple in RST.

CsatariGergely (approved):

doc/ref_cert/RC2/chapters/chapter04.md
gkunz, line 23: The reason for the failing link check seems to be that Sphinx rightfully does not use the header numbering when generating the header reference. In the final build on read the docs, this is the resulting link:

https://cntt.readthedocs.io/en/latest/ref_arch/kubernetes/chapters/chapter04.html#kubernetes-workloads

So, instead of removing the reference to the heading, I suggest to change it like this: 

```suggestion
This section focuses on the test cases covering the requirements in [RA-2 Chapter 4.9](../../../ref_arch/kubernetes/chapters/chapter04.md#kubernetes-workloads) for Kubernetes workloads.
```

gkunz (changes requested):

doc/ref_cert/RC2/chapters/chapter04.md
collivier: At the end of the day we will leverage rst link (internal or intersphinx)

collivier: done

pgoyal01 (approved):

PR #2750, opened 2021-12-10
Fix links and remove bogometers

collivier:
It's a key fix prior to rst conversion.


collivier:
Crucial. As it's trivial, I do merge it

PR #2752, opened 2021-12-10
Clean concreate deps in the requirement file

collivier:
They have to be put in an upper constraints file.
They seem outdated (sphinx<2) and at the time of writing useless.
    
It also removes falsy deps.


tox.ini
```diff
@@ -6,13 +6,9 @@ skipsdist = true
 
 [testenv:docs]
 basepython = python3
-deps = -r{toxinidir}/requirements.txt
+deps = -r{toxinidir}/test-requirements.txt
```
gkunz, line 9: Just curious: what's the semantic difference between test-requirements and requirements here?

collivier: @gkunz: requirements.txt is for package dependencies (especially in case of PBR which defacto reads it). Here it's just about dependencies needed for building docs ; no package at all. Yes maybe too much precise :)

gkunz: ok, thanks! Subtle difference.

gkunz (approved):

PR #2753, opened 2021-12-10
[RC1] Convert md to rst

collivier:
done by pandoc + manual edit especially tables in chapter02 and 03.


collivier:
No review activity but it must be merged to evaluate all issues in the biggest stream documents.
Let's go (we would revert it if needed).

PR #2765, opened 2021-12-15
[COMMON] Put rtd settings in tree

collivier:
See https://readthedocs.org/projects/cntt/builds/15541950/


collivier:
Sorry I do merge it quickly as the gate is failing https://readthedocs.org/projects/cntt/builds/15541950/
And I don't have the right to modify config by hand.

PR #2766, opened 2021-12-16
[RA1] md files: first clean for RST conversion

karinesevilla:
Back return removed
ToC removed
Numbering of titles and  sub-titles removed
HTML tags removed for tables and figues captions
Lines added after titles
Lists: harmonisation of bullets, alignment, lines added
Images: HTML tags removed

PR #2783, opened 2022-01-17
Fix CODEOWNERS

collivier:
We shouldn't block all changes in all streams by TSC Co-Chair [1].
See [2] which cannot be merged by @karinesevilla as stream leader.

[1] https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
[2] https://github.com/cntt-n/CNTT/pull/2766

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -8,31 +8,31 @@
 *       @walterkozlowski
 
 # Governance
-/doc/gov/                            @steelescotr @bobmon01 @walterkozlowski
+/doc/gov/                            @walterkozlowski
 
 # Reference Model
 /doc/ref_model/                      @walterkozlowski
 
 # Reference Architectures 1
-/doc/ref_arch/openstack/             @karinesevilla @walterkozlowski
+/doc/ref_arch/openstack/             @karinesevilla
 
 # Reference Implementation 1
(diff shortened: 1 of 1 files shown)


karinesevilla (approved):

pgoyal01:
We need an alternative to the stream leads because if the WSL creates a PPR then they should not merge it. Also, Scot Steele and Bob Monokan are WSL leads for governance stream.

walterkozlowski (approved):

collivier:
@pgoyal01 Sorry I didn't see your comment (you didn't ask for an update). Merge is just a validation of the review process. There is nothing wrong that the author merges its PR if it was reviewed by the main contributors. I would consider as much worse that a PR is blocked for a while waiting the few people who has the right to merge it. I would frankly agree that we add another a big RA1 contributor as second but there is no point to add TSC Co Chairs as dependencies.

pgoyal01:
@collivier @walterkozlowski I think this is wrong. An author should not be able to merge their own PR. An alternate is necessary to ensure that all comments are addressed -- not everyone makes their suggestions as changes and an alternate merger ensures that all comments are addressed and there are not just 3 approved reviewers.

Issue #2801, opened 2022-01-31
[RA2] Remove markdown files


collivier:
Switching to RST asks to remove all markdown files too else the content could be falsing linked to the former markdown.
\+ linkcheck at the end

CsatariGergely:
Uh-oh, thanks for recognizing this.

PR #2803, opened 2022-01-31
[RA2]: Removal of md files

CsatariGergely:
Closes: #2801


gkunz (approved):

pgoyal01 (commented):
Should we delete or create a md files directory in case if we need to check some items?

CsatariGergely:
> Should we delete or create a md files directory in case if we need to check some items?

I think we should just delete them. The md directory gets obsoleted very fast and they are in the git history for ever.

rgstori (approved):

petorre (approved):

PR #2879, opened 2022-03-21
[RM Ch] conf.py add numfig

pgoyal01:
Added numfig = True

This will enable Figure numbering.

Here is a suggested use for the Figure 1-1 

LIne 54: within the text use :numref:`Fig-1-1` and the Figure block as:

.. _Fig-1-1:
.. image:: ../figures/ch01_scope.png
   :alt: "Scope of Reference Model"

   Scope of Reference Model

@collivier @gkunz Can you confirm? Thanks

### DIFF
--- a/doc/ref_model/conf.py
+++ b/doc/ref_model/conf.py
@@ -19,3 +19,6 @@ linkcheck_ignore = [
 ]
 autosectionlabel_prefix_document = True
 autosectionlabel_maxdepth = 3
+numfig = True
+numfig_format = {'figure': 'Figure %s', 'table': 'Table %s',
+                 'code-block': 'Listing %s', 'section': 'Section %s'}


gkunz:
Hi @pgoyal01, yes, I can confirm. Since the indentation of the caption is important to make this work, I copied your example below as code block for reference for others reading this.

Moreover, you could also use the `:name: Fig-1-1` directive of the figure instead of the `_Fig-1-1` label, but that is just a matter of taste.

Final nitpick: I recommend not using numbers in the labels, but rather a descriptive label and leave the numbering to Sphinx.

```
.. _Fig-1-1:
.. image:: ../figures/ch01_scope.png
   :alt: "Scope of Reference Model"

   Scope of Reference Model
```

gkunz (approved):
LGTM

walterkozlowski (approved):

Issue #2881, opened 2022-03-21
[RA2] Clean all dead links to markdown files


collivier:
See sphinx-build -W -b linkcheck . build

rgstori:
@collivier what's the proper way to have links to other RA2 and RM files in RST?
like this?

>:ref:\`ref_model/chapters/chapter08:chapter 8 Hybrid Multi-Cloud\`

PR #2882, opened 2022-03-21
[RA2] add tox to build RA2

collivier:
It will allow gating all PRs vs RA2.
It also removes all trailing whitespaces.

It's worth mentioning that it remains to wrap the files
and to fix all links (linkchecks is currently falsy disabled) [1].

[1] https://github.com/cntt-n/CNTT/issues/2881


CsatariGergely (changes requested):
I do not see a tox.ini or a conf.py. I think this current commit contains only the removal of trailing whitespaces.

CsatariGergely (approved):

gkunz (approved):

PR #2892, opened 2022-03-25
[docs] fix for Jinja2 3.1.0 breaking builds with old versions of Sphinx

### LINKED ISSUE #2891: [docs] Jinja2 3.2.0 breaks builds with older versions of Sphinx
A recent update of Jinja2 breaks builds using the old version of Sphinx we are using. We still use Sphinx < 2 to build the overall documentation containing both markdown and rst files. Newer versions seem to have problems with this.

Once all files have been converted to rst, we can uplift Sphinx (see individual builds which already today use much newer versions of Sphinx).

See also https://github.com/readthedocs/readthedocs.org/issues/9037

As a temporary solution, I suggest to limit jinja2 to < 3.2.0 until we can remove the version restriction in the overall build.

gkunz:
A recent update of Jinja2 breaks builds using the old version of Sphinx we are
using. We still use Sphinx < 2 to build the overall documentation containing
both markdown and rst files. Newer versions seem to have problems with this.

Once all files have been converted to rst, we can uplift Sphinx (see individual
builds which already today use much newer versions of Sphinx).

See also readthedocs/readthedocs.org#9037

As a temporary solution, I suggest to limit jinja2 to < 3.1.0 until we can
remove the version restriction in the overall build.

Fixes #2891

Signed-off-by: Georg Kunz <georg.kunz@ericsson.com>

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -7,3 +7,4 @@ sphinx-markdown-tables
 sphinx_rtd_theme
 recommonmark>=0.7.1
 commonmark>=0.9.1
+jinja2<3.1.0


gkunz:
@CsatariGergely, @collivier can you confirm this locally? rtd seems to have a fix in place, but when building all specs from the root, the build fails for me. Thanks

collivier (changes requested):
Sorry I would frankly prefer we ask for sphinx > 4 as done in ONAP.
RTD leverages project date creation as I reported to ONAP and to CNTT.

collivier:
I think we do consider removing obsolete sphinx ext when we get rid of md. This package may disappear from our requirements.txt.

Very convinced by updating Shinx and get rid of useless deps.

gkunz:
@collivier: agree, this is just a temporary fix and not so pressing since the build works on rtd. I support getting the conversion done and finally uplifting Sphinx.

CsatariGergely (approved):

gkunz:
@collivier it would be helpful to merge this as a temporary solution to enable builds of the entire documentation. Keeping the repo in a broken state is not ideal. So, kindly asking for a re-review.

@walterkozlowski this fixes broken builds as reported by @CsatariGergely and me.

Thanks

pgoyal01 (approved):

gkunz:
@walterkozlowski can you please review? Thank you!

collivier:
@gkunz I think this pr can be safely closed. ok?

gkunz:
Closing since we moved to a new version of Sphinx after converting all md files.

PR #2898, opened 2022-04-06
[RM] Fix all figures

collivier:


collivier:
must be merged - else all next patchsets would be false voted -1

PR #2920, opened 2022-04-28
[RC] Remove airship links

collivier:
These job links no longer exist since Airship is being archived.


collivier:
Hold on - it must be split

PR #2932, opened 2022-05-03
Remove automation gaps in RC1 Chapter 8

collivier:
These gaps are only about RI1 and shouldn't be part of RC1.
As the links are dead since Airship removal, we do clean
this obsolete and falsy content.


collivier:
I do merge it quickly as the gates are broken

PR #2940, opened 2022-05-03
[GLOBAL] Fix requirements in gates

collivier:
Please note that all RTD projects created before Oct 20 2020
must set bounds to enforce a compatible Sphinx version
(see manual).

It also sets autosectionlabel_maxdepth = 4 needed by RA1 and
removes the falsy intersphinx mapping.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -83,4 +83,4 @@ intersphinx_mapping = {
 }
 
 autosectionlabel_prefix_document = True
-autosectionlabel_maxdepth = 3
+autosectionlabel_maxdepth = 4
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,9 +1,8 @@
-sphinx!=1.6.6,!=1.6.7,!=2.1.0,!=3.0.0,!=3.4.2,<3 # BSD
+sphinx!=1.6.6,!=1.6.7,!=2.1.0,!=3.0.0,!=3.4.2,>=4.2.0  # BSD
 sphinxcontrib.direct-copy
 sphinxcontrib.readme-to-index
 sphinxcontrib.relative-link-corrector
 sphinx_material
 sphinx-markdown-tables
-sphinx_rtd_theme
 recommonmark>=0.7.1
 commonmark>=0.9.1


collivier:
@walterkozlowski @pgoyal01 @karinesevilla @gkunz please consider this patch as urgent as it fixes ra1 builds which are currently broken

test-requirements.txt
```diff
@@ -1,9 +1,9 @@
-sphinx!=1.6.6,!=1.6.7,!=2.1.0,!=3.0.0,!=3.4.2,<3 # BSD
+sphinx!=1.6.6,!=1.6.7,!=2.1.0,!=3.0.0,!=3.4.2,>=4.2.0  # BSD
 sphinxcontrib.direct-copy
 sphinxcontrib.readme-to-index
 sphinxcontrib.relative-link-corrector
 sphinx_material
 sphinx-markdown-tables
-sphinx_rtd_theme
+sphinx-rtd-theme>=1.0.0  # MIT
```
CsatariGergely, line 7: Shouldn't we use material theme in all documents to keep the consistent look and feel?

collivier: I just added the right constraints to match  the RTD rules - it could be safely removed as it's not selected in conf.py

Right now nothing works. not a big deal if I do create a second PR.

CsatariGergely (approved):

doc/conf.py
```diff
@@ -78,9 +78,5 @@
     '**/.tox',
 ]
 
-intersphinx_mapping = {
```
gkunz, line 81: What is the problem with this? Removing this breaks all :doc: directives in my local build.

gkunz (changes requested):

doc/conf.py
collivier: gkunz why would the global doc build require an intersphinx mapping to itself? It worked here and I'm very surprised it could fail in your env.

collivier: @gkunz could u help me to reproduce your failure?

$ rm -rf _build .tox ; LC_ALL=C tox
...
  docs: commands succeeded
  congratulations :)

and I will see all std:doc in _build/html/objects.inv

karinesevilla (approved):

doc/conf.py
gkunz: @collivier because the doc role is often used in conjunction with a defined intersphinx mapping. See https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html#the-doc-role

In our docs, see for example https://github.com/cntt-n/CNTT/blob/master/doc/ref_model/chapters/chapter01.rst#definitionsterminologyabbreviations

The build does not fail, but the links are not getting generated.

collivier: for sure ref is not. checking.

collivier: Yes you're right for :doc: - and I confirm :ref: fallback to local labels. We have to loop to ourselves here.

BYW we will have to remove all :doc: to common in RM and RA1 soon.

collivier: @gkunz done

gkunz (approved):

collivier:
RA1 and RM built successfully after https://readthedocs.org/projects/cntt/builds/16811156/

PR #2985, opened 2022-05-20
[RC2] Fix PDF rendering issues

collivier:
It also formats tables and fixes duplicated title.
It cleans bundle data which are manually removed by GSMA.

### DIFF
--- a/doc/ref_cert/RC2/index.rst
+++ b/doc/ref_cert/RC2/index.rst
@@ -1,53 +1,6 @@
-Reference Conformance - Kubernetes based platform for Telecom
-=============================================================
-
-This is the Kubernetes Based Reference Conformance (RC-2)
-
-Release Information
--------------------
-
-**Bundle: 7**
-
-**Version: 0**
-
-**Release Date: 4th January 2022**
-
-Bundle/Version History
-----------------------
-
-============== ============= ===================
-Bundle.Version Date          Note
-============== ============= ===================
-0.0            15th May 2020 First Initial Draft
-4.0            25th Sep 2020 Baraque Release
-5.0            29th Jan 2021 Elbrus Release
-6.0            1st Jul 2021  Kali Release
-7.0            4th Jan 2022  Lakelse Release
-============== ============= ===================
-
-Overall Status
---------------
-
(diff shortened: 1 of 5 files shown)


pgoyal01:
Why are we removing section headers? There is something wrong if we can't use section headers because they repeat?

doc/ref_cert/RC2/index.rst
```diff
@@ -1,56 +1,10 @@
-Reference Conformance - Kubernetes based platform for Telecom
-=============================================================
-
-This is the Kubernetes Based Reference Conformance (RC-2)
-
-Release Information
```
CsatariGergely, line 6: Don't we need this information somewhere?

CsatariGergely (changes requested):

doc/ref_cert/RC2/index.rst
```diff
@@ -1,56 +1,10 @@
...
+=================================================
+Anuket Reference Conformance for Kubernetes (RC2)
+=================================================
+
+.. include:: chapters/chapter01.rst
+.. include:: chapters/chapter02.rst
+.. include:: chapters/chapter03.rst
+.. include:: chapters/chapter04.rst
+.. include:: chapters/chapter05.rst
+.. include:: chapters/chapter06.rst
```
gkunz, line 10: I am generally ok with putting everything onto a single page. However, does this concept also support header numbering which is important in the RM and RAs?

collivier: yes it works.

doc/ref_cert/RC2/index.rst
collivier: this data is in conf.py. see the PDF rendering

collivier:
@pgoyal01 the bundle histroy, etc. is not part of GSMA and it fully breaks all titles. I can share 2 PDFs to highlight the issues if it helps.

karinesevilla (approved):

collivier:
@CsatariGergely @gkunz @pgoyal01 I reduced the scope of this PR.
It now only removes the bundle history and converts a few tables.

gkunz (approved):

PR #3017, opened 2022-06-06
[RA1] Remove falsy external docs

collivier:
:doc: asks for an external lookup which is false here.


karinesevilla (approved):

collivier:
@pgoyal01 sorry to merge it - it's trivial, safe but it blocks my next very big patch (and there is no PR dependency in github)

PR #3018, opened 2022-06-06
[RA1] nit: fix last false external lookup

collivier:


collivier:
@karinesevilla @pgoyal01 trivial

PR #3036, opened 2022-06-07
[CNTT] Add README.rst version

pgoyal01:
Created an rst version of the md file


collivier (approved):
Just a quick note: the link will change a little bit after my PR is merged + this file must be adapted per branch (it's not inside doc where interspinx does its work).

PR #3047, opened 2022-06-09
Add test-requirements for CNTT RTD

collivier:

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -5,11 +5,10 @@ envlist = docs
 skipsdist = true
 
 [testenv:docs]
-basepython = python3
-deps = -r{toxinidir}/test-requirements.txt
+basepython = python3.9
+deps =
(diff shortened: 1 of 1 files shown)


collivier:
must be merged quickly - sorry

tox.ini
```diff
@@ -5,11 +5,10 @@ envlist = docs
 skipsdist = true
 
 [testenv:docs]
-basepython = python3
-deps = -r{toxinidir}/test-requirements.txt
+basepython = python3.9
```
CsatariGergely, line 8: Why 3.9 is important? The Ubuntu I'm using has only 3.8, so I need to spent time on some hacking to make builds work.

collivier:
python3.9 is coherent with the other builds. We can leverage python 3 ; it's less precise and could lead us into trouble (I could explain what occurred in ONAP).

PR #3054, opened 2022-06-09
[RC1][RC2][RA1] Use :numbered: in all stream toc

collivier:


collivier:
let's merge it, we are currently lacking reviewers. we would revert it if really needed.

PR #3055, opened 2022-06-09
Unify Sphinx config

collivier:
They are several differences in Sphinx config which breaks
user naviguation.

The proposal is, by default, to use default Material values,
and to apply any change to all streams per PR.

It also selects Figure in all captions (default is Fig.).


collivier:
must be merge before the release

PR #3069, opened 2022-06-14
[RA1 README] Update schedule

pgoyal01:


collivier (changes requested):
I would prefer us to remove this data which are not exported to GSMA.
As for ra1, we should rather consider a tag in git if needed.

collivier:
@pgoyal01 another solution could be to leverage the only directive
https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html

.. only:: html

CsatariGergely:
> @pgoyal01 another solution could be to leverage the only directive
> https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html
> 
> .. only:: html

I would use this directive to maintain the consistency with the other documents.

CsatariGergely:
I removed the version information and moved it to the root doc in #3077

karinesevilla:
@CsatariGergely @collivier @pgoyal01 I am closing this PR, as a better solution was found. The releases history will be centralized for all documents

PR #3070, opened 2022-06-14
[ra2] fix ra2 date in readme

rgstori:


pgoyal01 (approved):

karinesevilla (approved):

collivier (changes requested):
I would prefer us to remove this data which are not part of GSMA.
We should rather tag git repository with ra2 version if needed.

collivier:
@rgstori @pgoyal01 another solution could be to leverage the only directive
https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html

.. only:: html

CsatariGergely:
> @rgstori @pgoyal01 another solution could be to leverage the only directive
> https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html
> 
> .. only:: html

I would prefer this option.

rgstori:
would removing it work? as in #3071

CsatariGergely:
I did remove the version information and moved it to a central place in #3077

PR #3071, opened 2022-06-14
[RA2] another README.rst

rgstori:
removed the bundle version section and moved toc to top

### DIFF
--- a/doc/ref_arch/kubernetes/README.rst
+++ b/doc/ref_arch/kubernetes/README.rst
@@ -1,16 +1,30 @@
 Kubernetes based Reference Architecture
 =======================================
 
-This is Kubernetes based Reference Architecture (RA-2)
+Table of Contents
+-----------------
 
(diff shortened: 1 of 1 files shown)


doc/ref_arch/kubernetes/README.rst
```diff
@@ -59,11 +17,42 @@ Table of Contents
...
+Required component versions
 ----------------------------------------------
 
 ========== ===================
 Component  Required version(s)
 ========== ===================
 Kubernetes 1.23
 ========== ===================
+
+Bundle/Version History
```
rgstori, line 29: remove as well?

CsatariGergely: I think we should remove only the version history as I've added that to the top page of Anuket specs in #3077

rgstori:
ok closing this, let's continue with #3077

PR #3126, opened 2022-08-19
[RA1] Capital letters in the title

karinesevilla:


pgoyal01:
@karinesevilla Writing Style Guide would state that the title was correct but for some reason some folks decided to capitalize in 2 stream and so now we have also to do the same

collivier (approved):

PR #3130, opened 2022-08-22
[docs] Fixing broken link to terms of reference

### LINKED ISSUE #3081: [pcac]: Terms of Reference link is broken

gkunz:
The link to the terms of references assumes that the pdf has been copied
to read-the-docs. This changes the link to instead point to the lastest
version of the terms of reference on GitHub. It should always be up-to-date
(i.e., the latest version) and not correspond to a released version of the
specs.

Fixes #3081

Signed-off-by: Georg Kunz <georg.kunz@ericsson.com>


CsatariGergely (approved):

collivier (dismissed):
Your proposal makes sens'.
But I would suggest to move it out of the tree. why not in a LFN / Anuket website or as a blob object.

gkunz:
We could indeed add the terms of reference and the current charter to the main anuket.io page. It weirdly enough currently has a link to a CLA sample(!) in the resources section of the page.

ulikleber (approved):

rgstori (approved):

steelescotr (commented):
If GSMA finds the PR acceptable I approve

collivier:
> Hi @collivier: looking at the current terms of reference document, it turns out that it is outdated because it still refers to CNTT. If there is need for keeping an official terms of reference document (to be discussed elsewhere), then we should update the document.
> 
> In the meantime, we should either remove the broken link or point it to the currently outdated document as it nevertheless provides some context. I think both options are better than a broken link.

I'm fine to remove it.

gkunz:
> > Hi @collivier: looking at the current terms of reference document, it turns out that it is outdated because it still refers to CNTT. If there is need for keeping an official terms of reference document (to be discussed elsewhere), then we should update the document.
> > In the meantime, we should either remove the broken link or point it to the currently outdated document as it nevertheless provides some context. I think both options are better than a broken link.
> 
> I'm fine to remove it.

updated in https://github.com/cntt-n/CNTT/pull/3130/commits/e54b9ea12f3b0877209d0bd4eb0a7b285d739d32

CsatariGergely (approved):

steelescotr (approved):

karinesevilla (approved):

PR #3142, opened 2022-09-02
[RA1 Ch01] Acronyms added

karinesevilla:
For Chapter 8 Conformance

### DIFF
--- a/doc/ref_arch/openstack/chapters/chapter01.rst
+++ b/doc/ref_arch/openstack/chapters/chapter01.rst
@@ -163,12 +163,20 @@ Abbreviations
      - Definition
    * - API
      - Application Programming Interface
+   * - BGP VPN
+     - Border gateway Protocol Virtual Private network
+   * - CI/CD
+     - Continuous Integration/ Continuous Deployment
+   * - CNTT
+     - Cloud iNfrastructure Task Force
    * - CPU
      - Central Processing Unit
    * - DNS
      - Domain Name System
    * - DPDK
      - Data Plane Development Kit
+   * - DHCP
+     - Dynamic Host Configuration Protocol
    * - ECMP
      - Equal Cost Multi-Path routing
    * - ETSI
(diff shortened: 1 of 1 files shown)


doc/ref_arch/openstack/chapters/chapter01.rst
```diff
@@ -163,12 +163,20 @@ Abbreviations
      - Definition
    * - API
      - Application Programming Interface
+   * - BGP VPN
+     - Border gateway Protocol Virtual Private network
+   * - CI/CD
+     - Continuous Integration/ Continuous Deployment
```
collivier, line 169: blank char?

karinesevilla: Okay, removed

pgoyal01 (approved):

collivier (approved):

CsatariGergely:
@collivier @karinesevilla  This pr had only two approvals. Why was this already merged?

karinesevilla:
@CsatariGergely This PR was considered as minor, as it is just the addition of missing acronyms.

collivier:
@CsatariGergely you should check instead the direct pushes to master without PR. It obscures the process review (I could  be ok for a self "nit" PR merge) but it forbids the job verification and introduces new bugs to all substreams at the end.

CsatariGergely:
@collivier @karinesevilla I understand that you are in a rush and I think it is important to finish the GSMA work in time. Also I think it is important to follow the processes we agreed on. (Some more philosophy on this [here](https://github.com/cntt-n/CNTT/pull/3154#issuecomment-1242044952))
I totally agree that we should not push anything to master without a pr. @wmk-admin please follow the pr procedure defined [here](https://github.com/cntt-n/CNTT/blob/master/CONTRIBUTING.rst).

collivier:
@CsatariGergely  I would have expected more contributions than policies. I don't see any change proposal from all your comments counting reviewer. **TSC Co Chair activities is about the meeting itself noting more** -
Even if I generalize to the TSC activities (what's false here), TSC should help the project to be successful not the opposite.

Comparing direct commit and PR with less than 3 reviewer is technically speaking fully false.

@karinesevilla Would you mind adding a slot for my concerns about the current tracking to the next RA1 meeting?
We can hardly be successful if we do reach unreachable criteria.

PR #3154, opened 2022-09-05
[RA1] Remove duplicated {{ and }} in title

collivier:
The second one is falsy interpreted as part of title.
It may become a big issue if we dynamically convert them
to rst for GSMA docx.

### DIFF
--- a/doc/ref_arch/openstack/refs.bib
+++ b/doc/ref_arch/openstack/refs.bib
@@ -6,7 +6,7 @@
 	doi =		{10.17487/RFC1918},
 	url =		{https://www.rfc-editor.org/info/rfc1918},
         author =	{Robert Moskowitz and Daniel Karrenberg and Yakov Rekhter and Eliot Lear and Geert Jan de Groot},
-	title =		{{Address Allocation for Private Internets}},
+	title =		{Address Allocation for Private Internets},
 	pagetotal =	9,
 	year =		1996,
 	month =		feb,
@@ -21,7 +21,7 @@
 	doi =		{10.17487/RFC2119},
 	url =		{https://www.rfc-editor.org/info/rfc2119},
         author =	{Scott O. Bradner},
-	title =		{{Key words for use in RFCs to Indicate Requirement Levels}},
+	title =		{Key words for use in RFCs to Indicate Requirement Levels},
 	pagetotal =	3,
 	year =		1997,
 	month =		mar,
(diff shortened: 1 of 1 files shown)


pgoyal01 (approved):

karinesevilla (approved):

CsatariGergely:
This was merged with 2 approvals and with a failing job in less than a day. I think we should follow our own processes.

collivier:
@CsatariGergely I don't understand why you're focusing on RA1 process  - Please check master instead of well voted RA1 PR and you will see commits merged without any PR. And then any verification at all.

collivier:
@CsatariGergely RA1 is from far the best stream about its contents, its contributions and its processes. I would encourage u to promote those in the streams you're contributing.

CsatariGergely:
@collivier we have the same contribution rules in RA1 and other parts of Anuket Specifications documented [here](https://github.com/cntt-n/CNTT/blob/master/CONTRIBUTING.rst). I would like to encourage every sub-project to follow these rules including the ones where I'm contributing. We agreed to merge pr-s only with three approvals and this pr was merged with two. I agree that merging commits without a pr is also a problem and we should act on that behavior also.
I can not judge on which sub-project is the best in content or processes.
One of my reasons for looking into this is our discussion about the need for a documentation team. You commented in the [proposal page](https://wiki.anuket.io/display/HOME/Documentation+working+group#Documentationworkinggroup-Commentsabouttheproposal) that the model based on pr-s is working so I wanted to double check on this. For me the fact that we need to merge documentation related pr-s without the needed 3 approvals means that this process is not working at the moment.

collivier:
@CsatariGergely I would have expected more contributions than policies. I don't see any change proposal from all your comments counting reviewers (no revert ? no additional PR ?). It must be noted that TSC Co Chair activities is about chairing the meeting itself nothing more. Even if I generalize to the TSC activities (what's false here), TSC should help the project to be successful not the opposite. I'm very concerned if a few TSC members don't know the true active contributors, globally and per stream, or don't know what's changed in Anuket master since the last release.

And I don't list the road blocks such finding 3 experts in our community to review python upper contraints / tox / bibtex, nuances between docutils, sphinx and pandoc ,etc. How to manage jjb and tox changes in //... Without diving into the lack of PR dependency management here which forces us to merge + rebase compared to Gerrit.

PR #3167, opened 2022-09-09
Update references.rst

walterkozlowski:
Amended references as a result of GSMA/RM review


pgoyal01 (approved):

steelescotr (approved):

collivier (changes requested):
We must rather dump the common section needed by RM into RM RST. Else there is still a gap between docx where is the file is **manually** dumped and source where it's a link to somewhere else.
Please follow what RA1 did for the full automation. It's far better.

collivier:
In a long run, common will be deleted.

walterkozlowski:
@collivier, yes - perhaps in long run. This is 'now and here" situation.

collivier:
@walterkozlowski "now and here" solution should be to dump the content in RM as RA1 did.
Let's later when we could delete common.

walterkozlowski:
@collivier, I do not want us to loose these updates, so I want this PR to be merged.
I am OK with creating a new PR, to move (I would not say dump, sorry) separate list of references for RM.

karinesevilla (approved):

PR #3118, opened 2022-08-12
[RC2]: Correcting the title of the document

CsatariGergely:
Based on the discussion in #3045

PR #3172, opened 2022-09-12
nit: Fix #3118

collivier:
RC2 asks to wrap correctly.


collivier:
trivial but the gates are broken now

PR #3185, opened 2022-09-13
[RA1] Stop bookmarks in lowercase in single rst

collivier:
It also includes all shell commands in python
and runs doc8 vs the single rst too.

2 ref were not in lower case too which breaks
string substitutions.


collivier:
Needs to merge quickly to cherry-pick the full PR chain.
And it's pure python.

PR #3202, opened 2022-09-16
Switch to list table

collivier:
Pandoc fails when blank chars in cells then it
breajs tge full GSMA process.


collivier:
Please note there is no no visible changes in the rich diff.

karinesevilla (approved):

Issue #3191, opened 2022-09-15
[pcac]: Think about using a different theme instead of material


CsatariGergely:
The material-sphinx theme is not actively maintained anymore and it has an open bug related to Anuket (https://github.com/bashtage/sphinx-material/issues/127).
We should think about selecting an other theme. Some of the themes are listed here: https://sphinx-themes.org/

CsatariGergely:
#3212 Implements a pilot for the Alabaster theme for Sphinx. 

@collivier how can I try if Alabaster is correct where Material fails?

CsatariGergely:
Themes with document level navigation plane: 
https://sphinx-themes.org/sample-sites/piccolo-theme/
https://sphinx-themes.org/sample-sites/furo/
https://sphinx-themes.org/sample-sites/sphinx-book-theme/
https://sphinx-themes.org/sample-sites/sphinx-documatt-theme/

CsatariGergely:
Edit created a pilot for the Piccolo theme here: https://gerrit.opnfv.org/gerrit/c/opnfvdocs/+/73644

EditKoselak:
> Edit created a pilot for the Piccolo theme here: https://gerrit.opnfv.org/gerrit/c/opnfvdocs/+/73644

First attempt of the Piccolo theme:
![image](https://user-images.githubusercontent.com/61583300/202432873-baa7f282-3147-4065-9d4e-4085b31b9ea2.png)


We used the default Piccolo theme as a base and changed the followings:
- Color of the top navigation bar
- Size and padding of the logo in the top navigation bar
- Margins of the main text
- Color of the links in the main text and in the navigation bars + the link hover color in the navigation bars. We choose the greenish color from the Anuket brand colors, since it works better in dark mode.

The default Piccolo color is still used in the search button and in the arrow next to the Page contents label. If the proposed theme is approved, we can change these settings as well.

EditKoselak:
![image](https://user-images.githubusercontent.com/61583300/209103443-b8593f04-fe20-4855-827b-7b94625891cc.png)

The following updates have been done:

- changing the color of the search button
- changing the color of the Page contents label
- removing option to switch to dark mode

PR #3212, opened 2022-09-28
WIP: [gov]: Piloting sphinx alabaster theme

CsatariGergely:
This pr is only to figure out if the community likes the Sphinx Alabaster theme.
Part of the #3191 effort.


steelescotr (approved):
+1

CsatariGergely:
It looks like this: 
image

I miss the in document navigation.

collivier (changes requested):
very incomplete. That would lead to a disruptive navigation

CsatariGergely:
> very incomplete. That would lead to a disruptive navigation

Yes I totally agree. My only problem is that Material seems to be the most mature theme. I plan to pilot some others also if time permits, but so far I'm not convinced that we should use anything else than Material.

gkunz (approved):
I'd like to see this getting merged for Nile

CsatariGergely:
We decided to use the Picolo theme, so this needs to be closed.

PR #3294, opened 2022-12-26
NSM as federated network manager / Tungsten as CNI

wavell:
NSM as federated network manager / Tungsten as CNI


CsatariGergely:
@wawell are you sure you wanted to create this pr against Moselle? We just finished the work on Nile and the work on Orinoco starts soon, so Mosselle is a kinda old release already.

pgoyal01:
@CsatariGergely @rgstori Why is this authorized to be merged even without any reviews?

rgstori:
it's because the target branch is `stable:moselle` instead of `develop`
the target should be changed, but I would also recommend to add a branch protection rule for the `stable` branches (which I can't do as I'm not an admin)

rgstori:
I've now changed the target to `master`, but that has automatically brought over all the changes from moselle to nile - maybe worth to re-do this in a new PR based off `nile`?

CsatariGergely:
@wavell can you please create this pr against the master?

CsatariGergely:
> it's because the target branch is `stable/moselle` instead of `develop` the target should be changed, but I would also recommend to add a branch protection rule for the `stable` branches (which I can't do as I'm not an admin)

Maybe we should discuss this in the TSC or in a mail or here. According to @collivier having branch protection rules for the stable/* branches would kill our current cherry-picking practices. Off course I would be more happy with more rules and automation :)

PR #3305, opened 2023-01-23
Update CODEOWNERS

collivier:


karinesevilla (approved):

CsatariGergely (approved):
Okay with the change, however I have some generic questions.

I think we should indicate that RC1 is still a separate sub-project (or is it?).

Why do we have sub-project leader for /gov and for /tech. These are generic documents.

ildikov (approved):

steelescotr (approved):

PR #3306, opened 2023-01-24
Add redfish/ODIM to hardware management section of the reference model

Bharath-KKB:


SaadUSheikh (changes requested):
This is an excellent introduction to HIM but we should add below sub parts and suggest to add Monitoring part also.

1.server 
2. storage
3. Network 
4. KPI suggested / Monitoring 

@walterkozlowski comment: I agree with @SaadUSheikh about the need to have this detailed information.  The question is where and how to add it.  Anyhow, I think it should be a separate PR or a couple of PRs.  I would suggest that @SaadUSheikh  should approve this PR under the condition that the aspects he mentioned above are dealt with in subsequent PRs.

walterkozlowski (changes requested):
I like the contents.  The only change I would suggest at this point is to modify the attached diagram to show how this capability fits into the whole model, as described in previous sections.

jbhartley:
This is an excellent introduction to Redfish / ODIM, from a "Reference Model" perspective it would be very helpful to articulate where this fits relative to: ".Infrastructure Lifecycle Model" & "Configuration and Lifecycle Management" (Chap 9).  
As part of this "model" fit it would be helpful to know the fit relative to other known technologies, such as (not comprehensivec list): ODIM/redfish, pxe, cloud-init , IPMI/BMC, Canonical JuJu sit in relative to the layer model.

jbhartley (approved):
@Bharath-KKB as per discussion. I am approving this pull request and ask if you could please add additional material on RedFish/ODIM fits relative to other Infrastructure Automation and Life-Cycle components (Chapter 9). This relationship/fit should also be considered relative to Chap 3 (where placeed) and Chap 5 APIs.

Bharath-KKB:
Many thanks to Saad, Walter and John for the comments. I will work on the comments and update you.

walterkozlowski (approved):

doc/ref_model/chapters/chapter03.rst
```diff
@@ -357,10 +357,94 @@ The hardware infrastructure manager needs to support the following functional as
 - **Additional Management Functions**: include software and configuration life cycle management, identity management,
   access management, policy management (e.g. to enforce security policies), etc.
 
-Left for future use
--------------------
+The Redfish Standard and the ODIM Implementation
```
CsatariGergely, line 360: Can you please clarify the purpose of introducing Redfish and ODIM here? Is this a for information, kinda information sharing or the purpose is to mandate Redfish or ODIM in RM?

doc/ref_model/chapters/chapter03.rst
```diff
@@ -357,10 +357,94 @@ The hardware infrastructure manager needs to support the following functional as
...
+-	Gather crucial configuration information about southbound resources
+
+-	Manipulate groups of resources (“aggregates”) in a single action
+
+-	Listen to similar events from multiple southbound resources
+
+2.	PLugin layer
+
+This consists of one or more plugins, which abstract, translate, and expose southbound resource information to the resource aggregator through RESTful APIs.
+
```
CsatariGergely, line 423: It looks like this line makes the build fail (./chapters/chapter03.rst:423: D003 Tabulation used for indentation)

https://github.com/anuket-project/anuket-specifications/actions/runs/4342322387/jobs/7583535428

CsatariGergely (changes requested):

doc/ref_model/chapters/chapter03.rst
Bharath-KKB: Is there a guide covering these things ? I am new to Anuket doc reqs.

doc/ref_model/chapters/chapter03.rst
Bharath-KKB: The Introduction in the reference model is informational. Redfish provides a standard interface that can manage compute/network/storage devices. 
ODIM implements and exposes a single redfish API. ODIM can also help manage non redfish capable devices by way plugins (ACI fabric adapter being an example). We think it could be a good fit for the role of 'Hardware Infrastructure Manager' in Anuket reference implementation.

Bharath-KKB: Further this is not a mandatory component of the Reference Implementation, but is an optional component.

doc/ref_model/chapters/chapter03.rst
CsatariGergely: Sorry, we do not have a documentation guide, this could be fixed by simply providing a correct rst file. 
You can always build your doc locally with `tox -e docs` in the RM folder.

TFredberg (changes requested):
First let me say I am sorry that I give all feedback in one comment.
I do not see that it is a good thing to chop up the comments in many small individual improvement suggestions, since that would likely not get the wanted improvement as a whole.

After commenting on the whole text I also see that it could look negative, but the purpose of this comment is to ensure the Anuket RM suggestion adds value AND suggest how ODIM can be positioned ad by that improved over time, through reasonable expectations in areas where it can add value to and from ODIM as well.

360:
"The Redfish Standard and the ODIM Implementation"
It becomes very hard to understand this chapter when Redfish is mixed with ODIM. I would suggest an internal structure where Redfish is first described and then describe ODIM and where it uses Redfish. Now these two becomes intertwined which makes it hard to understand their relationships.

363:
"This section proposes a redfish based hardware infrastructure manager for the Anuket project. "
Reformulate, since the RM shall not suggest implementations, and the Anuket is not a "project" in the sense of an implementation project.

365:
"GSMA's "Cloud Infrastructure Reference Model (NG.126)"
Why reference the GSMA version of this RM when you can have an internal reference to Chapter 9. It could separately be mentioned that this is also adopted by GSMA if that add some value.

369:
"The schema allows the OpenAPI and json-schema toolchains to be used in implement a Redfish Client."
Textual problems with the sentence "...to be used in implement a ..."

372:
"... is a Linux Foundation projectand is the most mature ..."
Textual problem (projectand).
I don't think this RM shall have a statement (or even take a stand) of what currently is most mature of the Open Redfish clients. 

380:
"DMTF specifies the Redfish interface and model, seeking to expand the manageability domains. The prescription ..."
This paragraph is hard to follow and would deserve a figure showing the Redfish Profile prescription mechanism, the JSON input and the ODIM specific Redfish Profile instance.

404;
"The ODIM implementation supports the account ..."
This paragraph comes before the description of what ODIM is, and it is almost a carbon copy of a paragraph in the 2. Plugin layer.
Suggest to remove this occurrence.  

427:
"The ODIM project ..." paragraph (and onwards) is not part of the Plugin layer, but that is hard to tell from the way that the text is formatted.

431: Poor resolution if the figure. Hard to read content of the Service Layer.

437:
Inconsistent use of Registered Trademark symbol on the word Redfish. This is first occurrence with the symbol, but the word have been mentioned many more times before. The Redfish work also exist Capitalized and in all lower case.

441: Hard to follow when "domain model" an "data model" are used as synonyms mixed with each other. OK to explain that the domain model follows the Redfish data model, but stick to domain model for all consecutive explanations.

445: Why is there a company name (HPE) in the Open ODIM architecture?

449:
Second time the Plugin layer is described in a list. First time it is a Component and second time it is a Layer in the architecture.
  I suggest to either start with the architecture and then describe the implementation (components), which I would prefer, or
describe each component for itself with sub-layers inside them.
  I also see that the Plugin layer only describe FM related events over the Kafka MB, but I would guess that it also goes for PM/Telemetry, and if so, I think it should be mentioned. Are there more things that goes over the MB?
  Is there an Open or ODIM-internal model or API description (of the REST and Kafka) in between the Resource Aggregator (layers)?

455: 
Spelling of "fugure"
The figure do not show how ODIM can play a role as HW Infra Manager. Suggest to either make an overlay box suggesting where ODIM fits in, or just add "e.g. ODIM" after the HW Infrastructure Manager text in the figure.

441:
The paragraph is a bit unstructured and throw the reader into details and also status about the ODIM project immediately. I would suggest that this paragraph only deals with how ODIM fits into the NFV NW Ref model. Then there could be another paragraph or bulleted list that deals
with each of Compute, Network and Storage.
Here you should though pay attention to that Storage Resource exist on 2 possible levels that are rather differently managed and used in the system. I am not convinced that Swordfish are relevant for both of these. This should be discussed in the text. 
1. A Storage Box (appliance box with a complete Storage subsystem generally exposing SMB, FTP, other storage access protocols i.e. as a DAS or NAS).
2. A storage device enclosure e.g. a JBOF
Then it should also be mentioned that  storage devices on servers are not normally exposed as separate devices, but handled through the servers. This also ought to be mentioned for completeness of the readers understanding.

You seems to describe that ODIM have not implemented anything in the Networking area, but suggest that the user can do that themselves. I do not understand why the user shall first have a high level Networking manager that shall convert their network configurations into Redfish, pass it onto the HW Infra Managers NorthBound Redfish interface (as you seem to suggest with ODIM) and then have to implement internal networking for ODIM, just to find out that there are no one on the market that can take a Redfish modeled network configuration into their switch boxes.
As we spoke about, that you have a possibility to use ODIM southbound converters of non-Redfish protocols don't really help out the flow of the two first problems (why Redfish in the NorthBound and no ODIM implemented Networking). 
I would feel that networking is far more complex than what a YANG-Redfish converter and what a Redfish schema can take care of.

471:
I think the chapter content is a very good description of the status of the ODIM project and with a very clear figure. Maybe the title don't really reflect that fully.
But when arriving here in the PR, I think I would recommend to rewrite the proposal (reusing a lot of the text), but focus on a description where ODIM is positioned as an Open project toolbox, that can be used and contributed to, by companies that like to build their own HW Infrastructure Manager.
When I started reading, I got the impression that it intended to described how ODIM could be taken and directly fitted into the HW Infrastructure Manager. Either as a product or as a ready toolbox that just needed to be configured before it solved the problem.
It is rather clear that this is not the status, and if we like to get it there I think that expectations needs to be properly managed.

Bharath-KKB:
> Please add an editable source of your picture following the [image handling guides](https://github.com/anuket-project/anuket-specifications/blob/42bc2d9389123e54748ef27c082dd438f9070eda/artefacts/README.md?plain=1#L2).

Done

doc/ref_model/chapters/chapter03.rst
Bharath-KKB: Fixed

Bharath-KKB:
> 

Updated as per most comments except for lines 380/449/455/441/471

walterkozlowski:
Action on all to review the latest changes.

Bharath-KKB:
Closing this PR. Will raise a new PR for adding redfish only content.

PR #3343, opened 2023-04-17
fix broken GSMA operator platform link

rgstori:
old link returns 404
propose changing to  https://www.gsma.com/futurenetworks/5g-operator-platform/
or
https://www.gsma.com/futurenetworks/wp-content/uploads/2022/04/GSMA-Operator-Platform-Telco-Edge-Requirements-2022-v2.0.pdf


tomvanpelt (approved):
The "best" link depends on the goal. The latest version of OPG.02 would be available at [https://www.gsma.com/futurenetworks/wp-content/uploads/2023/03/GSMA-Operator-Platform-Group-Operator-Platform-Telco-Edge-Requirements-v4.pdf](https://www.gsma.com/futurenetworks/wp-content/uploads/2023/03/GSMA-Operator-Platform-Group-Operator-Platform-Telco-Edge-Requirements-v4.pdf), but it would be replaced by its successor in less than 6 months from now. The suggested link is less specific as the page links to multiple documents and document versions, but it seems to be the best option as it avoids having to update twice yearly.

rgstori:
just realised it's a duplicate of #3341

PR #3358, opened 2023-04-28
[RA2]: Separate build with GSMA doc biulding capabilities

CsatariGergely:
- Triggered only if the RA2 docs are modified
- Storing the biuld artifacts
- Building the GSMA docx also


petorre (approved):

rgstori:
shall we delete `build-docs.yml` then?

PR #3367, opened 2023-05-16
[RM Ch8]: Rephrasing O-RAN and Anuket relations

CsatariGergely:


TFredberg (commented):
I don't think that Anuket support of an O-Cloud is predicated on that RAx must support the O-Cloud. I don't think it is bad to do the investigation/analysis and check if Anuket RAx could support the O-Cloud, but it is not a "This would require ... how an Anuket RA could support ...". The lowest level of under-cloud that is needed is a Bare Metal resource supply, but the next possible layer is to also support an RAx component level.

CsatariGergely:
> I don't think that Anuket support of an O-Cloud is predicated on that RAx must support the O-Cloud. I don't think it is bad to do the investigation/analysis and check if Anuket RAx could support the O-Cloud, but it is not a "This would require ... how an Anuket RA could support ...". The lowest level of under-cloud that is needed is a Bare Metal resource supply, but the next possible layer is to also support an RAx component level.

RAx must support the technology and API-s required by an O-Cloud when the investigation is done the building principles or Anuket should be considered. This is why I think this is a "This would require ... how an Anuket RA could support ...".

walterkozlowski:
@CsatariGergely , @TFredberg , @petorre , @steelescotr , @bfcohen :  
This seems to be a circular discussion. Let me propose to cut this paragraph short, for the time being at least, and finish it just after the sentence "A potential alignment between Anuket and O-RAN that could be investigated is to see if Anuket could serve as an under-cloud to the O-Cloud." This potential investigation is important, e,g., this investigation is required by my organization (and I believe, others within GSMA).  And no-one is objecting to such investigation, I presume. We do not need to define the outcome before the investigation is completed (and we have not started it yet).  
After merging PR#3345 in the shortened form, as I propose above, further discussion about possible alignment with the industry initiatives like O-RAN can and should be moved to the TSC, so delegates from other organizations may provide their views.  After all, Anuket is not a dictatorship, and we should not stop the community to have an open discussion on such important topics.

doc/ref_model/chapters/chapter08.rst
```diff
@@ -912,11 +912,12 @@ connectivity in between the Network Functions. The O-Clouds also have the set of
 Acceleration Adaptation Layer (AAL) of how they are used from the Network Functions for their Radio-near functions.
 
 A potential alignment between Anuket and O-RAN that could be investigated is to see if Anuket could serve as an
```
pgoyal01, line 914: Anuket provides a set of specifications and so the question is whether Anuket specifications are aligned with O-cloud specifications or can serve as the O-cloud specifications.

This sentence needs rewording too many could' (possibility) and "is to see" (possibility). Why only O-cloud? 

```suggestion
Investigate alignment between Anuket and O-RAN's O-Cloud specifications.
```

doc/ref_model/chapters/chapter08.rst
```diff
@@ -912,11 +912,12 @@ connectivity in between the Network Functions. The O-Clouds also have the set of
...
-under-cloud to the O-Cloud. This would require Anuket to support the O-RAN HW Accelerators (as stated in RM Ch3 section
-Example of O-RAN AAL Interface :ref:`chapters/chapter03:example of o-ran acceleration abstraction layer interface`
-and a Layer2-centric networking infrastructure. It would enable the operators to have an internal Telco Cloud that
-supports both Core and RAN Network Functions, and in the extension possibly also other workloads in a shared Cloud
-that supports required Telco features and characteristics.
+under-cloud to the O-Cloud. This would require an analyzis how an Anuket Reference Architecure based on open source
```
pgoyal01, line 915: I line with previous sentence change:
```suggestion
This would require an analyzis how an Anuket Reference Architecure based on open source
```

pgoyal01 (changes requested):

PR #3412, opened 2023-09-18
[RA2, RC2, RI2]: Removal of .readthedocs.yaml

CsatariGergely:
Readthedocs yaml files caused a circular reference in the documentation. Even removal of the config file from the root project and the RM did not solve the problem (no, not even in RM). This pr attempts to fix the problem by removing all of the config files.


rgstori (approved):

collivier (changes requested):
Because of RTD forcing the use of .readthedocs.yaml, we don't have another choice than splitting the git repository per stream. This PR must be applied on next RM's git repository and not here.

Everything is already ready on my own repositories but sadly I don't have the rights needed on anuket-project organization. **Please give me them to complete the migration.**

Be free to ask for any  detail about the migration during a Weekly Technical meeting.

CsatariGergely:
This PR is not relevant anymore.

